PR #5, opened 2020-07-12
Plotting tools

ryantibs:
Hi all, I made some big additions here to the covidcast R package with respect to plotting tools.  

@capnrefsmmat @krivard @jsharpna The pull request might be a tiny bit nonstandard to handle because I also restructured the folders, and renamed the R package from "covidcastR" to "covidcast". Now the R package sits in R-packages (and the Python package sits in Python-packages).  This is probably going to mess up the package documentation in the docs folder, which I didn't try to fix (sorry).  Also, I think as soon as the pull request is merged, the `httr:user_agent` call on line 446 of covidcast.R is going to be wrong (I'm guessing).  Its input used to be "covidcastR"; it should probably be changed to "R-packages/covidcast" (I'm guessing).  

There are a bunch of TODOs left on my list that I didn't get to; some to do with plotting tools; some to do with basic signal processing, like wrapping my correlations code in the R-notebooks/signal-correlations.Rmd file into a function, to include in the covidcast R package.  I'm not sure what our workflow is going to be: for example, whether I should list them here in the issues tracker (since it's a public GitHub).  I'll post a message on slack, and then @capnrefsmmat @krivard @jsharpna can decide whether to put them in the issues tracker.


capnrefsmmat:
Very nice! The plots look quite good and these look like useful features to have. The maps will be particularly nice when we're exploring new signals...

A to-do list for me:

- [ ] Fix the merge conflicts. (Should be easy; it's just the documentation changes I made in #4.)
- [ ] Change `covidcast_signal` back to its previous behavior of only calling `covidcast_meta` when necessary, since this is required for us to be able to load and test signals not yet in the metadata.
- [ ] Update the pkgdown and Sphinx configs so they continue to drop the documentation in `docs/` with the correct directory names. I'm going to keep the prior directory names for the documentation pages, because I don't want to break links. (We have no way to redirect from old pages to new pages on GitHub Pages.)
- [ ] Update the repo and package READMEs to have `install_github` use the new path.
- [ ] Put the example plots into the package vignette, or into a separate new mapping/plotting vignette.
- [ ] Update delphi-epidata docs page to use "covidcast" instead of "covidcastR".
- [ ] Test that the class feature works well with dplyr. (I assume it will forward to the data frame methods if you do, say, `df %>% filter(...)`; will the attributes survive and be correct?)
- [ ] Check that the county FIPS codes included in the package include our weird exception cases.
- [ ] Draft an email to the API user mailing list announcing the new features (and name).

(And don't worry, the user-agent need not changed. It's just a value sent in the HTTP request headers that will let us see how much each client is being used. I may keep it as covidcastR just so we can distinguish between this package and the Python one, to know how much each is being used.)

ryantibs:
Thanks Alex.  Two quick comments/replies: the class structure should respect all existing functionality except the `print()` function from `dplyr`(I think it's called `print.rowwise_df()`), I overwrote this to give us a custom print function.  

And with respect to calling `covidcast_meta()`, makes sense to have this fail gracefully so we can test out signals not in the meta data.  Note that I used the mean and the standard deviation from the the meta data to define the auto range for the color scale for the choropleth maps (like we do on covidcast.cmu.edu), and the auto range for the size scale for the bubble maps.  These are set as attributes of the `covidcast_signal` object.   So in the initial call to `covidcast_signal()`, we should try to get these from the meta data, but just fail gracefully if meta data is not available, and set these attributes to NULL.  If these are NULL and the `range` isn't specified in the call to `plot()`, then we can have it fail outright there, and tell the user to set the `range`.

krivard:
> we can have it fail outright there, and tell the user to set the range.

Why not just calculate the mean and stdev from the df directly?

ryantibs:
Response to Katie's question: because the given `covidcast_signal` data frame need not contain all the relevant time values for this calculation.  The meta data (originally) calculated this mean and standard deviation over all time.  I'm not sure but it seems like to me we must have changed it to being a sliding history (based on the fact that the meta data has changed for me, slightly, pinging it on successive days).  In any case, whatever time values the calculation is based on in the meta data, need not be present in the given data frame (which could contain just one time value, for example).  Silently using the given mean and standard deviation to compute the range could be misleading because the high/low colors in the resulting map would be very different from what they might be expecting.

capnrefsmmat:
We haven't changed the metadata calculation yet; it changes slightly because it's over all history, and all of history gets updated daily with new data.

ryantibs:
Ah OK.  Well I wasn't describing it well but the way we originally discussed it when first David implemented it was that "all history" would mean the beginning of time until when David implemented it :), and then it would be fixed from then on.  In any case, I'm actually happy with the way it looks now (and the custom scale for cases/deaths): the colors on the map looks sufficiently dynamic to me.  This is just off hand, not based on a systematic inspection.

capnrefsmmat:
> Ah OK. Well I wasn't describing it well but the way we originally discussed it when first David implemented it was that "all history" would mean the beginning of time until when David implemented it :), and then it would be fixed from then on.

That's the logic we used for the direction calculations and the scaling they use ("all history up to April 22nd"), but the map scale is based on all of history from the metadata endpoint.

krivard:
I absolutely agree that the client should use the mean and stdev from the meta signal when it is available. For `wip` signals there is no metadata, so I don't really see the point of making the user provide a mean and stdev when we could just compute it for them. We can give them the option of providing them if you like, but I don't think it should be required.

ryantibs:
Ah I see.  Well if this was *only* done for `wip` signals, then sure, that sounds fine---we're the only users of `wip` signals so we should do what's best for us.

But still it can be a dangerous to make this happen silently, in general, without even a warning.  If I have a `covidcast_signal` object `x`, say which is found in the meta data and has the right attributes, but I set `attributes(x)$mean = NULL` and `attributes(x)$stdev = NULL` either accidentally or on purpose, and then call `plot(x)`, then I get a choropleth map with an unexpected color scale, compared to what I should see.  So we should at least issue a warning that the mean and standard deviation aren't found in the attributes, and hence we're setting the range based on the observed mean and standard deviation of the given object, so the color scale or size scale may not match what they expect, or something like that.

By the way, my OCD is kicking in and compelling me to say that we use both API `stderr` and `stdev` in the API, which are two conflicting uses of the abbreviation for "standard"!

krivard:
> So we should at least issue a warning that the mean and standard deviation aren't found in the attributes

Yep, that seems wise -- I think we're on the same page now. :)

capnrefsmmat:
I just did a Bad Git Thing and rebased someone else's branch to resolve the merge conflicts; I should have done a merge, but somehow I find rebasing easier. Ryan, if you want to make any additional commits, you'll have to force-pull from this branch **first**, otherwise merging things will get hairy. Sorry.

The branch now contains a bunch of things:

- Expanded documentation. You can preview the pkgdown site by opening `docs/covidcastR/index.html`.
- The plotting demo notebook has been turned into a package vignette that gets rendered as part of the documentation website. I've also expanded the getting-started vignette to show how to use the census data provided in the package.
- `covidcast_signal` should now work with signals without metadata, with a warning when we set plot scales.
- I removed some of the plotting options relating to choosing a date range, in favor of having the user use standard filtering on the data frame before giving it to the plotting functions. That's consistent with how ggplot or other plotting packages would handle things.
- I fixed some check warnings, although a couple remain.
- Various other tweaks you can see in the commit history.

I think this is ready for people to test, using

```r
devtools::install_github("cmu-delphi/covidcast", ref = "plotting-tools",
                         subdir = "R-packages/covidcast")
```

We'll want some testing before we merge this, since there are a lot of plot options and details to get right.

Only remaining items:

- We'll have to update the delphi-epidata docs to use the new name.
- It may be worth adding `covidcast_signals` at the same time. We're already changing the functions and adding a class to the output; might be worth changing it so we support multiple signals, so we don't have to change it repeatedly. (For example, I think we'd want to change from having a single signal attribute on `covidcast_signal` objects to having separate columns for the source and signal, so one can get multiple signals in one data frame.)

PR #6, opened 2020-07-17
R/Python client support for issue dates

capnrefsmmat:
Currently only for Python; will extend to the R client later. I'm basing this on the branch for #5 so the R client changes will all happen at once. We should only merge this when #5 is ready.

I'm planning to reuse the same example used here in the R documentation vignette, so there's a good example of issue dates in both and users don't have to go hunting.


capnrefsmmat:
I now think this is ready: the R and Python clients both work, the R vignettes work (which is basically my testing...), and there are good examples.

This can be merged, but before I can release the new covidcast Python client on PyPI, we'll need to push delphi-epidata 0.0.7 to PyPI with issue date support. I've already pinged David and linked him to https://github.com/cmu-delphi/delphi-epidata/pull/156

PR #7, opened 2020-07-24
correlation app ready for review

nloliveira:
It needs review of app description as well (English review).


krivard:
@statsmaths, are you available to review?

statsmaths:
Yes, happy to take a look at this! Will try to work on it right now and let you know if I run into any issues.

statsmaths:
@nloliveira Taking a look at the correlation app right now. The code looks nice and is well documented; the shiny interface also looks great and is quite easy to understand (at least for anyone vaguely familiar with the Delphi data).

I did have trouble getting the app to fully run, though. The current code imports the package **covidcastR**; I am not sure where to get that package. I switched it just **covidcast** (no R) and the code ran fine and spun up the site. However, I then ran into an error when trying to create the plot. I traced this back, I think, to the function `covidcast::covidcast_signal`. Perhaps the covidcastR package has a different version of the function and that's the issue? This happened when running with the default parameters.

Here's the traceback in the console.

```
Warning: Error in charToDate: character string is not in a standard unambiguous format
  224: stop
  223: charToDate
  222: as.Date.character
  220: covidcast_signal
  217: get_sensors [helpers.r#111]
  216: eventReactiveHandler [/Users/admin/Desktop/covidcast/docs/R-notebooks/CorrelationShinyApp/server.R#43]
  214: handlerFunc
  201: func
  199: f
  198: Reduce
  189: do
  188: hybrid_chain
  187: <reactive:eventReactive(input$runButton)>
  186: .func
  183: contextFunc
  182: env$runWith
  175: ctx$run
  174: self$.updateValue
  172: sensors
  167: renderPlot [/Users/admin/Desktop/covidcast/docs/R-notebooks/CorrelationShinyApp/server.R#53]
  165: func
  125: drawPlot
  111: <reactive:plotObj>
   95: drawReactive
   82: origRenderFunc
   81: output$corrPlot
    1: shiny::runApp
```

The error on the web app is attached as a screen shot.

Capture d’écran 2020-07-27 à 17 53 55

capnrefsmmat:
Yep, the new version of covidcast changed the date format from YYYYMMDD to YYYY-MM-DD in the `covidcast_signal` arguments, to be consistent with `as.Date` and some other new functions in the package. Sorry for the breakage.

nloliveira:
Thanks @statsmaths! Alex updated the package (including the name change) yesterday, and I had already opened the PR. I'll make the necessary changes and I'll ping you when it's ready for review again.

nloliveira:
@statsmaths changes were only in the date processing and it should work now

statsmaths:
Great, thanks for the quick response. All seems good now with the changed code. 

I ran a number of queries and the results were returned as expected. As said above, the code was clean and well documented.

The only thing left that I found is regarding the eventual use-case of this app. Is the idea that users will clone the repository and run the application locally? Or is it intended to be served from somewhere as a service? If it's the first, there should probably be some minimal documentation about how to spin up the application. Maybe this is somewhere and I just did not know where to find it.... It's straightforward if you're familiar with shiny apps, but not everyone who uses R, let along Python, will be. If it's intended as a service to be run from somewhere, this is not needed as much, but one design flaw is that the code seems to fetch new data server-side whenever it's requested. That could get slow with ever moderate usage, and could be greatly sped up with some build-in caching. Maybe not something to do right now, but just a thought for future development.

docs/R-notebooks/CorrelationShinyApp/helpers.r
```diff
@@ -0,0 +1,323 @@
...
+    transmute (geo_value = 1000*as.numeric(STATE) + as.numeric(COUNTY),
+               population = POPESTIMATE2019)
+  county_pop$geo_value <- sprintf("%05d", county_pop$geo_value)
+  population_mapping$county = county_pop
+  
+  state_pop = read.csv(census_url) %>% 
+    mutate(geo_value = as.numeric(STATE)) %>% 
+    group_by(geo_value) %>% 
+    summarise(population = sum(POPESTIMATE2019))
+  state_crosswalk <- read.csv("https://raw.githubusercontent.com/cmu-delphi/covid-19/main/geographical_scope/StateToHHS.csv?token=AHQABX4BXLJ3S3SJFSOGPFC7EBAIA") %>%
```
capnrefsmmat, line 26: This is a private repository -- we can't use this URL in a public notebook.

The covidcast package now includes `county_census`, `state_census`, and `msa_census` data frames that have the populations you want, although we don't have HRRs in the package yet. You should probably use these for consistency.

docs/R-notebooks/CorrelationShinyApp/ui.R
```diff
@@ -0,0 +1,177 @@
...
+                                          multiInput(
+                                            inputId = "signals_to_plot",
+                                            label = h4("Signals to correlate with response"), 
+                                            choices = signal_meta, 
+                                            selected = c("doctor-visits: smoothed_cli",
+                                                         "fb-survey: smoothed_cli",
+                                                         "fb-survey: smoothed_hh_cmnty_cli",
+                                                         "google-survey: smoothed_cli", 
+                                                         "ght: smoothed_search", 
+                                                         "indicator-combination: nmf_day_doc_fbs_ght")
```
capnrefsmmat, line 113: This combination signal is no longer produced. You probably want `nmf_day_doc_fbc_fbs_ght`.

PR #20, opened 2020-08-05
Add CI framework, linting, and basic test for Python package

chinandrew:
Summary of changes:

1. Add github actions CI to run pylint and pytest on PR and push to master.
2. Add dev requirements file for testing
3. Fix linting errors
4. Add one basic unit test to get that started

Other notes
1. There are a few lint warnings regarding function interfaces (such as the above) which I've ignored in the comments with `pylint: disable` so I don't change any code. e.g. there's a warning brought up by having a function called `signal` and arguments within that function called `signal`. Would probably recommend renaming the argument to `signal_name` or something to avoid this, and in general would be nice to fix these but would consider low pri.


chinandrew:
> Looks good, except your `.idea` directory snuck into this PR; fine to merge once that's gitignored.

Oops thanks for catching, removed.

PR #21, opened 2020-08-06
Add unit tests for remaining functions except `signal`

chinandrew:
Summary of changes:
1. Unit tests added for all methods except for signal, which is incomplete and will be done in a different PR. All external calls are mocked out.
2. Slight correction to tests directory since the file wasn't in the right subdir.
3. Use the .pylintrc that was already present and remove redundant disables.
4. Rename github action to include "python"

Other notes
- Linting is disabled on test directory for now, but don't feel super strongly so open to debate. 
- Also side note is that the CI right now runs in ~30-1min with no multiplier, just to keep in mind for [billing](https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions). We get a few thousand runs a month so I imagine it'll be fine for a while, but they tend to add up surprisingly quick. We can always alter the triggers if needed to reduce volume, as right now it runs on each commit to a PR and then once on merge.

### DIFF
--- a/.github/workflows/python_ci.yml
+++ b/.github/workflows/python_ci.yml
@@ -1,7 +1,7 @@
 # This workflow will install Python dependencies, run tests and lint with a variety of Python versions
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
 
-name: Lint and Testing
+name: Python Lint and Testing
 
 on:
   push:
--- a/.github/workflows/python_ci.yml
+++ b/.github/workflows/python_ci.yml
@@ -28,7 +28,7 @@ jobs:
         pip install -r Python-packages/covidcast-py/requirements.txt
     - name: Lint with pylint
       run: |
-        pylint Python-packages/covidcast-py/covidcast/
+        pylint Python-packages/covidcast-py/covidcast/ --rcfile Python-packages/covidcast-py/.pylintrc
     - name: Test with pytest
       run: |
         pytest Python-packages/covidcast-py/
--- a/Python-packages/covidcast-py/covidcast/covidcast.py
+++ b/Python-packages/covidcast-py/covidcast/covidcast.py
@@ -9,7 +9,7 @@ import pandas as pd
 VALID_GEO_TYPES = {"county", "hrr", "msa", "dma", "state"}
 
 
-def signal(data_source, signal, start_day=None, end_day=None, geo_type="county",  # pylint: disable=R0913,W0621
+def signal(data_source, signal, start_day=None, end_day=None, geo_type="county",  # pylint: disable=W0621
            geo_values="*", as_of=None, issues=None, lag=None):
     """Download a Pandas data frame for one signal.
 
@@ -237,7 +237,7 @@ def metadata():
     return meta_df
 
 
-def _fetch_single_geo(data_source, signal, start_day, end_day, geo_type,  # pylint: disable=R0913,R0914,W0621
+def _fetch_single_geo(data_source, signal, start_day, end_day, geo_type,  # pylint: disable=W0621
                       geo_value, as_of, issues, lag):
     """Fetch data for a single geo.
 
--- a/Python-packages/covidcast-py/covidcast/covidcast.py
+++ b/Python-packages/covidcast-py/covidcast/covidcast.py
@@ -2,9 +2,8 @@
 import warnings
 from datetime import timedelta
 
-from delphi_epidata import Epidata
-
 import pandas as pd
+from delphi_epidata import Epidata
 
 VALID_GEO_TYPES = {"county", "hrr", "msa", "dma", "state"}
 
--- a/Python-packages/covidcast-py/tests/covidcast/test_covidcast.py
+++ b/Python-packages/covidcast-py/tests/covidcast/test_covidcast.py
@@ -7,13 +7,14 @@ from covidcast import covidcast
 
 
 def test_signal():
+    # TODO test happy path.
     with pytest.raises(ValueError):
         covidcast.signal("source", "signal", geo_type="not_a_real_geo")
-    # TODO Test happy path.
 
 
 @patch("delphi_epidata.Epidata.covidcast_meta")
 def test_metadata(mock_covidcast_meta):
+    # not generating full DF since most attributes used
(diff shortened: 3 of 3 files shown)


capnrefsmmat:
Thanks. Just to make sure I understand the docs correctly: by default, if we run out of GitHub Actions time, it just doesn't run until the next billing cycle, right? We don't get a surprise bill.

Also, bizarrely, the tests fail for me, even though debugging reveals the two data frames here match except for column order:

```
mock_covidcast = <MagicMock name='covidcast' id='140378474042480'>

    @patch("delphi_epidata.Epidata.covidcast")
    def test__fetch_single_geo(mock_covidcast):
        # not generating full DF since most attributes used
        mock_covidcast.side_effect = [{"result": 1,  # successful API response
                                       "epidata": [{'time_value': 20200622, 'issue': 20200724}],
                                       "message": "success"},
                                      {"result": 1,  # second successful API
                                       "epidata": [{'time_value': 20200821, 'issue': 20200925}],
                                       "message": "success"},
                                      {"message": "error: failed"},  # unsuccessful API response
                                      {"message": "success"}]  # no epidata
    
        # test happy path with 2 day range
>       assert covidcast._fetch_single_geo(
            None, None, date(2020, 4, 2), date(2020, 4, 3), None, None, None, None, None).equals(
                pd.DataFrame({"time_value": [datetime(2020, 6, 22), datetime(2020, 8, 21)],
                              "issue": [datetime(2020, 7, 24), datetime(2020, 9, 25)]}, index=[0, 0]))
E       AssertionError: assert False
E        +  where False = <bound method NDFrame.equals of        issue time_value\n0 2020-07-24 2020-06-22\n0 2020-09-25 2020-08-21>(  time_value      issue\n0 2020-06-22 2020-07-24\n0 2020-08-21 2020-09-25)
E        +    where <bound method NDFrame.equals of        issue time_value\n0 2020-07-24 2020-06-22\n0 2020-09-25 2020-08-21> =        issue time_value\n0 2020-07-24 2020-06-22\n0 2020-09-25 2020-08-21.equals
E        +      where        issue time_value\n0 2020-07-24 2020-06-22\n0 2020-09-25 2020-08-21 = <function _fetch_single_geo at 0x7fac6888aa60>(None, None, datetime.date(2020, 4, 2), datetime.date(2020, 4, 3), None, None, None, None, None)
E        +        where <function _fetch_single_geo at 0x7fac6888aa60> = covidcast._fetch_single_geo
E        +        and   datetime.date(2020, 4, 2) = date(2020, 4, 2)
E        +        and   datetime.date(2020, 4, 3) = date(2020, 4, 3)
E        +    and     time_value      issue\n0 2020-06-22 2020-07-24\n0 2020-08-21 2020-09-25 = <class 'pandas.core.frame.DataFrame'>({'issue': [datetime.datetime(2020, 7, 24, 0, 0), datetime.datetime(2020, 9, 25, 0, 0)], 'time_value': [datetime.datetime(2020, 6, 22, 0, 0), datetime.datetime(2020, 8, 21, 0, 0)]}, index=[0, 0])
E        +      where <class 'pandas.core.frame.DataFrame'> = pd.DataFrame

tests/covidcast/test_covidcast.py:56: AssertionError
========================================================= short test summary info =========================================================
FAILED tests/covidcast/test_covidcast.py::test__fetch_single_geo - AssertionError: assert False
======================================================= 1 failed, 5 passed in 0.86s =======================================================
```

chinandrew:
> Thanks. Just to make sure I understand the docs correctly: by default, if we run out of GitHub Actions time, it just doesn't run until the next billing cycle, right? We don't get a surprise bill.

That's my understanding, unless we've set up a nonzero spending limit. If we have a $0 limit and hit the quota, it should say something like "“The job was not started because the spending limit for Actions and Packages has been exceeded.” Should be able to check this in the Org Settings > Billing. That's also where the total use should be shown, as I think it's per-account, not per-repo, and I see the delphi-epidata also has one small job running on merge to master.

Also unrelated, but may be worth making the check required if we want to be safe on the merges (Repo Settings > Branches > Add Rule > Require Status Checks), but if it's just the two of us for now I don't feel strongly.


> Also, bizarrely, the tests fail for me, even though debugging reveals the two data frames here match except for column order:
I was actually wondering if this might happen since the dicts to generate the DFs are unordered. Let me changed to lists and we'll see if it still happens.

chinandrew:
> > Also, bizarrely, the tests fail for me, even though debugging reveals the two data frames here match except for column order:
> 
> I was actually wondering if this might happen since the dicts to generate the DFs are unordered. Let me changed to lists and we'll see if it still happens.

Actually since the function return is a json I instead added a function to sort df's before comparison. Let me know if it still fails. I'm a bit curious what factor causes it to pass on CI and for me locally but not you, but I guess that's what I get for thinking "works on my machine." 

Also at some point may be worth refactoring some of these more general utility/helper functions out on their own to decouple from the more specialized methods, but also going to leave it for now.

chinandrew:
> Test passes now. I'm on Python 3.6, which is the first version where dicts maintained their insertion order by default -- so probably the ordering changed between 3.5 (on CI) and 3.6.
> 
> I'm surprised `.equals` cares about column order; I would have thought column order doesn't matter in a data frame

That makes sense, good to know. To double check, we're working under 3.5 as the minimum version? 

Also two other notes on the CI:
1. You should be able to [skip the checks](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-status-checks#checks) for commits if you want, such as if you're just changing documentation.
2. If there's a real concern with hitting the quota from just PRs, we can avoid having the other repos/workflows fail to deploy new master branches if we switch to a different CI (e.g. [Circle](https://circleci.com/)). That way if the second CI hits the free tier limit, there's no concern that the Github Actions will be affect unless they independently hit the quota, which seems unlikely. It's nice to standardize under one tool, but not a huge problem to run two CIs. I also wouldn't be opposed to just chipping in and buying a few thousand more runs each month if that becomes a problem.

capnrefsmmat:
> That makes sense, good to know. To double check, we're working under 3.5 as the minimum version?

Now that I check, I see `setup.py` declares the minimum version to be 3.6. Sorry, I got confused with the R package, where I used some feature only available in R 3.5.0 onward, and so 3.5 stuck in my head. We can go with Python 3.6, I think.

PR #22, opened 2020-08-07
Python Type Annotations

chinandrew:
Summary of changes
1. Add type annotations to covidcast functions. 
2. Update CI to 3.6

Relatively low pri PR here, and open to discussing whether or not it's actually worth adding. I personally find type annotations helpful. Note that these annotations are not enforced at runtime by Python, and are purely for documentation and are often integrated into IDEs. I'm also running a type checker (mypy) on it as part of the CI, which can be slightly annoying and potentially not worth it - we can see how it goes and disable if needed.

A related discussion is if it's worth spending time standardizing some of the argument types to simplify code. I don't know enough about the end user, use case, likelihood of breaking existing user flows, and overall prioritization to make any judgments, but for example, seems like if an argument is 
```
geo_values: Union[str, Iterable[str]] = "*"
```
 and later it's standardized to a list if the input is a string
```
    if isinstance(geo_values, str):
        # User only provided one, not a list
        geo_values = [geo_values]
```
we might want to consider just having a list (and/or other iterable) as the only option for arguments, even if it's one input value.

### DIFF
--- a/.github/workflows/python_ci.yml
+++ b/.github/workflows/python_ci.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.5]
+        python-version: [3.6]
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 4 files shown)


capnrefsmmat:
I haven't used Python's optional typing before, but I'm happy with it as a way of improving readability and declaring the intent of functions

PR #25, opened 2020-08-11
Methods for getting GeoDFs and plotting

chinandrew:
Summary of changes:
- Add methods for merging shapefile data to signal data
- Add methods for detecting metadata off of the columns in a DF
- Add [shapefile reference files from 2019](https://www.census.gov/geographies/mapping-files/time-series/geo/cartographic-boundary.html) census at the 1:5,000,000 scale (chosen kind of arbitrarily, the 500k seemed too large)

Packaging for the shapefiles is **not** done yet. let me know if it's best to do that in this PR as well.

Screenshot from 2020-08-11 01-32-08

Couple notes:
- Since `geo_type` isn't passed along with the input df, I'm having it be explicitly stated in the args. There's ways we can try to pass it as a df attribute but don't know how reliable that will be. 
- Just states and counties at the moment given their higher priority
- Row order not preserved, could just sort by geo_value to keep things neat
- It's a left join to the shape data, so all states/counties will always be present regardless of the input (i.e. the output dimension for a given `geo_type` is always the same). Alternatively, could only include the regions we know we'll have a signal in, so removing stuff like US Virgin Islands or Guam. 
- Open to better naming scheme


capnrefsmmat:
Neat, geopandas looks pretty slick. I'm used to dealing with PostGIS's ideas of geometry columns, so this makes sense.

One consideration: Is it possible to attach metadata to the output from `covidcast.signal`, so it has its own `geo_type` attached and doesn't require the user to pass it to the plotting tools? The R package takes this approach. The problem is having to rely on pandas to keep the metadata attached through operations on the data frame.

chinandrew:
> Neat, geopandas looks pretty slick. I'm used to dealing with PostGIS's ideas of geometry columns, so this makes sense.
> 
> One consideration: Is it possible to attach metadata to the output from `covidcast.signal`, so it has its own `geo_type` attached and doesn't require the user to pass it to the plotting tools? The R package takes this approach. The problem is having to rely on pandas to keep the metadata attached through operations on the data frame.

I considered a few options, but none seemed amazing so I went with explicit approach. Open to suggestions:
- Try to attach a `geo_type` attribute as you suggested, but with the downsides you mention. 
- Changing the return type of `signal` to a dict or other custom object that has attributes, but this changes the interface of the function
- Guess the `geo_type` from the actual field values, but this isn't foolproof.

capnrefsmmat:
There is one extra option. It may be reasonable to add a `signals()` method that returns multiple signals in one data frame, say if you want several to pass to a plot or to your analysis code. If so, each row of the data frame would need to specify which source and signal the observation is from... and could also specify the geo_type.

That is, we could have data_source, signal, and geo_type columns as part of the data frame, and operations on the data would work on each (data_source, signal, geo_type) combination in the data frame. Since pandas will preserve columns through operations (unless you get rid of them, of course), the data will survive.

Would that work or would it get awkward to use?

chinandrew:
> There is one extra option. It may be reasonable to add a `signals()` method that returns multiple signals in one data frame, say if you want several to pass to a plot or to your analysis code. If so, each row of the data frame would need to specify which source and signal the observation is from... and could also specify the geo_type.
> 
> That is, we could have data_source, signal, and geo_type columns as part of the data frame, and operations on the data would work on each (data_source, signal, geo_type) combination in the data frame. Since pandas will preserve columns through operations (unless you get rid of them, of course), the data will survive.
> 
> Would that work or would it get awkward to use?



> There is one extra option. It may be reasonable to add a `signals()` method that returns multiple signals in one data frame, say if you want several to pass to a plot or to your analysis code. If so, each row of the data frame would need to specify which source and signal the observation is from... and could also specify the geo_type.
> 
> That is, we could have data_source, signal, and geo_type columns as part of the data frame, and operations on the data would work on each (data_source, signal, geo_type) combination in the data frame. Since pandas will preserve columns through operations (unless you get rid of them, of course), the data will survive.
> 
> Would that work or would it get awkward to use?

I can see adding a geo_type column to the `signal()` method. Would mean a change in the return value of the dfs for anyone currently using it, would that be ok? Then if we want a `signals()` method, it just calls `signal()` multiple times and concatenates them together.

capnrefsmmat:
I think it's okay to add columns to the `signal` return value; we did the same thing in the R package recently.

The reason I haven't fully implemented a `signals()` method in the R package was that I couldn't figure out how to make it work with the existing plotting and mapping tools in the package -- if you pass them a data frame with multiple signals or geo types, what should they do? That's why I raised the idea here -- it interacts with how you'd do the mapping here.

I do like the idea of joining on geometries, because then the plotting code doesn't really have to know what geo type it is specifically; it just needs to know what polygons to use.

chinandrew:
Does it make sense to also add a `data_source` and `signal` column? And then downstream functions take in `(data_source, signal, geo_type)` to filter by . Downside is data duplication since `signal()` starts returning 3 columns which all contain only 1 value.

capnrefsmmat:
Yeah, that sounds right. That's what the R package is doing, although I still hadn't solved the question of how to make the plotting code handle it.

The data duplication is annoying, but dealing with one data frame is nicer than dealing with multiple data frames with custom metadata attributes somehow crammed in, which might not work well.

chinandrew:
Got it, let's go with that then for now. I'll open up a separate PR for it tonight.

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -1,28 +1,42 @@
 """This contains the plotting and geo data management methods for the COVIDcast signals."""
 
 import os
-from datetime import date
 
 import geopandas as gpd
 import pandas as pd
 
+from covidcast import covidcast
```
krivard, line 8: Can/should this import be reworded using .-notation to match `__init__.py`?

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -0,0 +1,82 @@
...
+
+import geopandas as gpd
+import pandas as pd
+
+from covidcast import covidcast
+
+SHAPEFILE_PATHS = {"county": "../shapefiles/county/cb_2019_us_county_5m.shp",
+                   "state": "../shapefiles/state/cb_2019_us_state_5m.shp"}
+
+CURRENT_PATH = os.path.dirname(os.path.realpath(__file__))
```
capnrefsmmat, line 13: It's probably better to use [pkg_resources](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#resourcemanager-api) to get the path to data distributed through the package, since I dunno what happens for eggs and all the zillion ways Python packages are distributed, and whether exact paths are preserved for them all

Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -244,6 +244,32 @@ def metadata() -> pd.DataFrame:
     return meta_df
 
 
+def _detect_metadata(data: pd.DataFrame,
+                     data_source_col: str = "data_source",
+                     signal_col: str = "signal",
+                     geo_type_col: str = "geo_type") -> Tuple:
+    """Given a DataFrame, return the signal attributes of that DataFrame.
+
+    Errors if any of the columns are heterogeneous.
```
capnrefsmmat, line 253: probably better to be explicit and say `"Raises ``ValueError`` if..."`

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -0,0 +1,82 @@
...
+    output_cols = list(data.columns) + ["geometry", "state_fips"]
+    shapefile_path = os.path.join(CURRENT_PATH, SHAPEFILE_PATHS[geo_type])
+    geo_info = gpd.read_file(shapefile_path)
+
+    if geo_type == "state":
+        output = _join_state_geo_df(data, geo_value_col, geo_info)
+    elif geo_type == "county":
+        output = _join_county_geo_df(data, geo_value_col, geo_info)
+    else:
+        raise ValueError("Unsupported geography type.")
```
capnrefsmmat, line 38:
```suggestion
        raise ValueError("Unsupported geography type; only state and county supported.")
```

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -0,0 +1,82 @@
...
+               geo_value_col: str = "geo_value",
+               geo_type_col: str = "geo_type") -> gpd.GeoDataFrame:
+    """Append polygons to a dataframe for a given geography.
+
+    Default values correspond to return of
+    covidcast.signal(). Currently only supports counties and states.
+
+    :param data: DataFrame of values and geographies
+    :param geo_value_col: name of column containing values of interest
+    :param geo_type_col: name of column containing geography type
```
capnrefsmmat, line 26: For the docs, it'd be good for this docstring to explain what a GeoDF is and what package it comes from, and what it means to "append polygons to a dataframe". That is, it should link to geopandas, say explicitly that there will be a new column with a certain name, say that the column is the geometry of each location, etc.

Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -244,6 +244,32 @@ def metadata() -> pd.DataFrame:
...
+
+    :param data: dataframe with geo_value an
```
capnrefsmmat, line 255: did you accidentally a word?

Python-packages/covidcast-py/shapefiles/county/cb_2019_us_county_5m.cpg
```diff
@@ -0,0 +1 @@
+UTF-8
```
capnrefsmmat, line 1: Did these shapefiles come from the Census? (Maybe the ones we already used in the R package?) Would be good to have a note somewhere with the source, for future updating and for credit. And in case there are licensing restrictions. (Though Census products would all be public domain)

Python-packages/covidcast-py/covidcast/covidcast.py
chinandrew: Agreed, added.

Python-packages/covidcast-py/covidcast/plotting.py
chinandrew: Added some more info, let me know what you think

Python-packages/covidcast-py/covidcast/covidcast.py
chinandrew: Oops, fixed. I don't know if your comment was intentional or not lol.

Python-packages/covidcast-py/shapefiles/county/cb_2019_us_county_5m.cpg
chinandrew: Yea they're from here https://www.census.gov/geographies/mapping-files/time-series/geo/cartographic-boundary.html. Added a note to the plotting docstring.

Python-packages/covidcast-py/covidcast/plotting.py
chinandrew: I don't have a strong preference, if convention is to use relative imports we should stick with that for consistency.

Python-packages/covidcast-py/covidcast/plotting.py
chinandrew: Added, though I haven't worked with packaging static files in this way before so worth double checking. For some reason the package installs and runs fine regardless of if I include `package_data` in the[ setup.py](https://setuptools.readthedocs.io/en/latest/setuptools.html#including-data-files). Need to dig a bit deeper to fully understand whats going on.

The other thing I learned is that geopandas can read from URLs as well as zip files, so you can actually run the following. I know we discussed not pulling data at runtime and I think that's still the right idea, but just wanted to point this out as an FYI.
`gpd.read_file("https://www2.census.gov/geo/tiger/GENZ2019/shp/cb_2019_us_county_20m.zip")`

chinandrew:
> Looks pretty good -- my requested changes are mainly docs-related.
> 
> Once the mapping code is done, we probably want to add a section to the Examples page in the docs with a bunch of examples, including examples of using `get_geo_df` to do other things besides our own plotting functions. (Just so people realize they can work with the polygons directly.) But that can come after the other PR is ready

Yea I think there's a good amount of documentation work that needs to get done, will add to the second PR.

Python-packages/covidcast-py/covidcast/covidcast.py
capnrefsmmat: https://knowyourmeme.com/memes/i-accidentally

Python-packages/covidcast-py/covidcast/plotting.py
capnrefsmmat: As long as we can run the tests successfully, I suspect it's fine; my main concern is with the distribution of eggs and wheels and such.

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -0,0 +1,97 @@
...
+                   "state": "shapefiles/cb_2019_us_state_5m.zip"}
+
+
+def get_geo_df(data: pd.DataFrame,
+               geo_value_col: str = "geo_value",
+               geo_type_col: str = "geo_type") -> gpd.GeoDataFrame:
+    """Append polygons to a dataframe for a given geography and return a geoDF with this info.
+
+    This method takes in a pandas DataFrame object and returns a GeoDataFrame object from the
+    geopandas package (https://geopandas.org/).
```
capnrefsmmat, line 24:
```suggestion
    `geopandas package <https://geopandas.org/>`__.
```

reStructuredText links are weeeiiird

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -0,0 +1,97 @@
...
+
+    After detecting the geography type (either county or state) for the input, loads the
+    GeoDataFrame which contains state and geometry information from the Census for that geography
+    type. Left joins the input data to this, so all states/counties will always be present
+    regardless of the input i.e. the output dimension for a given geo_type is always the same
+    regardless of input. Finally, returns the columns containing the input columns along with a
+    `geometry` (polygon for plotting) and `state_fips` (FIPS code which will be used in the
+    plotting function to rearrange AK and HI) column. Coordinate system is GCS NAD83.
+
+    Default arguments for column names correspond to return of covidcast.signal(). Currently
```
capnrefsmmat, line 34:
```suggestion
    Default arguments for column names correspond to return of :py:func:`covidcast.signal`. Currently
```

for cross-refs in the rendered docs

Python-packages/covidcast-py/covidcast/plotting.py
chinandrew: thanks, I haven't worked with sphinx before so will have to read up on it

PR #27, opened 2020-08-17
2. Choropleth Plot Functions

### LINKED ISSUE #13: Find or produce county and state data files for the Python package
The R package has some convenience data included to look up counties, MSAs, etc., and includes polygons to map some of them. Even if the Python package doesn't include mapping features, we should find a good source of similar data. 

If there's an existing Python package that includes the data and polygons we want, we could just include a couple examples showing how to use it. If there isn't, we can adapt `R-packages/data-raw/make.R` to produce the data files in a format the Python package can use easily.

### LINKED ISSUE #18: Provide examples or integration to produce state and county maps in Python package
The R package provides mapping; the Python package does not. Admittedly it's not common for Python packages to provide their own plotting in the way it is for R packages to.

We should look at [Cartopy](https://scitools.org.uk/cartopy/docs/latest/) and other relevant Python packages and see what it'd take to make mapping easy in the Python package. That need not mean implementing maps ourselves -- we could, for example, discover that by rearranging our data in a certain way, it's easy to provide to Cartopy. Then we can just provide the right adapter and give some examples of usage in the documentation.

This probably depends on #13.

chinandrew:
Closes #13 and closes #18 
Plotting function now implemented in code (previously was just me messing around a Jupyter Notebook).

Current functionality - running something like this yields the plot:
```
from covidcast import covidcast, plotting
from datetime import date
from matplotlib import pyplot as plt

covid = covidcast.signal('fb-survey','smoothed_cli', start_day = date(2020,8,3), end_day = date(2020,8,4), geo_type = "county")  
plotting.plot_choropleth(covid) 
plt.show()
```
Screenshot from 2020-08-16 23-44-27

and you can pass keyword arguments to adjust details if needed, e.g. color scale
```
plotting.plot_choropleth(covid, cmap="viridis") 
plt.show()
```

Screenshot from 2020-08-16 23-45-43

Todos:
- Error/Exception handling
- Tests - It might be tricky testing the final plotting function, unsure what the best approach is. maybe building the expected DFs and/or jpgs? open to ideas here.
- Update documentation


Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -1,18 +1,55 @@
...
 import os
+from datetime import date
+from typing import Tuple
 
 import geopandas as gpd
+import numpy as np
 import pandas as pd
-
 from covidcast import covidcast
+from matplotlib import pyplot as plt
```
capnrefsmmat, line 11: probably needs to go in setup.py/requirements.txt

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -1,18 +1,55 @@
...
+    kwargs["vmax"] = kwargs.get("vmax", meta["mean_value"] + 3 * meta["stdev_value"])
+    kwargs["cmap"] = kwargs.get("cmap", "YlOrRd")
+
+    sm = plt.cm.ScalarMappable(cmap=kwargs["cmap"])
+    plt.colorbar(sm, boundaries=np.linspace(kwargs["vmin"], kwargs["vmax"], 8),
+                 orientation="horizontal", fraction=0.02, pad=0.05)
+    plt.title(f"{data_source}: {signal}, {day_to_plot.strftime('%Y-%m-%d')}")
+    for shape in _project_and_transform(data_w_geo):
+        shape.plot("value", ax=ax, **kwargs)
+    return data_w_geo
```
capnrefsmmat, line 50: It's been a long time since I used matplotlib. But should we return a matplotlib object here, such as the figure object, so the user can further customize the plot by calling various matplotlib functions?

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -1,18 +1,55 @@
...
 
+def plot_choropleth(data: pd.DataFrame,
+                    time_value: date = None,
+                    **kwargs) -> gpd.GeoDataFrame:
+    """Given the output df of covidcast.signal(), plot the choropleth.
+
+    Plot is in a style similar to the website https://covidcast.cmu.edu.
+
+    :param data: DataFrame of values and geographies
+    :param time_value: date of interest for the geography
```
capnrefsmmat, line 27:
```suggestion
    :param time_value: If multiple days of data are present in ``data``, map only values from this day. Defaults to plotting the most recent day of data in ``data``.
```

Python-packages/covidcast-py/covidcast/plotting.py
chinandrew: added

Python-packages/covidcast-py/covidcast/plotting.py
chinandrew: I'm not very familiar with mpl practices but that sounds right. Will change to return the figure.

chinandrew:
> Just some suggestions for the draft.
> 
> As for testing... the most rudimentary testing option would be a script that produces multiple maps in one document, and we eyeball it to ensure they rendered. A more sophisticated option would be saving the "reference" images and comparing against them, but this means that if e.g. matplotlib anti-aliases text slightly differently, the tests will fail. Web browsers have spent a lot of engineering time making robust versions of tests like this.

So this has been a bit of a pain. [MPL](https://matplotlib.org/devel/testing.html#writing-an-image-comparison-test) has a built in image comparison thing that I can't get to work and keeps writing corrupt files, so ended up implementing it myself in this draft. [I've also read](https://stackoverflow.com/a/27950953) that it's not recommended to do image comparison for reasons similar to what you describe, and so an alternative is to use the associated numpy array with something like `np.frombuffer(fig.canvas.tostring_rgb(), dtype=np.uint8) `, and save it as a reference file instead of an image, which I've also implemented. Not pushed yet since I didn't want two identical tests being confusing, but basically looks like this: 
```
    fig = plotting.plot_choropleth(test_df)
    data = np.frombuffer(fig.canvas.tostring_rgb(), dtype=np.uint8)
    loaded = np.load("test.npz")
    np.array_equal(data, loaded["a"])  # I had just called the test reference data 'a'
```
I think it's less readable but potentially a more robust method, and you also only read data and avoid writing new images to disk  + having to clean them up.

Another thing is I get a `DeprecationWarning: Gdk.Cursor.new is deprecated` warning when importing pyplot from matplotlib, which I think is some rendering issue. Not sure if it's specific to me, will have to see dig a bit deeper.

capnrefsmmat:
> Another thing is I get a DeprecationWarning: Gdk.Cursor.new is deprecated warning when importing pyplot from matplotlib, which I think is some rendering issue. Not sure if it's specific to me, will have to see dig a bit deeper.

Sounds like you're using GNOME under Linux? GDK is a rendering backend there. Looks like Matplotlib wants you to use their Cairo backend instead.

I think you'll need to test file to specify a single Matplotlib backend instead of relying on the default -- for example, the default on my Mac is the Cocoa backend, which immediately throws an error because of the way I installed Python. Even if it worked, the PNG may not match. You'd need

```python
import matplotlib
matplotlib.use("Cairo")
```

(or AGG) at the very top of the file, before importing `pyplot`.

capnrefsmmat:
Looks like Cairo is not installed by default on macOS, but AGG is, so maybe AGG is the right backend choice.

Switching to AGG locally, I get this:

```
tests/covidcast/test_plotting.py F
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

    def test_plot_choropleth():
        test_df = pd.read_csv(os.path.join(CURRENT_PATH, "../reference_data/test_signal.csv"),
                              dtype=str)
        test_df["time_value"] = test_df.time_value.astype("datetime64[D]")
        test_df["value"] = test_df.value.astype("float")
>       plotting.plot_choropleth(test_df)

tests/covidcast/test_plotting.py:29: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
covidcast/plotting.py:47: in plot_choropleth
    orientation="horizontal", fraction=0.02, pad=0.05)
../../../../../miniconda3/lib/python3.6/site-packages/matplotlib/pyplot.py:2100: in colorbar
    ret = gcf().colorbar(mappable, cax = cax, ax=ax, **kw)
../../../../../miniconda3/lib/python3.6/site-packages/matplotlib/figure.py:2129: in colorbar
    cb = cbar.colorbar_factory(cax, mappable, **cb_kw)
../../../../../miniconda3/lib/python3.6/site-packages/matplotlib/colorbar.py:1566: in colorbar_factory
    cb = Colorbar(cax, mappable, **kwargs)
../../../../../miniconda3/lib/python3.6/site-packages/matplotlib/colorbar.py:1072: in __init__
    mappable.autoscale_None()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <matplotlib.cm.ScalarMappable object at 0x7feeb40e0860>

    def autoscale_None(self):
        """
        Autoscale the scalar limits on the norm instance using the
        current array, changing only limits that are None
        """
        if self._A is None:
>           raise TypeError('You must first set_array for mappable')
E           TypeError: You must first set_array for mappable

../../../../../miniconda3/lib/python3.6/site-packages/matplotlib/cm.py:374: TypeError
```

I think something is wrong with your your `colorbar` is being instantiated with a `ScalarMappable`, but I don't really understand the details without digging in further.

chinandrew:
agg is also what I needed to use in order to pull the matrix numpy matrix, and if that's what you have on mac let's go with that.  For some reason it also brings up a deprecation warning (fewer than before) which is raised by `matplotlib/backends/backend_gtk3.py`.... Cairo does too, so I'm in general just confused. It also seems to not matter if the import is before or after importing pyplot, and in the tests I can import agg later and it runs fine. Not going to worry too much about this though. I've also changed the test to the np array comparison so I don't have to deal with saving out a file and cleaning it to do the image compare.


On your specific error, I can't seem to replicate it running agg on 3.3.0, and even after downgrading to 3.0.3 it still doesn't cause problems. I assume you can't run the choropleth function at all? Seems like the colorbar is [expecting](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.colorbar.html) a ScalarMappable (the supposed optional case errors for me). I've made one small edit to the code to call the colorbar on the fig vs plt which I'm not convinced will do anything, but let me know. If that doesnt work, can you try adding `sm.set_array([])` right below `sm = plt.cm.ScalarMappable(cmap=kwargs["cmap"])`?

On documentation for the site, do you want to add this as a new page or append/rearrange the [examples](https://cmu-delphi.github.io/covidcast/covidcast-py/html/examples.html)  page?

chinandrew:
Also i had accidentally skipped checks in one of my previous commits while it was a draft. I've run them locally and they passed, and if i end up with another commit I'll re-enable them.

chinandrew:
On color scales, I added `legend=True` to each region and plotted them on the same figure, and the auto-generated legends were all the same scale. If i altered vmax, they'd change  (see below)  

same vmin/max
Screenshot from 2020-08-23 16-19-34


altered vmax
Screenshot from 2020-08-23 15-32-27

Also adjusted the colorbar to be continuous so its accurate and to match the R plot

Screenshot from 2020-08-23 16-16-23

capnrefsmmat:
I set `sm._A = []` in the test and it fails with

```
    def test_plot_choropleth():
        matplotlib.use("agg")
        # load expected choropleth as an array
        expected = np.load(os.path.join(CURRENT_PATH, "../reference_data/expected_plot_array.npz"))
        test_df = pd.read_csv(os.path.join(CURRENT_PATH, "../reference_data/test_signal.csv"),
                              dtype=str)
        test_df["time_value"] = test_df.time_value.astype("datetime64[D]")
        test_df["value"] = test_df.value.astype("float")
    
        fig1 = plotting.plot_choropleth(test_df)
        # get np array representation of plot
        data1 = np.frombuffer(fig1.canvas.tostring_rgb(), dtype=np.uint8)
>       assert np.array_equal(data1, expected["expected_1"])
E       assert False
E        +  where False = <function array_equal at 0x7fe7ec3f82f0>(array([255, 255, 255, ..., 255, 255, 255], dtype=uint8), array([255, 255, 255, ..., 255, 255, 255], dtype=uint8))
E        +    where <function array_equal at 0x7fe7ec3f82f0> = np.array_equal

tests/covidcast/test_plotting.py:30: AssertionError
```

However, if I hack `plot_choropleth` to save the png, I get this:

test

Looks fine. I wonder if we're running into very slight rendering differences across platforms, and if there's any way to get around them or just mark the test as flaky.

Modulo this issue, I think the code looks good. If we have to fall back to "build the docs to see if the maps look right" as a testing strategy, I guess that's what we have to do, unless there's an obvious easy way to fix this issue...

chinandrew:
Huh that's weird. Will look into it later tonight but the first thing I noticed was that your colorbar seems to differ from mine by 0.001 in a few cases.....

chinandrew:
Ok well now the tests are failing locally for me to. Testing plots is the worst. 

- Looking at the 3,686,400 total values (12.8 x 9.6 x 100dpi x 3 channels for RGB), ~9000 mismatch, and they're all off by at most 2. I'd feel better if they were all off by 1 and just chalk them up to floating point error, two seems acceptable. 
- The plots still look the same. I saved the plots and diffed the images with the following output. The text and colorbar don't get used in the comparison, so the diff looks mainly in the borders, with maybe some areas off by a tiny bit. Hard to tell.

Screenshot from 2020-08-24 23-35-17

And if I diff against your png:

Screenshot from 2020-08-24 23-40-46

I'm not sure we'll be able to get consistent across different platforms, but hopefully it shouldn't matter too much. For now, I've changed the assertion to use `np.allclose()` and set the threshold to be <=2 for any given RGB value. Let me know if this makes tests pass for you. If not, we can either adjust the threshold a little bit if it's reasonable, try a few more plotting things to try to get consistency (simpler values? adjust border edges?), or just do the manual inspection each time on build.

capnrefsmmat:
I guess that's just antialiasing, maybe, that differs between platforms? I'll try this out today.

capnrefsmmat:
Unfortunately the test still fails, and:

```
(Pdb) p np.sort(data1 - expected["expected_1"])[-500:]
array([245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245,
       245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245,
       245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245,
       245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245,
       245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245,
       245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245, 245,
       245, 245, 245, 245, 245, 246, 246, 246, 246, 246, 246, 246, 246,
       246, 246, 246, 246, 246, 246, 246, 246, 246, 246, 246, 246, 246,
       246, 246, 246, 246, 246, 246, 246, 246, 246, 246, 246, 246, 246,
       246, 246, 246, 246, 246, 247, 247, 247, 247, 247, 247, 247, 247,
       247, 247, 247, 247, 247, 247, 247, 247, 247, 247, 247, 247, 247,
       247, 247, 247, 247, 247, 247, 247, 247, 247, 247, 247, 247, 247,
       247, 247, 248, 248, 248, 248, 248, 248, 248, 248, 248, 248, 248,
       248, 248, 248, 248, 248, 248, 248, 248, 248, 248, 248, 248, 248,
       249, 249, 249, 249, 249, 249, 249, 249, 249, 249, 249, 249, 249,
       249, 249, 249, 249, 249, 249, 249, 249, 250, 250, 250, 250, 250,
       250, 250, 250, 250, 250, 250, 250, 250, 250, 250, 250, 250, 250,
       250, 250, 250, 250, 250, 250, 250, 250, 250, 250, 250, 250, 250,
       250, 250, 251, 251, 251, 251, 251, 251, 251, 251, 251, 251, 251,
       251, 251, 251, 251, 251, 251, 251, 251, 251, 251, 251, 251, 251,
       251, 251, 251, 251, 251, 251, 251, 251, 251, 251, 251, 251, 251,
       251, 251, 251, 251, 251, 252, 252, 252, 252, 252, 252, 252, 252,
       252, 252, 252, 252, 252, 252, 252, 252, 252, 252, 252, 252, 252,
       252, 252, 252, 253, 253, 253, 253, 253, 253, 253, 253, 253, 253,
       253, 253, 253, 253, 253, 253, 253, 253, 253, 253, 253, 253, 253,
       253, 253, 253, 253, 253, 253, 253, 253, 253, 253, 253, 253, 253,
       253, 253, 253, 253, 253, 253, 253, 253, 253, 253, 253, 253, 253,
       253, 253, 253, 253, 253, 253, 253, 253, 253, 253, 253, 253, 253,
       253, 253, 253, 253, 254, 254, 254, 254, 254, 254, 254, 254, 254,
       254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254,
       254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254,
       254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 255, 255, 255,
       255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255,
       255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255,
       255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255,
       255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255,
       255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255,
       255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255,
       255, 255, 255, 255, 255, 255], dtype=uint8)
```

Evidently the rendering on my platform is sufficiently different that many pixels are as different as they can be.

I'm going to guess, from the diff you showed, that this is the title text and we're seeing some kind of font rendering difference between platforms. That's going to be hard to prevent.

capnrefsmmat:
(Also do you just want to add `sm._A = []` to the code permanently? I've had to add it each time to get the tests to run)

capnrefsmmat:
Oh, one other bug to fix that became apparent when looking at the docs: `_join_county_geo_df` uses megacounty values to fill in counties that were not observed. It should also fill in the issue date, time value, lag, source, etc. You can see in your plotting examples where these are not being filled in for megacounties:

```
>>> covidcast.get_geo_df(data)
     geo_value time_value direction      issue  lag     value    stderr  sample_size geo_type data_source        signal                                           geometry state_fips
0        24510 2020-08-04      None 2020-08-06  2.0  0.375601  0.193356     587.6289   county   fb-survey  smoothed_cli  POLYGON ((-76.71131 39.37193, -76.62619 39.372...         24
1        31169        NaT      None        NaT  NaN  0.928208       NaN          NaN      NaN         NaN           NaN  POLYGON ((-97.82082 40.35054, -97.36869 40.350...         31
```

chinandrew:
> (Also do you just want to add `sm._A = []` to the code permanently? I've had to add it each time to get the tests to run)

Oops couldve sworn I did that but guess not, will add now.

chinandrew:
> Unfortunately the test still fails, and:
> 
> Evidently the rendering on my platform is sufficiently different that many pixels are as different as they can be.
> 
> I'm going to guess, from the diff you showed, that this is the title text and we're seeing some kind of font rendering difference between platforms. That's going to be hard to prevent.

I'm pretty sure the text isn't used in the comparison, since I tried removing them and then the tests still pass:

passes
```
    fig1 = plotting.plot_choropleth(test_df)
    data1 = np.frombuffer(fig1.canvas.tostring_rgb(), dtype=np.uint8) 
    assert np.allclose(data1, expected["expected_1"], atol=2, rtol=0)
```

also passes
```
    fig1 = plotting.plot_choropleth(test_df)
    data1 = np.frombuffer(fig1.canvas.tostring_rgb(), dtype=np.uint8) 
    fig1.axes[-1].remove()
    fig1.gca().set_title("")
    assert np.allclose(data1, expected["expected_1"], atol=2, rtol=0)
```
For your diff, these are uint8s, so 0-1 = 255. Still, differences of 10 are a bit annoying, even if it is just AA. We can up that threshold, though it's hard to tell at what point it's not longer a useful threshold.... or maybe at this point we just verify on build. I'm decently comfortable without automated tests for MPL and relying on them not to drastically alter their interface. Could be extra careful and separate as much of our logic out vs the MPL functions so it isolates them from any bugs on our end.

(255,255,10)

(245,245,20)

chinandrew:
> Oh, one other bug to fix that became apparent when looking at the docs: `_join_county_geo_df` uses megacounty values to fill in counties that were not observed. It should also fill in the issue date, time value, lag, source, etc. You can see in your plotting examples where these are not being filled in for megacounties:
> 
> ```
> >>> covidcast.get_geo_df(data)
>      geo_value time_value direction      issue  lag     value    stderr  sample_size geo_type data_source        signal                                           geometry state_fips
> 0        24510 2020-08-04      None 2020-08-06  2.0  0.375601  0.193356     587.6289   county   fb-survey  smoothed_cli  POLYGON ((-76.71131 39.37193, -76.62619 39.372...         24
> 1        31169        NaT      None        NaT  NaN  0.928208       NaN          NaN      NaN         NaN           NaN  POLYGON ((-97.82082 40.35054, -97.36869 40.350...         31
> ```

Thanks for catching, will fix.

PR #30, opened 2020-08-21
3. Plotting documentation

chinandrew:
Summary of changes:
- Add documentation for the new plotting functionality.

This PR follows #34 and #27.

New rendered docs have not been committed yet. waiting until we're ready for deploy.


Python-packages/covidcast-py/docs/conf.py
```diff
@@ -10,10 +10,9 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-# import os
-# import sys
-# sys.path.insert(0, os.path.abspath('.'))
-
+import os
+import sys
+sys.path.insert(0, os.path.abspath('..'))
```
chinandrew, line 15: needed to do this to get everything to run, unsure if we want to leave commented out or not

chinandrew:
This is still waiting on the other PRs so not ready to merge, but content is good to start reviewing. 

Also I've gitignored the rendered files since the plots make them pretty large. If there's a reason we need them version controlled I can re-add.

capnrefsmmat:
We don't yet have automation to build the rendered docs and deploy them, so we rely on GitHub Pages and an Action that copies them over to the `gh-pages` branch. So we'll need the final versions committed at some point, unfortunately.

capnrefsmmat:
I've pushed a couple extra commits with some copyedits. For the second one, should `sphinx_autodoc_typehints` be added to requirements.txt? Not sure what belongs there vs. in setup.py.

chinandrew:
> We don't yet have automation to build the rendered docs and deploy them, so we rely on GitHub Pages and an Action that copies them over to the `gh-pages` branch. So we'll need the final versions committed at some point, unfortunately.

Got it, sounds good. I'll just wait until we're finalized then before starting to commit those files again.

chinandrew:
> I've pushed a couple extra commits with some copyedits. For the second one, should `sphinx_autodoc_typehints` be added to requirements.txt? Not sure what belongs there vs. in setup.py.

Our requirements.txt is what the CI installs before running tests, so it will include linting and testing packages. May be worth renaming to something like "testing_requirements". Could also consider another one like a "dev_requirements" that is all the testing requirements + the packaging and build requirements (sphinx, wheel) that lets people install everything they need in one go, though for simplicity's sake having one requirements at this scale seems fine.

setup.py is the dependencies for the package itself, so `pytest` wouldn't be needed there.

chinandrew:
This git history is a bit messed up due to rendered file messiness on my end. Going to squash into a new branch and open a separate PR

PR #54, opened 2020-08-29
Refactor Python tests to use reference files

### LINKED ISSUE #39: Figure out how to save GeoDFs for testing
Have looked into this a little bit w/ the GeoPandas [docs](https://geopandas.org/io.html). Can't save to shapefile since it's a datetime, so would have to cast that col to a string. When I save to geojson, 5 out of 3233 geometries don't line up for some reason. When I plot them, they look malformed. Need to look more into this.

chinandrew:
Closes #39. Figured if we're coming up on a release, would be better to do this sooner than later.

Summary of changes:
- Add reference files for geodataframes to compare in tests. Previously were just comparing a few values and verifying DF shape and whether some columns were populated. Turns out GeoPackages seem to work over GeoJSON, and files are also smaller.
- Moved geometry to last column instead of penultimate column, which seems to be convention for these files.
- Added one test that uses an actual signal returned from the api just for good measure.


capnrefsmmat:
Oddly enough, I get two test failures that look like this:

```
>       pd.testing.assert_frame_equal(expected3, output3)

tests/covidcast/test_plotting.py:45: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pandas/_libs/testing.pyx:66: in pandas._libs.testing.assert_almost_equal
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   AssertionError: DataFrame.iloc[:, 0] are different
E   
E   DataFrame.iloc[:, 0] values are different (26.78571 %)
E   [left]:  [me, nh, de, dc, gu, mp, as, sc, ne, wa, nm, pr, sd, tx, ca, ky, oh, al, ga, wi, ar, or, pa, ms, co, ut, ok, tn, wv, ny, in, ks, nv, il, vt, ct, mt, mn, md, hi, az, ri, mo, nc, va, wy, la, mi, ma, id, fl, ak, vi, nj, nd, ia]
E   [right]: [ca, me, nh, de, dc, gu, mp, as, sc, ne, wa, nm, pr, sd, tx, ky, oh, al, ga, wi, ar, or, pa, ms, co, ut, ok, tn, wv, ny, in, ks, nv, il, vt, ct, mt, mn, md, hi, az, ri, mo, nc, va, wy, la, mi, ma, id, fl, ak, vi, nj, nd, ia]

pandas/_libs/testing.pyx:180: AssertionError
```

The data frames seem to match, except their rows are in a different order. My guess is that `merge` produces an ordering that depends on the ordering of the inputs, and maybe the shapefile ordering depends on dictionary ordering or something silly like that. There appears to be a `sort` argument to `merge` that would ensure the output is in a consistent ordering.

chinandrew:
> Oddly enough, I get two test failures that look like this:
> 
> ```
> >       pd.testing.assert_frame_equal(expected3, output3)
> 
> tests/covidcast/test_plotting.py:45: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
> pandas/_libs/testing.pyx:66: in pandas._libs.testing.assert_almost_equal
>     ???
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
> 
> >   ???
> E   AssertionError: DataFrame.iloc[:, 0] are different
> E   
> E   DataFrame.iloc[:, 0] values are different (26.78571 %)
> E   [left]:  [me, nh, de, dc, gu, mp, as, sc, ne, wa, nm, pr, sd, tx, ca, ky, oh, al, ga, wi, ar, or, pa, ms, co, ut, ok, tn, wv, ny, in, ks, nv, il, vt, ct, mt, mn, md, hi, az, ri, mo, nc, va, wy, la, mi, ma, id, fl, ak, vi, nj, nd, ia]
> E   [right]: [ca, me, nh, de, dc, gu, mp, as, sc, ne, wa, nm, pr, sd, tx, ky, oh, al, ga, wi, ar, or, pa, ms, co, ut, ok, tn, wv, ny, in, ks, nv, il, vt, ct, mt, mn, md, hi, az, ri, mo, nc, va, wy, la, mi, ma, id, fl, ak, vi, nj, nd, ia]
> 
> pandas/_libs/testing.pyx:180: AssertionError
> ```
> 
> The data frames seem to match, except their rows are in a different order. My guess is that `merge` produces an ordering that depends on the ordering of the inputs, and maybe the shapefile ordering depends on dictionary ordering or something silly like that. There appears to be a `sort` argument to `merge` that would ensure the output is in a consistent ordering.

That makes sense, I've changed it to sort on merge. Thanks for running the test. I'm a little curious why it doesn't work universally since according the the documentation, if `sort=False` (the default), then the sorting works as follows, and seems like it should still be deterministic.

```
    left: use only keys from left frame, similar to a SQL left outer join; preserve key order.
    right: use only keys from right frame, similar to a SQL right outer join; preserve key order.
    outer: use union of keys from both frames, similar to a SQL full outer join; sort keys lexicographically.
    inner: use intersection of keys from both frames, similar to a SQL inner join; preserve the order of the left keys.
```

capnrefsmmat:
That would suggest the ordering would come from the shapefiles -- I'm wondering if the shapefiles don't preserve order (or we have different version of the shapefile-parsing package or something). But the test passes now, so looks good.

PR #56, opened 2020-08-29
Further improve Python client docs

capnrefsmmat:
Some further improvement to the Sphinx docs before release.

@chinandrew, a couple questions for your review:

* I noticed that `plot_choropleth` had a type annotation saying it returned a `gpd.GeoDataFrame`, which is false, since we changed it to return the figure object. Is the type annotation checked by any of our tooling? I've never used Python type annotations, so I don't know how mypy fits in here.
* `plot_choropleth` pointed to the GeoSeries `plot` method, but I assume it actually uses the `GeoDataFrame` method, so I switched the links. Did I guess right?

### DIFF
--- a/Python-packages/covidcast-py/covidcast/plotting.py
+++ b/Python-packages/covidcast-py/covidcast/plotting.py
@@ -81,21 +81,20 @@ def get_geo_df(data: pd.DataFrame,
                join_type: str = "right") -> gpd.GeoDataFrame:
     """Augment a :py:func:`covidcast.signal` data frame with the shape of each geography.
 
-    This method takes in a pandas DataFrame object and returns a GeoDataFrame object from the
-    `GeoPandas package <https://geopandas.org/>`_.
-
-    Shapefiles are 1:5,000,000 scale and sourced from the `2019 US Census Cartographic Boundary
-    Files
-    <https://www.census.gov/geographies/mapping-files/time-series/geo/cartographic-boundary.html>`_.
+    This method takes in a pandas DataFrame object and returns a GeoDataFrame
+    object from the `GeoPandas package <https://geopandas.org/>`_. The
+    GeoDataFrame will contain the geographic shape corresponding to every row in
+    its ``geometry`` colummn; for example, a data frame of county-level signal
+    observations will be returned with the shape of every county.
 
     After detecting the geography type (only county and state are currently
-    supported) of the input, builds a GeoDataFrame that contains state and
-    geometry information from the Census for that geography type. By default, it
-    will take the signal data (left side) and geo data (right side) and right
-    join them, so all states/counties will always be present regardless whether
-    ``data`` contains values for those locations. ``left``, ``outer``, and
-    ``inner`` joins are also supported and can be selected with the
-    ``join_type`` argument.
+    supported) of the input, this function builds a GeoDataFrame that contains
+    state and geometry information from the Census for that geography type. By
+    default, it will take the signal data (left side) and geo data (right side)
+    and right join them, so all states/counties will always be present
+    regardless whether ``data`` contains values for those locations. ``left``,
+    ``outer``, and ``inner`` joins are also supported and can be selected with
+    the ``join_type`` argument.
 
     For right joins on counties, all counties without a signal value will be
     given the value of the megacounty (if present). Other joins will not use
@@ -103,8 +102,15 @@ def get_geo_df(data: pd.DataFrame,
     <https://cmu-delphi.github.io/delphi-epidata/api/covidcast_geography.html>`_
     for information about megacounties.
 
-    Default arguments for column names correspond to those used by
-    :py:func:`covidcast.signal`. Currently only supports counties and states.
+    By default, this function identifies the geography for each row of the input
+    data frame using its ``geo_value`` column, matching data frames returned by
+    :py:func:`covidcast.signal`, but the ``geo_value_col`` and ``geo_type_col``
+    arguments can be provided to match geographies for data frames with
+    different column names.
+
+    Geographic data is sourced from 1:5,000,000-scale shapefiles from the `2019
+    US Census Cartographic Boundary Files
+    <https://www.census.gov/geographies/mapping-files/time-series/geo/cartographic-boundary.html>`_.
 
     :param data: DataFrame of values and geographies
     :param geo_value_col: name of column containing values of interest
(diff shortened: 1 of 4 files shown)


Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -74,39 +79,44 @@ def get_geo_df(data: pd.DataFrame,
...
+    ``join_type`` argument.
+
+    For right joins on counties, all counties without a signal value will be
+    given the value of the megacounty (if present). Other joins will not use
+    megacounties. See the `geographic coding documentation
+    <https://cmu-delphi.github.io/delphi-epidata/api/covidcast_geography.html>`_
+    for information about megacounties.
+
+    Default arguments for column names correspond to those used by
+    :py:func:`covidcast.signal`. Currently only supports counties and states.
```
chinandrew, line 107:
```suggestion
    :py:func:`covidcast.signal`.
```

This is already mentioned above, though if we wanted to put it more front and center, could move it to the second sentence of the docstring.

chinandrew:
Cool, thanks for the edits. My sphinx is failing to build the html right now so I can't review it in the final rendered format, but the changes look good.

>    I noticed that `plot_choropleth` had a type annotation saying it returned a `gpd.GeoDataFrame`, which is false, since we changed it to return the figure object. Is the type annotation checked by any of our tooling? I've never used Python type >
annotations, so I don't know how mypy fits in here.

Hm yea that should be been caught. Let me look into why mypy didn't complain. Maybe it can't parse the plt.subplots for return type, but then it should consider it an `Any` and still raise something.


>    `plot_choropleth` pointed to the GeoSeries `plot` method, but I assume it actually uses the `GeoDataFrame` method, so I switched the links. Did I guess right?

Yup that's right, thanks.

Python-packages/covidcast-py/covidcast/plotting.py
capnrefsmmat: Good point; I've rearranged the docstring a bit more and made other edits while I was at it

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -31,19 +32,23 @@ def plot_choropleth(data: pd.DataFrame,
     For visual purposes, Alaska and Hawaii are moved the lower left corner of the contiguous US
```
chinandrew, line 32:
```suggestion
    For visual purposes, Alaska and Hawaii are moved to the lower left corner of the contiguous US
```

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -31,19 +32,23 @@ def plot_choropleth(data: pd.DataFrame,
...
-
-    :param data: DataFrame of values and geographies.
-    :param time_value: If multiple days of data are present in ``data``, map only values from this \
-        day. Defaults to plotting the most recent day of data in ``data``.
-    :param kwargs: Optional keyword arguments passed to plot().
+    By default, the `colormap
+    <https://matplotlib.org/tutorials/colors/colormaps.html>`_ used is
+    ``YlOrRd`` and is binned into the signal's historical mean value ± 3
+    standard deviations. Custom arguments can be passed in as ``kwargs`` for
+    customizability. These arguments will be past to the GeoPandas ``plot``
```
chinandrew, line 39:
```suggestion
    customizability. These arguments will be passed to the GeoPandas ``plot``
```

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -74,39 +79,51 @@ def get_geo_df(data: pd.DataFrame,
...
+    GeoDataFrame will contain the geographic shape corresponding to every row in
+    its ``geometry`` colummn; for example, a data frame of county-level signal
+    observations will be returned with the shape of every county.
+
+    After detecting the geography type (only county and state are currently
+    supported) of the input, this function builds a GeoDataFrame that contains
+    state and geometry information from the Census for that geography type. By
+    default, it will take the signal data (left side) and geo data (right side)
+    and right join them, so all states/counties will always be present
+    regardless whether ``data`` contains values for those locations. ``left``,
```
chinandrew, line 95:
```suggestion
    regardless of whether ``data`` contains values for those locations. ``left``,
```

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -74,39 +79,51 @@ def get_geo_df(data: pd.DataFrame,
...
+    data frame using its ``geo_value`` column, matching data frames returned by
+    :py:func:`covidcast.signal`, but the ``geo_value_col`` and ``geo_type_col``
+    arguments can be provided to match geographies for data frames with
+    different column names.
+
+    Geographic data is sourced from 1:5,000,000-scale shapefiles from the `2019
+    US Census Cartographic Boundary Files
+    <https://www.census.gov/geographies/mapping-files/time-series/geo/cartographic-boundary.html>`_.
 
     :param data: DataFrame of values and geographies
```
chinandrew, line 115:
```suggestion
    :param data: DataFrame of values and geographies.
```
Thought i had fixed these in a past PR but maybe got mixed up in merge conflict.

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -74,39 +79,51 @@ def get_geo_df(data: pd.DataFrame,
...
     :param geo_value_col: name of column containing values of interest
```
chinandrew, line 116:
```suggestion
    :param geo_value_col: Name of column containing values of interest.
```

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -74,39 +79,51 @@ def get_geo_df(data: pd.DataFrame,
...
     :param geo_type_col: name of column containing geography type
```
chinandrew, line 117:
```suggestion
    :param geo_type_col: Name of column containing geography type.
```

chinandrew:
Ok I think the issue with mypy is that it doesn't recognize a `Figure` or a `GeoDataFrame`, considers them both as type `Any`, and passes. We could add custom types, as well as much stricter linting that would cause this to error (mainly around disallowing `Any` in various forms), but don't know if that's worth it at the moment. I'll make an issue for it and we an revisit. In the meantime. I realized that theres a few config options I should add to it that should help with enforcement and general style. Will also ticket that - should be a quick one and we can adjust over time.

On documentation, I've got things rendering again and left a few comments. Do we want to be consistent with saying "data frame" vs "DataFrame"?

capnrefsmmat:
Thanks for the edits -- applied and will now merge.

> Do we want to be consistent with saying "data frame" vs "DataFrame"?

I would tend to use "data frame" except when I specifically want to refer to the pandas type, so I'd say "returns a data frame" but then "use DataFrame methods to...".

Issue #57, opened 2020-08-30
Adjust mypy settings to be more useful


chinandrew:
Right now mypy is fairly bare bones. We should add some flags to enforce behaviour regarding untyped defs and also having `None` not always return Optional.

Example of config file https://github.com/markkohdev/mypy-example/blob/master/mypy.ini

chinandrew:
Fixed in #62

PR #62, opened 2020-09-03
Add signal aggregation functions and update mypy

### LINKED ISSUE #50: Add tools to Python client to rearrange data for analysis
Python package version of #48
For example, suppose I want to do a regression where I predict cases using various other signals. I need to obtain the signals, join them together, maybe lag some variables, and produce a data frame with one row per location and time, containing all measurements at that location/time.

We should provide a function, name to be determined, that

    takes a bunch of signal data frames
    joins them together and names the value columns after the signals
    optionally lags some variables, e.g. if I want to predict cases with lagged signals
    returns one big data frame

Then we should put a few examples in the package docs showing how we can use this to conduct a data analysis.

chinandrew:
Closes #50 and #57

Add a function that takes in multiple signal DFs, lets you lag times, and joins them on the geo value and time.
- defaults to outer join, with join type arg optional if inner join desired
- assumes a single geo type across all dfs
- adds the data source, signal name, and index of that df in the input list (to avoid duplicate col names on join) to each column of the output and drops the original data source and signal columns. If there's a downstream use for them, we can keep.
- open to suggestions for better function names

Also updated some mypy configs, so it now does things like errors if a function is missing type annotations.

TODOs:
- [x] Support duplicate signal+data_source values. - This is now supported by indexing each df that gets joined together\
- [x] Have warning or error if you pass in multiple geo_types
- [x] Documentation in the function - Will add documentation for the website in a separate PR once this interface is finalized.
- [x] Additional test cases

Example usages
```
In [16]: test_input1                                                                                                                                                                                                                       
Out[16]: 
  geo_value        time  value geo_type signal data_source
0         a  2020-01-01      2    state      i           x
1         b  2020-01-01      4    state      i           x
2         c  2020-01-01      6    state      i           x
3         a  2020-01-02      8    state      i           x

In [17]: test_input2                                                                                                                                                                                                                       
Out[17]: 
  geo_value        time  value geo_type signal data_source
0         a  2020-01-01      1    state      j           y
1         b  2020-01-01      3    state      j           y
2         c  2020-01-01      5    state      j           y
3         d  2020-01-01      7    state      j           y

In [18]: test_input3                                                                                                                                                                                                                       
Out[18]: 
  geo_value        time  value geo_type signal data_source
0         b  2020-01-01    0.5    state      k           z
1         c  2020-01-01    1.5    state      k           z
2         d  2020-01-01    2.5    state      k           z
3         b  2020-01-02    3.5    state      k           z

In [19]: covidcast.aggregate_signals([test_input1, test_input2, test_input3], dt=[0, 0, 1])                                                                                                                                               
Out[19]: 
  geo_value        time  x_i_value  y_j_value  z_k_value geo_type
0         a  2020-01-01        2.0        1.0        NaN    state
1         b  2020-01-01        4.0        3.0        NaN    state
2         c  2020-01-01        6.0        5.0        NaN    state
3         d  2020-01-01        NaN        7.0        NaN    state
4         a  2020-01-02        8.0        NaN        NaN    state
5         b  2020-01-02        NaN        NaN        0.5    state
6         c  2020-01-02        NaN        NaN        1.5    state
7         d  2020-01-02        NaN        NaN        2.5    state
8         b  2020-01-03        NaN        NaN        3.5    state

```

If you have extra columns, it looks like this:
```
In [27]: test_input1                                                                                                                                                                                                                       
Out[27]: 
  geo_value        time  value geo_type signal data_source  extra_col
0         a  2020-01-01      2    state      i           x         -1
1         b  2020-01-01      4    state      i           x         -1
2         c  2020-01-01      6    state      i           x         -1
3         a  2020-01-02      8    state      i           x         -1

In [28]: test_input2                                                                                                                                                                                                                       
Out[28]: 
  geo_value        time  value geo_type signal data_source  extra_col
0         a  2020-01-01      1    state      j           y         -2
1         b  2020-01-01      3    state      j           y         -2
2         c  2020-01-01      5    state      j           y         -2
3         d  2020-01-01      7    state      j           y         -2

In [29]: test_input3                                                                                                                                                                                                                       
Out[29]: 
  geo_value        time  value geo_type signal data_source  another_extra_col
0         b  2020-01-01    0.5    state      k           z                 -3
1         c  2020-01-01    1.5    state      k           z                 -3
2         d  2020-01-01    2.5    state      k           z                 -3
3         b  2020-01-02    3.5    state      k           z                 -3

In [30]: covidcast.aggregate_signals([test_input1, test_input2, test_input3], dt=[0, 0, 1])                                                                                                                                               
Out[30]: 
  geo_value        time  x_i_value  x_i_extra_col  y_j_value  y_j_extra_col  z_k_value  z_k_another_extra_col geo_type
0         a  2020-01-01        2.0           -1.0        1.0           -2.0        NaN                    NaN    state
1         b  2020-01-01        4.0           -1.0        3.0           -2.0        NaN                    NaN    state
2         c  2020-01-01        6.0           -1.0        5.0           -2.0        NaN                    NaN    state
3         d  2020-01-01        NaN            NaN        7.0           -2.0        NaN                    NaN    state
4         a  2020-01-02        8.0           -1.0        NaN            NaN        NaN                    NaN    state
5         b  2020-01-02        NaN            NaN        NaN            NaN        0.5                   -3.0    state
6         c  2020-01-02        NaN            NaN        NaN            NaN        1.5                   -3.0    state
7         d  2020-01-02        NaN            NaN        NaN            NaN        2.5                   -3.0    state
8         b  2020-01-03        NaN            NaN        NaN            NaN        3.5                   -3.0    state
```


Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -253,6 +254,49 @@ def metadata() -> pd.DataFrame:
...
+    :param signals: List of DataFrames to join.
+    :param dt: List of lags in days for each of the input DataFrames in ``signals``.
+      Defaults to `None`.
+    :param join_type: Type of join to be done between the DataFrames in ``signals``.
+      Defaults to `outer`.
+    :return: DataFrame of aggregated signals.
+    """
+    join_cols = ["time", "geo_value"]
+    if dt is not None and len(dt) != len(signals):
+        raise ValueError("Length of `lags` must be same as length of `signals`")
```
capnrefsmmat, line 281:
```suggestion
        raise ValueError("Length of `dt` must be same as length of `signals`")
```

Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -253,6 +254,49 @@ def metadata() -> pd.DataFrame:
...
+    dt_dfs = []
+    first_geo_type = _detect_metadata(signals[0])[2]
+    for i, (df, lag) in enumerate(zip(signals, dt)):
+        df_c = df.copy()  # make a copy so we don't modify originals
+        source, sig_type, geo_type = _detect_metadata(df_c)
+        if geo_type != first_geo_type:
+            raise ValueError("Multiple geo_types detected. Only a single geo_type is supported.")
```
capnrefsmmat, line 288:
```suggestion
            raise ValueError("Multiple geo_types detected. All signals must have the same geo_type to be aggregated.")
```

capnrefsmmat (commented):
Overall looks good. Just two minor error message tweaks.

But, a question. We added the source and signal columns to data frames because eventually we want to be able to fetch multiple signals at once with some kind of `covidcast.signals()` function. My draft in the R package produces one big data frame with the signals.

In the R package, then, an aggregation function like this would need to take the data from "long" format (one row per observation) to "wide" format (one row per location/time, with multiple columns for multiple signals).

What do you think is the best way to support working with multiple signals at once in the Python package? Have them in one big data frame, or in a list of data frames like `aggregate_signals` takes?

The answer can be different between R and Python; in R, the conversion from long to wide would be trivial with `pivot_wider`, whereas in Python maybe we don't expect people to work with the data frames in the same way.

chinandrew:
> Overall looks good. Just two minor error message tweaks.
> 
> But, a question. We added the source and signal columns to data frames because eventually we want to be able to fetch multiple signals at once with some kind of `covidcast.signals()` function. My draft in the R package produces one big data frame with the signals.
> 
> In the R package, then, an aggregation function like this would need to take the data from "long" format (one row per observation) to "wide" format (one row per location/time, with multiple columns for multiple signals).
> 
> What do you think is the best way to support working with multiple signals at once in the Python package? Have them in one big data frame, or in a list of data frames like `aggregate_signals` takes?
> 
> The answer can be different between R and Python; in R, the conversion from long to wide would be trivial with `pivot_wider`, whereas in Python maybe we don't expect people to work with the data frames in the same way.

Just to make sure I'm understanding this, is the question whether or not the input should be

a) 1 DF like
```
geo,signal,value
x,a,1
y,a,2
x,b,3
y,b,4
```
vs b) 2 DFs like 
```
geo,signal,value
x,a,1
y,a,2

geo,signal,value
x,b,3
y,b,4
```
?

I think the arguments for (a) as the return type for some multiple signal function are
- it's a DF returned, which people should be comfortable using and is simple to filter by
- if you specify multiple times and multiple states, they all come in one DF, so why should multiple signals be any different?
- same interface as R package, which is nice but not necessary as you mention

Arguments for (b)
- If I'm pulling 5 signals, doesn't seem too unreasonable to expect a list of 5 signals
- Since the multi-signal function doesnt exist yet, it seems kinda odd to have a function that takes in the concatenation of things returned by a function vs just a list of the things returned.
- to go from (b) to (a) is basically one line of code with `pd.concat(df_list)`

One way to think about this is
1. Philosophically, what do we consider the basic unit within this package, which right now seems to be the signal DF that gets returned by signal(). Is it a collection of values across time/geo for single data source+type (as it is defined now)? Or can it accommodate multiple sources? If it's the latter, I think that's a strong case for using concatenated DFs. 
2. If the former, is multiple sources/types it's own object with it's own unique uses, or is it simply a convenient collection of the base signals? Do some functions expect a "multi-signal" object, or are they really just expecting multiple signals? If the latter, I think that suggests a list of DFs while the former suggests concatenating them. 

My preference for (1), without knowing much about end users, is that the signal as currently implemented seems fine as the basic unit, and if you have 'fb-survey' and 'doctor-visits' values, that's two signals. The fact that we use the term "multiple signals" vs "multiple source/type" support also seems to agree with this. For (2), I don't see a strong reason to group things together into one object. If you want to compare two data sources, you're comparing two signals, not one "multi-signal", and therefore you should receive two signals and pass in two signals to whatever function you want to use. Therefore I net out at using a list of DFs vs one giant concatenated one. 

Don't feel super strongly about this so open to discussion. Could be persuaded that a single DF provides sufficient added usability or potentially that people expect it. (Also, my actual preference as a user would be that everything is JSON/dicts, but I get that's not right for our audience.)

chinandrew:
Also regardless of the multi-signal implementation, another option for the aggregation return type is the pandas multi-index DF. The few times I've used a multi-index DF I found them a bit weird to use, though I don't have a whole lot of experience so it's possible pandas superusers like them. Example below
```
In [34]: covid1 = covidcast.signal('fb-survey','smoothed_cli', start_day = date(2020,8,3), end_day = date(2020,8,4), geo_type = 'state')                                                                                                   

In [35]: covid2 = covidcast.signal('fb-survey','raw_cli', start_day = date(2020,8,3), end_day = date(2020,8,4), geo_type = 'state')                                                                                                        

In [36]: covid3 = covidcast.signal('doctor-visits','smoothed_cli', start_day = date(2020,8,3), end_day = date(2020,8,4), geo_type = 'state')                                                                                               

In [37]: c = pd.concat([covid1,covid2,covid3])                                                                                                                                                                                             

In [38]: c.pivot(index = ["geo_value", "time_value"], columns =  ["data_source", "signal"] ,values = "value")                                                                                                                              
Out[38]: 
data_source             fb-survey           doctor-visits
signal               smoothed_cli   raw_cli  smoothed_cli
geo_value time_value                                     
ak        2020-08-03     0.780718  0.512821      1.323974
          2020-08-04     0.770648  0.000000      1.555860
al        2020-08-03     1.046256  0.962165     10.927861
          2020-08-04     1.074746  0.994775      9.772409
ar        2020-08-03     0.803524  0.801964      8.255912
...                           ...       ...           ...
wi        2020-08-04     0.448432  0.459794      3.733620
wv        2020-08-03     0.466872  0.181361      2.555040
          2020-08-04     0.558640  0.831694      2.490196
wy        2020-08-03     0.488934  0.644777      2.075715
          2020-08-04     0.385025  0.213306      2.130241
```

capnrefsmmat:
I think I can buy your arguments here. In particular, returning a list makes it very easy to pass the results to the mapping functions, e.g. to make a grid of maps of multiple signals. If we returned a data frame of multiple signals, all mapping functions would need to change to handle this case.

I just pushed a few changes. Most importantly, it now works with data frames returned by `signal()` (whoops) and has documentation. If the changes look good to you, I'm ready to merge.

Python-packages/covidcast-py/docs/getting_started.rst
```diff
@@ -63,6 +63,40 @@ such as the state of Pennsylvania for the month of May 2020:
...
+2 2020-05-01        pa    NaN    62.0
+3 2020-05-02        mo    NaN    14.0
+4 2020-05-02        ny    NaN   232.0
+
+The resulting ``cases_v_deaths`` data frame contains one row per location per
+day. The death value is the number of deaths on that day; the cases value is the
+number of cases *7 days prior*, matching the ``dt`` provided to
+:py:func:`covidcast.aggregate_signals`. The case values shown above are ``NaN``
+because the input data frame did not contain case numbers for late April; rows
+for mid-May contain both case and death values.
```
chinandrew, line 95: Should we just use an example that contains (at least partially) non-nan values here?

capnrefsmmat: Fair point; will change

PR #64, opened 2020-09-05
Add Geo functions that replicate R package interface

### LINKED ISSUE #55: Provide functions to look up FIPS codes, CBSA IDs, etc.
As of #29, the R package now has functions `fips_to_name`, `name_to_fips`, etc. for working with FIPS codes and CBSA IDs. See [the documentation](https://cmu-delphi.github.io/covidcast/covidcastR/reference/name_to_fips.html).

This would be very handy for the Python package, because it's pretty inconvenient to have to figure out the right code for a county or MSA to be able to query data for it. I suspect the Python versions can work very similarly to the R versions, using the Census shapefiles (or other Census data, such as the Census data frames produced for the R package) to get names and IDs.

chinandrew:
This closes #55
Summary of changes:
- Add fips_to_name, cbsa_to_name, abbr_to_name, name_to_abbr, name_to_cbsa, name_to_fips functions along with helps which replicate R package equivalents.
- Convert rda reference files that R package uses into csv files and add to packaging instructions.
- update pytest to ignore user warnings in the output.

Notes:
I've tried to replicate the R package functionality and interfaces as closely as possible, with some minor differences:
1. I'm using lists of dictionaries instead of list of lists, but general behavior remains the similar:
    - if `ties_method = first`, returns a dict of a n str:str key value pairs for n inputs (I need to finish implementing this. TODO below). 
    - if `ties_method = all`, returns a list of dicts of str:list (even if its only 1 dict or length 1 list). Can't do multiple str:str for a given region like R (e.g. Miami County 3 times in the example below) because we can't have duplicate keys in a dict. 

2. using `re` package for regular expressions, which doesn't have a perl option
3. printing the non unique warning instead of using `warnings`, since a warning implies something noncritical went wrong, while the non unique case is simply "Everyone thing as intended, there's just some more info you should know"

R
```
name_to_fips("Miami")
#> Warning: Some inputs were not uniquely matched; returning only the first match in each case.
#> Miami-Dade County 
#>           "12086" 
name_to_fips("Miami", ties_method = "all")
#> [[1]]
#> Miami-Dade County      Miami County      Miami County      Miami County 
#>           "12086"           "18103"           "20121"           "39109" 
#> 
name_to_fips(c("Allegheny", "Miami", "New "), ties_method = "all")
#> [[1]]
#> Allegheny County 
#>          "42003" 
#> 
#> [[2]]
#> Miami-Dade County      Miami County      Miami County      Miami County 
#>           "12086"           "18103"           "20121"           "39109" 
#> 
#> [[3]]
#>   New Haven County  New London County  New Castle County  New Madrid County 
#>            "09009"            "09011"            "10003"            "29143" 
#>    New York County New Hanover County    New Kent County 
#>            "36061"            "37129"            "51127" 

```

Python
```
In [1]: from covidcast import geography                                                                                                                                                                                                    

In [2]: geography.name_to_fips("Miami")                                                                                                                                                                                                    
Out[2]: {'Miami-Dade County': '12086'}

In [3]: geography.name_to_fips("Miami", ties_method="all")                                                                                                                                                                                 
Out[3]: [{'Miami-Dade County': ['12086'], 'Miami County': ['18103', '20121', '39109']}]

In [4]: geography.name_to_fips(["Allegheny", "Miami", "New "], ties_method="all")                                                                                                                                                          
Out[4]: 
[{'Allegheny County': ['42003']},
 {'Miami-Dade County': ['12086'], 'Miami County': ['18103', '20121', '39109']},
 {'New Haven County': ['09009'],
  'New London County': ['09011'],
  'New Castle County': ['10003'],
  'New Madrid County': ['29143'],
  'New York County': ['36061'],
  'New Hanover County': ['37129'],
  'New Kent County': ['51127']}]


```

TODO
- [x] Handle the multi-input case e.g. `geography.name_to_fips(["Allegheny", "Miami", "New "])`. Currently it'll only return the first value `{'Allegheny County': '42003'}` instead of all three `{"Allegheny County": "42003",  "Miami-Dade County": "12086",  "New Haven County": "09009"}` - This is handled via a python 3.6+ (the version we require in setup.py) specific detail where the insertion order of lists is retained.
- [x] Finish tests for remaining methods, add additional tests for state filtering and `fixed`/`ignore_case` params
- [x] (done-ish) Try coming up with some neater `_lookup` implementation that doesn't involve so many nested statements. Low pri style thing but it's late and I'm probably overlooking something easier.
- [x] Add warning for non unique value scenario
- [x] Documentation

Also, this line was in the R package, but the comment seems to disagree with what the code is doing (removing the states). Is removing states the intended behavior?
```
  # Leave states in county_census (so we can find state fips)
  df = county_census # %>% dplyr::filter(COUNTY != 0)

  {"Miami-Dade County": ["12086"],"Miami County": ["18103", "20121", "39109"]}

```


ryantibs:
@chinandrew If you look carefully at that line, it's just defining `df` to be `county_census`, because the comment symbol `#` comes before any of the filtering.  

(It's a little weird to read but I did it this way to remind myself that previously I wanted to filter out states, and now I decided against it, so you can use the FIPS codes to search states too.)

ryantibs:
(Oh and, looks like great work, as usual!  Not being the Python expert, I'll let @capnrefsmmat review ... but we should really start finding help for reviewing this PRs, because Alex is doing pretty much all of them.)

capnrefsmmat:
For the multi-input case, I expect you want to be able to do it to an entire data frame column at once, so if it can work on a Series that'd be great.

Python-packages/covidcast-py/covidcast/geography.py
```diff
@@ -0,0 +1,63 @@
...
+    return _lookup(name, df.NAME, df.CBSA, ignore_case, fixed, ties_method)
+
+
+def name_to_fips(name, ignore_case=False, fixed=False, ties_method="first", state=None):
+    df = COUNTY_CENSUS.loc[COUNTY_CENSUS.STNAME == abbr_to_name(state)[state].upper()] if state \
+        else COUNTY_CENSUS
+    return _lookup(name, df.CTYNAME, df.FIPS, ignore_case, fixed, ties_method)
+
+
+def _lookup(key, keys, values, ignore_case=False, fixed=False, ties_method="first"):
```
capnrefsmmat, line 47: In what cases would we *not* want to ignore case? I expect most US county/MSA names are not particularly case sensitive...

chinandrew: I meant to ask that as well. I've pulled the defaults from the R code, but also thought going True might be better. I thought one argument against might be the edge case where someone sets `fixed=TRUE` but `ignore_case=TRUE` by default might lead to confusion, but I think the right implementation there is to have fixed override ignore_case with an info message. If we do want to go True, we should do in both R and Python.

chinandrew:
> @chinandrew If you look carefully at that line, it's just defining `df` to be `county_census`, because the comment symbol `#` comes before any of the filtering.
> 
> (It's a little weird to read but I did it this way to remind myself that previously I wanted to filter out states, and now I decided against it, so you can use the FIPS codes to search states too.)

Got it, thanks for clarifying. Completely missed the comment since my Python IDE doesn't syntax highlight R code :(

chinandrew:
> For the multi-input case, I expect you want to be able to do it to an entire data frame column at once, so if it can work on a Series that'd be great.

Yup that's supported

```
In [4]: series = pd.Series(["Allegheny", "Miami", "New "])                                                                                                                                                                                 

In [5]: series                                                                                                                                                                                                                             
Out[5]: 
0    Allegheny
1        Miami
2         New 
dtype: object

In [6]: geography.name_to_fips(series, ties_method="all")                                                                                                                                                                                  
Out[6]: 
[{'Allegheny County': ['42003']},
 {'Miami-Dade County': ['12086'], 'Miami County': ['18103', '20121', '39109']},
 {'New Haven County': ['09009'],
  'New London County': ['09011'],
  'New Castle County': ['10003'],
  'New Madrid County': ['29143'],
  'New Hampshire': ['33000'],
  'New Jersey': ['34000'],
  'New Mexico': ['35000'],
  'New York': ['36000'],
  'New York County': ['36061'],
  'New Hanover County': ['37129'],
  'New Kent County': ['51127']}]
```

capnrefsmmat:
So I started writing documentation for the manual and realized an ergonomic issue here. Basically, I imagine there are two common use-cases for these functions:

Case 1. I know what counties I want to query but not their FIPS codes, so I do

```python
counties = covidcast.name_to_fips(["Allegheny", "Los Angeles", "Miami-Dade"])

d = covidcast.signal("doctor-visits", "smoothed_cli",
                     start_day=date(2020, 5, 1), end_day=date(2020, 5, 1),
                     geo_values=counties.values())
```

Note how I had to use `counties.values()`, since `name_to_fips` returns a dictionary.

Case 2. I want to give human-readable names to the data returned from `signal()`, e.g. so that my plots have good legends. I'd like to do

```python
d = covidcast.signal(...)
d.geo_value = covidcast.fips_to_name(d.geo_value)
```

but I can't, because it's a dictionary. Of course, maybe I'm doing it the wrong way here. In R I'd change the values in the data frame because that's how ggplot2 would determine the line legends if I plotted a line for each county. That's why the R functions return a vector of the same length as the input. But maybe the easiest way to do it in matplotlib would make use of the dictionaries?

capnrefsmmat:
I've just pushed the draft docs so you can see the example I was trying out.

chinandrew:
> So I started writing documentation for the manual and realized an ergonomic issue here. Basically, I imagine there are two common use-cases for these functions:
> 
> Case 1. I know what counties I want to query but not their FIPS codes, so I do
> 
> ```python
> counties = covidcast.name_to_fips(["Allegheny", "Los Angeles", "Miami-Dade"])
> 
> d = covidcast.signal("doctor-visits", "smoothed_cli",
>                      start_day=date(2020, 5, 1), end_day=date(2020, 5, 1),
>                      geo_values=counties.values())
> ```
> 
> Note how I had to use `counties.values()`, since `name_to_fips` returns a dictionary.
> 
> Case 2. I want to give human-readable names to the data returned from `signal()`, e.g. so that my plots have good legends. I'd like to do
> 
> ```python
> d = covidcast.signal(...)
> d.geo_value = covidcast.fips_to_name(d.geo_value)
> ```
> 
> but I can't, because it's a dictionary. Of course, maybe I'm doing it the wrong way here. In R I'd change the values in the data frame because that's how ggplot2 would determine the line legends if I plotted a line for each county. That's why the R functions return a vector of the same length as the input. But maybe the easiest way to do it in matplotlib would make use of the dictionaries?

That use case makes sense. I wasn't sure what best analogy to the R lists were given they exhibit both behaviours, and I think using a list makes the most sense. 
```
> x = name_to_fips("Autauga CountY", ignore.case = TRUE)
> x["Autauga County"]
Autauga County 
       "01001" 
> x[1]
Autauga County 
       "01001" 
``` 
The current implementation would require something like 
```
mapping = [fips_to_name(d.geo_value)]
d.geo_value = [mapping[i] for i in d.geo_value]
``` 
which is clunkier. 


I can make  `ties_method="first"` case just return a list. The `ties_method="all"` is a bit trickier, but a list of lists might be reasonable as well.  Should be a quick change, thanks for calling it out.

capnrefsmmat:
> How do you envision people end up using a return like this? Expand into multiple rows, manual inspection, something else?

I'm not sure. I don't think `ties_method="all"` makes much sense in the vectorized case -- it only makes sense when you're trying to look up one county and need to find the one you want.

I think the vectorized output  for `fips_to_name` makes sense, since FIPS codes are unique and you'll often get a column full of them from `covidcast.signal()`, but perhaps `name_to_fips` should retain its current output style. That suits the case where you're quickly looking things up. In the example I gave,

```python
counties = covidcast.name_to_fips(["Allegheny", "Los Angeles", "Miami-Dade"])

d = covidcast.signal("doctor-visits", "smoothed_cli",
                     start_day=date(2020, 5, 1), end_day=date(2020, 5, 1),
                     geo_values=counties.values())
```

I actually had to carefully choose three counties that didn't have ties -- so in practice, it's probably more important to handle ties well than it is to avoid having to use `.values()`.

Does that make sense, or is there a use case for `name_to_fips` that I'm missing?

chinandrew:
> ```python
> counties = covidcast.name_to_fips(["Allegheny", "Los Angeles", "Miami-Dade"])
> 
> d = covidcast.signal("doctor-visits", "smoothed_cli",
>                      start_day=date(2020, 5, 1), end_day=date(2020, 5, 1),
>                      geo_values=counties.values())
> ```

Agreed on the ties handling part. I think a list works here for the "first" ties as follows:

> ```python
> counties = covidcast.name_to_fips(["Allegheny", "Los Angeles", "Miami-Dade"])
> # returns ["42003", "06037", "12086"]     
> d = covidcast.signal("doctor-visits", "smoothed_cli",
>                      start_day=date(2020, 5, 1), end_day=date(2020, 5, 1),
>                      geo_values=counties)
> ```

And in the ties case, a list of lists would need to be flattened, while a list of dicts would also need some weird comprehension since each dict may have multiple keys that each have their own list.

E.g. with all ties and ["Allegheny", "Miami", "New "] input returns:
``
            [
                {"Allegheny County": ["42003"]},
                {"Miami-Dade County": ["12086"],
                 "Miami County": ["18103", "20121", "39109"]},
                {"New Haven County": ["09009"], "New London County": ["09011"],
                 "New Castle County": ["10003"], "New Madrid County": ["29143"],
                 "New York County": ["36061"], "New Hanover County": ["37129"],
                 "New Kent County": ["51127"], "New Hampshire": ["33000"],
                 "New Jersey": ["34000"], "New Mexico": ["35000"],
                 "New York": ["36000"]}
            ]
```
 I think the list of lists is the easiest to parse, but you're unable determine (in both the first and all ties case) if the return values were from one name with multiple fips or multiple names (e.g. the miami data county and miami county example). We could also just add an argument for return_type that lets people choose.


I had also been operating under the assumption that we wanted the return values to all be similar, but that could be challenged too. For example, state name to state abbr not returning a dict seems fine.

**Comment by chinandrew** [comment]
Ok how about this implementation (what's currently pushed).


- If `ties_method = "first"`, we return a 1D list of values of length equal to the input. e.g. `name_to_fips(["Allegheny", "Miami", "New "])` returns `["42003", "12086", "09009"]`. This means:
    - You can use the return value to assign to a column or pass into the signal function as described above, because we don't have to worry about dict keys not being duplicates. e.g. `name_to_fips(["Allegheny", "Miami", "Allegheny "])` returns `["42003", "12086", "42003"]` and not `{"Allegheny":"42003", "Miami":"12086"}`
    - You will always have a populated list, so even if you have no matches you will get a Truthy list back. e.g. `name_to_fips(["aliuyeb3daed", "aiulfbalifu"])` returns `[None, None]` and `bool([None, None])==True`. This doesn't bother me too much if it reduces complexity elsewhere. To get actual truth value you'd run something like `any(i for i in returned_list)`
    - You cannot tell what the matched key was. e.g. `name_to_fips("Miami")` returns `"12086"` which corresponds to Miami-Dade, but you have no way of knowing that. This is the main downside in my mind, but you can figure this out with `ties_method="all"`.

- If `ties_method="all"`, you just get the full list of dicts that was the original implementation
    - Like the `"first"` case, output list length equals input list length. If you have an empty dict returned, it still shows up, so `name_to_fips(["Alle12321gheny", "M13ziamiu"]` returns `[{}, {}]`. 

    - `name_to_fips(["Allegheny", "Miami", "New "])`  returns
```          
            [
                {"Allegheny County": ["42003"]},
                {"Miami-Dade County": ["12086"],
                 "Miami County": ["18103", "20121", "39109"]},
                {"New Haven County": ["09009"], "New London County": ["09011"],
                 "New Castle County": ["10003"], "New Madrid County": ["29143"],
                 "New York County": ["36061"], "New Hanover County": ["37129"],
                 "New Kent County": ["51127"], "New Hampshire": ["33000"],
                 "New Jersey": ["34000"], "New Mexico": ["35000"],
                 "New York": ["36000"]}
            ]
```

- In both cases you get a list returned 

In addition, we can add (probably in a separate PR for cleanliness), the `return_type` argument that lets them select list or dict. 
- list would be the above returns
-  for "first", dict would return the original dict implementation: `{"Allegheny County": "42003", "Miami-Dade County": "12086", "New Haven County": "09009"}` 
- for "all", dict would return something like this :

```          
{
"Allegheny":
                {"Allegheny County": ["42003"]},
"Miami":
                {"Miami-Dade County": ["12086"],
                 "Miami County": ["18103", "20121", "39109"]},
"New ":
                {"New Haven County": ["09009"], "New London County": ["09011"],
                 "New Castle County": ["10003"], "New Madrid County": ["29143"],
                 "New York County": ["36061"], "New Hanover County": ["37129"],
                 "New Kent County": ["51127"], "New Hampshire": ["33000"],
                 "New Jersey": ["34000"], "New Mexico": ["35000"],
                 "New York": ["36000"]}
}
```


Would that cover all our bases?

**Comment by capnrefsmmat** [comment]
Yes, I've just played with the new method, and I like how easy it is to use:

```python
foo = covidcast.signal("fb-survey", "raw_cli", geo_type="msa", start_day=date(2020, 9, 9))
foo.geo_value = covidcast.cbsa_to_name(foo.geo_value)
foo.head()
```

Getting all ties seems to work well too.

I think the only things remaining are to update the docstrings and examples to match. Oh, and any reason the warning message is done with `print()` instead of a warning?

chinandrew:
> I think the only things remaining are to update the docstrings and examples to match. 

Will work on docstrings/examples tonight.

> Oh, and any reason the warning message is done with `print()` instead of a warning?
It was originally a warning, and then I reasoned warnings were for alerting someone of some undesirable/unintended behaviour that occurs but is not fatal, e.g. feature deprecation, falling back to some other code (maybe a matplotlib backend), etc. In this case, the function with "first" will execute exactly as expected with no errors, and so we simply want to let the user know there's some other options. That being said, I can see an argument which says that despite "first" being explicit behaviour, it's still a default argument and what we're saying is "hey we tried to get the corresponding name/code, but there were unexpected ties so we defaulted to the first one." I can buy that, I'll change it back to a warning.

chinandrew:
Updated warnings, docstrings, and your examples in the sphinx docs.

PR #68, opened 2020-09-08
Add HRR plots to Python package

### LINKED ISSUE #42: Add HRR maps in Python package
Currently geo/plotting functions  only support state and county. Add support for MSAs as well. Blocked by #40

chinandrew:
This closes #42 and closes #42 

Summary of changes:
- Add [HRR shapefiles](https://data.cms.gov/widgets/ia25-mrsk). I als found them [here](https://atlasdata.dartmouth.edu/downloads/supplemental#boundaries) but went with the .gov site.
- Add method for plotting HRRs and tests
- Add example to documentation

Notes:
I visually inspected a few regions in a plot to see if it was plotting right since neither the R package or the website (as far as I can tell) have HRR plots yet, so don't have a solid reference image.

Screenshot from 2020-09-08 00-08-20


capnrefsmmat:
Hm. Is it just me, or in your example map, are the HRRs shifted slightly north compared to the base land mass layer?

chinandrew:
> Hm. Is it just me, or in your example map, are the HRRs shifted slightly north compared to the base land mass layer?
> 
> I wonder if the CMS shapefiles have a projection problem.

Oh that does look to be the case, good catch. Weird, I'll look into it tonight.

chinandrew:
Ok so this is a bit annoying. The CMS files are shifted for some unknown reason, maybe something on their export. The Dartmouth files aren't shifted, but they're unprojected and have alaska and hawaii pre-moved  to slightly different locations, which causes problems:


Screenshot from 2020-09-09 00-25-46

Screenshot from 2020-09-09 00-23-30

Either way we'd need to do something custom. 

I've done the easy thing for now and just shifted the coordinates on the CMS file. I tried to grab centroids for the rhode island, alaska, and hawaii polgyons (which have 1 HRR for the whole state) from the state and the HRR shapefiles to get the diff, but it wasn't consistent and none were super accurate. Probably some combination of weird coordinates + different resolutions. I ended up eyeballing it and got pretty close. One thing I noticed was that it isn't exactly a shift, but there's also some weird scaling thing going on also. The top of the map if you zoom way in is a bit too high, and the bottom of the map is a bit too low. You need to zoom wayyyy in to see though, far more than anyone looking at HRRs would ever need.

Figure_2

Up by WA
Screenshot from 2020-09-09 01-38-24
down by AZ (not the same scale, just some arbitrary zoom via matplotlib gui
Screenshot from 2020-09-09 01-38-56


I think this is fine for now, let me know if you agree. If not, I can spend some time looking into the Dartmouth files. Work there would be translate/scaling AK and HI into their "normal" positions and then having the function transform them back into the desired ones.

capnrefsmmat:
Weird. I wonder if the CMS shapefiles are lying about being WGS84 and instead have some funky projection. (I dunno what other lat/lon coordinate system people would use that is noticeably different, though... usually the other reference systems are different on the order of a few meters.)

I think your rescaling is close enough. Can you file a follow-up issue to figure out a better solution? That way we'll remember to look at this again, even if it's not a high priority for now. Then I'm happy to merge this.

chinandrew:
> Weird. I wonder if the CMS shapefiles are lying about being WGS84 and instead have some funky projection. (I dunno what other lat/lon coordinate system people would use that is noticeably different, though... usually the other reference systems are different on the order of a few meters.)
> 
> I think your rescaling is close enough. Can you file a follow-up issue to figure out a better solution? That way we'll remember to look at this again, even if it's not a high priority for now. Then I'm happy to merge this.

Added #71

Issue #44, opened 2020-08-27
Test plotting functions


chinandrew:
Come up with some way to test the plotting functions. Testing plots sucks since everything renders differently. Long thread on it in #27.

PR #69, opened 2020-09-08
[Low Pri] Add plotting tests which run in CI.

chinandrew:
This helps #44 but does not necessarily close. Ideally we want platform independent tests that can work locally, but trying to do this in #27 caused a bunch of pain (see that PR for more info). Opening this PR as a stop gap since it's better than nothing and hopefully prevent unintentional breaking changes. Downside would be if someone who isn't on Linux makes an intentional breaking change, they couldn't easily fix the test. In that case, they'd probably have to just skip/remove it until it could be added.  Could also do platform specific tests, each with their own reference files, but that's a bit more annoying.

Summary of changes:
- Add a pytest decorator so the plot tests only run on linux (which is what the CI uses and can pass on). Eventually may want to make this more specific to just CI using a pytest mark, but I think this works for now (don't see a problem with everyone on linux running tests).
- Mock out the network calls from `_signal_metadata()`

Running this on a nonlinux platform will look like this, where `s`=skipped means the test is not run.
```
Python-packages/covidcast-py/tests/covidcast/test_plotting.py s....                                      [100%]
```
On Linux it will be the standard run
```
Python-packages/covidcast-py/tests/covidcast/test_plotting.py .....                                      [100%]
```
CI:
Screenshot from 2020-09-08 01-07-24

TODO:
- [x] add more tests for more plots


chinandrew:
Added code that fixes #77

PR #72, opened 2020-09-10
Fix time_value bug on Python choropleth

chinandrew:
Turns out 
```
data.loc[data.time_value == day_to_plot, :]
```
doesn't work because `day_to_plot` is a datetime.date object and the dataframe column is a pandas timestamp object, so instead we need to be doing this

```
data.loc[data.time_value == pd.to_datetime(day_to_plot), :]
```

Otherwise an empty dataframe gets returned and the function errors downstream. Added a basic test that just runs the function and will fail if it errors.

Guess it was a case that was never manually tested and we didn't have tests around plotting :(.

### DIFF
--- a/Python-packages/covidcast-py/tests/covidcast/test_plotting.py
+++ b/Python-packages/covidcast-py/tests/covidcast/test_plotting.py
@@ -18,6 +18,7 @@ NON_GEOMETRY_COLS = ["geo_value", "time_value", "direction", "issue", "lag", "va
 
 
 def test_plot_choropleth():
+    # see PR #72
     test_county = pd.read_csv(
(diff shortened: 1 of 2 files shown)


Python-packages/covidcast-py/tests/covidcast/test_plotting.py
```diff
@@ -16,6 +17,15 @@
                      "sample_size", "geo_type", "data_source", "signal", "state_fips"]
 
 
+def test_plot_choropleth():
```
capnrefsmmat, line 20:
```suggestion
def test_plot_choropleth():
    # see PR #72
```

PR #82, opened 2020-09-16
Add animation function for Python function

### LINKED ISSUE #46: Animate signals over time
Given  a signal over multiple dates, return an animation (maybe using [matplotlib animation](https://matplotlib.org/3.1.1/api/animation_api.html))

chinandrew:
Closes #46 

Summary of changes
- Added a method that takes in a signal df and saves out an mp4 animation. Gifs worked but didn't look great (See #46 for more info). Takes in args for the video (path, resolution, fps) and plotting kwargs.  Prints each date it processes since it takes a while to run (~3 min/month), so wanted to give some indicator of progress.
- changed single -> double quotes in the setup.py for consistency 
- updated some indentation on docs

Other notes:
- I have no idea how to test this properly/simply, other than running it and verify that _some_ file got output. Open to ideas here.

TODO:
- [x] Add Sphinx Documentation (probably use https://github.com/sphinx-contrib/video for embedding the video)

Usage Examples:

```
import covidcast                                     
from datetime import date         

df = covidcast.signal("fb-survey","smoothed_cli", start_day = date(2020,8,1), end_day = date(2020,8,5), geo_type = "county")
covidcast.animate(df, "test_animation.mp4")     
```
yields the following MP4 File: [test_animation.mp4.zip](https://github.com/cmu-delphi/covidcast/files/5230846/test_animation.mp4.zip) (would be unzipped, this is a zip file since that's the only upload supported format by github)
Screenshot:
Screenshot from 2020-09-16 01-23-19

```
covidcast.animate(df, "test_animation.mp4",  dpi=30, cmap="viridis",)        
```
Screenshot:
Screenshot from 2020-09-16 01-21-45


Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -170,6 +172,29 @@ def get_geo_df(data: pd.DataFrame,
...
+    :param dpi: Dots per inch for output video. Defaults to 150 on a 12.8x9.6 figure (1920x1440).
+    :return: None
+    """
+    # probesize is set to avoid warning by ffmpeg on frame rate up to 4k resolution.
+    writer = imageio.get_writer(filepath, fps=fps, input_params=["-probesize", "75M"])
+    num_days = (max(data.time_value) - min(data.time_value)).days
+    day_list = [min(data.time_value) + timedelta(days=x) for x in range(num_days+1)]
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        for d in day_list:
+            print(f"Processing {d.date()}")
```
capnrefsmmat, line 190: If you don't mind another import, [tqdm](https://tqdm.github.io/) is a very nice progress bar for Python

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -170,6 +172,29 @@ def get_geo_df(data: pd.DataFrame,
...
+            plot_choropleth(data, time_value=d, **kwargs)
+            plt.savefig(f"{tmpdirname}/{d}.jpg", format="jpg", dpi=dpi)
```
capnrefsmmat, line 192: You could probably use `io.BytesIO` to save the figure to memory, then `imread` from that, so no temp files needed... maybe?

Also, is jpg the best option here, since it ensures each frame will have compression artifacts even before mp4 compression? Maybe png

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -170,6 +172,29 @@ def get_geo_df(data: pd.DataFrame,
     return output
 
 
+def animate(data: pd.DataFrame, filepath: str, fps: int = 3, dpi: int = 150, **kwargs) -> None:
+    """Generate an animated mp4 file of a signal over time.
+
+    :param data: DataFrame for a single signal over time.
+    :param filepath: Path where video will be saved.
+    :param fps: Frame rate in frames per second for animation. Defaults to 3.
+    :param dpi: Dots per inch for output video. Defaults to 150 on a 12.8x9.6 figure (1920x1440).
```
capnrefsmmat, line 181: Probably should document `kwargs` to say they go to `plot_choropleth`, e.g. so you can set figsize

capnrefsmmat (approved):
Very nice. I'm going to enjoy having this feature available when we write blog posts or give presentations. I have a couple suggestions, but otherwise I think we just need to document this and try it a few times.

Not sure about testing. I expect automated tests of videos to be pretty gross, and it also sounds like it's pretty slow. I think as long as we use it somewhere that it gets run relatively often -- like it ends up in Sphinx docs or something -- we don't have to worry too much about accidentally breaking it and not noticing.

Python-packages/covidcast-py/covidcast/plotting.py
chinandrew: Ok this is weird. I swear two nights ago when I tried this with png it was giving me the weird bold text issue (even on mp4, which didn't make sense) but it's to be working now.

Good call on the bytesio, just added that

Python-packages/covidcast-py/covidcast/plotting.py
chinandrew: added

Python-packages/covidcast-py/covidcast/plotting.py
chinandrew: oops forgot that, added.

chinandrew:
Ok turns out that package doesn't support looping, and also wasn't installing properly, so I just added raw html to the docs.

chinandrew:
One browser specific issue I've found is with firefox, the penultimate day gets dropped and it jumps from the 29th to the 31st. It's in the source file, if you select the video and download it it's there too, but if you play the video with the in browser player it skips the 30th. Even if you slow down to 0.5x speed, so something weird is going on. I think it's a linux + firefox + html5 thing, but need to look into it more.

Doesn't happen with chrome, though fun fact is that chrome videos won't autoplay unless they're muted.

Have not tried this on other OSes yet.

Python-packages/covidcast-py/docs/plot_examples.rst
```diff
@@ -96,6 +102,41 @@ which can be stored and altered further.
 >>> fig = plotting.plot_choropleth(data)
 >>> fig.set_dpi(100)
 
+Animations
+----------
+To create an animation, simply pass the signal Dataframe to :py:func:`covidcast.animate`
```
capnrefsmmat, line 107:
```suggestion
To create an animation, simply pass the signal DataFrame to :py:func:`covidcast.animate`.
```

Python-packages/covidcast-py/docs/plot_examples.rst
```diff
@@ -96,6 +102,41 @@ which can be stored and altered further.
...
+
+.. raw:: html
+
+    <video width="640" height="480" playsinline autoplay muted loop>
+      <source src="_static/example_default_animation.mp4" type="video/mp4">
+        Your browser does not support the HTML5 Video element
+    </video>
+
+Video frame rate and resolution are adjustable. Like the static maps, additional plotting
+keyword arguments can be provided.
```
capnrefsmmat, line 126:
```suggestion
keyword arguments can be provided and are passed to :py:func:`covidcast.plot_choropleth`.
```

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -170,6 +173,35 @@ def get_geo_df(data: pd.DataFrame,
     return output
 
 
+def animate(data: pd.DataFrame, filepath: str, fps: int = 3, dpi: int = 150, **kwargs) -> None:
+    """Generate an animated video file of a signal over time.
+
+    Given a signal DataFrame, generates the choropleth for each day to form an animation of each
+    day. Accepts arguments for video parameters as well as optional plotting arguments.
```
capnrefsmmat, line 180:
```suggestion
    Given a signal DataFrame, generates the choropleth for each day to form an animation of the
    signal. Accepts arguments for video parameters as well as optional plotting arguments.
```

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -170,6 +173,35 @@ def get_geo_df(data: pd.DataFrame,
...
+    Supported output formats are available in the
+    `imageio ffmpeg documentation <https://imageio.readthedocs.io/en/stable/format_ffmpeg.html>`_
```
capnrefsmmat, line 182:
```suggestion
    `imageio ffmpeg documentation <https://imageio.readthedocs.io/en/stable/format_ffmpeg.html>`_.
```

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -170,6 +173,35 @@ def get_geo_df(data: pd.DataFrame,
     return output
 
 
+def animate(data: pd.DataFrame, filepath: str, fps: int = 3, dpi: int = 150, **kwargs) -> None:
+    """Generate an animated video file of a signal over time.
+
+    Given a signal DataFrame, generates the choropleth for each day to form an animation of each
+    day. Accepts arguments for video parameters as well as optional plotting arguments.
+    Supported output formats are available in the
```
capnrefsmmat, line 181:
```suggestion
    Supported output formats are listed in the
```

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -170,6 +173,35 @@ def get_geo_df(data: pd.DataFrame,
...
+    `imageio ffmpeg documentation <https://imageio.readthedocs.io/en/stable/format_ffmpeg.html>`_
+
+    :param data: DataFrame for a single signal over time.
+    :param filepath: Path where video will be saved. Filename must contain supported extension.
+    :param fps: Frame rate in frames per second for animation. Defaults to 3.
+    :param dpi: Dots per inch for output video. Defaults to 150 on a 12.8x9.6 figure (1920x1440).
+    :param kwargs: Optional keyword arguments passed to ``plot_choropleth()``
```
capnrefsmmat, line 188:
```suggestion
    :param kwargs: Optional keyword arguments passed to :py:func:`covidcast.plot_choropleth`.
```

capnrefsmmat (commented):
I haven't tried building and viewing the docs, but I think with these tweaks, it should be ready to merge. This is going to be really fun to use

capnrefsmmat:
(I'll add that I tested this on macOS and it worked perfectly on the first try, though as you noted, it is a bit slow)

PR #85, opened 2020-09-22
first commit of evalcast

jacobbien:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970469125926) by [Unito](https://www.unito.io/learn-more)


R-packages/evalcast/R/download_signals.R
```diff
@@ -0,0 +1,52 @@
...
+      state_fips <- e$county_geo %>%
+          select(.data$STATE, .data$COUNTYNAME, .data$FIPS) %>%
+          mutate(fips = str_sub(.data$FIPS, end = 2),
+                 geo_value = tolower(.data$STATE)) %>%
+          distinct(.data$geo_value, .data$fips)
+    out <- out %>%
+      left_join(state_fips, by = "geo_value") %>%
+      mutate(geo_value = .data$fips) %>%
+      select(-.data$fips)
+  }
```
ryantibs, line 49: @jacobbien Here I think you're trying to convert state abbreviations to FIPS codes, if I'm understanding it correctly? 

First, we have to get rid of the dependency on `county_geo` because I just merge a PR in #122 that overwrites this and potentially results in different columns.

Second, you should just rely on the utilities in `covidcast` for mapping between geo codes and names.  I think this will do it for you:

```
out$geo_value = name_to_fips(paste0("^", abbr_to_name(toupper(out$geo_value)), "$"))
```

Here the regex anchors "^" and "$" force a match to `v` that anchors it at the start and end of the string, respectively.  (To avoid matches to things like "Virginia" matching to both "Virginia" and "West Virginia", or "Wyoming" matching to both "Wyoming" and "Wyoming County".  There might be a cleaner way to do it ... but this should work.)

Try this as an example, if you want to see the state FIPS codes: 

```
name_to_fips(paste0("^", abbr_to_name(state.abb), "$"))
```

R-packages/evalcast/R/get_predictions.R
```diff
@@ -0,0 +1,119 @@
...
+#' @return a list of predictions cards
+#' @importFrom assertthat assert_that
+#' @export
+get_predictions <- function(forecaster,
+                            name_of_forecaster,
+                            signals,
+                            forecast_dates,
+                            incidence_period,
+                            ahead,
+                            geo_type,
```
ryantibs, line 41: @jacobbien Wondering if there's a reason to have repeated arguments here, since it seems like `signals`, `incidence_period`, `ahead`, and `geo_type` would all typically be known by the supplied `forecaster` itself.  Could these just not be read off from the `forecaster`?  For example, they could be saved as attributes of `forecaster`.  

As it is now, it looks like with your `baseline_forecaster()` example, I have to specify `signals`, `incidence_period`, `ahead`, and `geo_type` twice: once in the call to `baseline_forecaster()`, and once in the call to `get_predictions()`.  Not a big deal, just pointing it out to see if we can simplify.

ryantibs: Ahhh never mind.  I think I misunderstood the workflow (please do send me that notebook if you get a chance!).  I understand it now: `get_predictions()` simply calls `forecaster` and passes on these arguments to it.  My bad!  You can resolve this convo.

R-packages/evalcast/R/download_signals.R
ryantibs: See (bd6912b).

R-packages/evalcast/R/download_signals.R
```diff
@@ -31,21 +31,11 @@ download_signal <- function(...) {
...
-      state_fips <- e$county_geo %>%
-          select(.data$STATE, .data$COUNTYNAME, .data$FIPS) %>%
-          mutate(fips = str_sub(.data$FIPS, end = 2),
-                 geo_value = tolower(.data$STATE)) %>%
-          distinct(.data$geo_value, .data$fips)
-    out <- out %>%
-      left_join(state_fips, by = "geo_value") %>%
-      mutate(geo_value = .data$fips) %>%
-      select(-.data$fips)
+    out$geo_value = covidcast::name_to_fips(paste0("^",
```
JedGrabman, line 37: recommend using `fixed = TRUE` in name_to_fips instead of paste0 and regex.

JedGrabman: Nevermind on the `fixed` recommendation. That still matches on substrings, so adding `^` and `$` looks necessary.

I'll note that `District of Columbia` is the name of both the state and county, which results in a warning but ultimately returns the correct fips value, but is technically reliant on the order in county_census.

ryantibs: It's a good thought but unfortunately that doesn't work (I tried it).  Consider:
```
grep("Wyoming", c("Wyoming", "Wyoming County"), fixed = TRUE)
```

ryantibs: DC: it's a good point.  I guess we're lucky because the ordering is by increasing FIPS code, so "11000" appears before "11001".  See my next commit for the most robust fix; it fetches all the matches and extracts just the ones that end with "000".

jacobbien:
Thanks @ryantibs .  One thing though.  The location column for state level data used to have just 2 digits but now it has 5 digits.  The reason this matters is that we want to be consistent with `get_covidhub_predictions`, which ports data over from the COVIDHub which uses the convention of 2-digit FIPS codes for states.  See [here](https://github.com/reichlab/covid19-forecast-hub/blob/master/data-locations/locations.csv) for their convention.  If we want to stick with 5-digit codes then we can modify `get_covidhub_predictions` to pad with three 000s on the right.

ryantibs:
Oops, thanks for catching that @jacobbien.  My last commit ed6b019 fixes this.  Tomorrow I'm actually just going to add two convenience functions `state_to_fips()` and `fips_to_state()` to the `covidcast` package, and then we can merge that into this branch (shouldn't create any conflicts), and just rely on `state_to_fips()`.  

By the way this was a good example of why it's helpful to create branches (meaning, bad practice on my part!).  The main evalcast branch (this one) broke because we changed `county_geo` in `covidcast` and I wanted to fix it ASAP ... but I should have created a separate branch for fixing this FIPS code bug and submitted a PR ... then we could have had all this discussion over on that PR, and I'm sorry to clog this evalcast PR with extra discussion and extra commits.

ryantibs:
Hi @jacobbien, once you merge my PR #146, my review is officially done and I think we should be close to being able to merge evalcast into main!  

The documentation is in good shape (I made a comprehensive pass), so once @JedGrabman finishes the vignette, it would be great to build the doc and vignette using pkgdown, like we did with the covidcast package.  cc @capnrefsmmat 

As I explained in #146, my PR is mostly a review, not new feature development.  There's several things I want to add to evalcast, but I'm out of cycles right now and need to move on to something else.  (I'll do it on a new branch a bit later, and this shouldn't hold us up from merging evalcast into main.)

ryantibs:
@jacobbien Thanks for the thorough review and merge.  How far are we away from merging evalcast into main?  Aside from the vignettes being worked on by @JedGrabman (Jed, you can give us an update on your timeline?), what's left?

jacobbien:
@ryantibs - I think we're basically ready.  The one thing that might be good to do first is add some more tests.  I wrote some new ones as I was going over your review.  If I add those to the package then I think I'd be comfortable having this merged into main.  Should also probably bump the version number, which I always forget to do.

JedGrabman:
@ryantibs Got sidetracked by some bug fixes, but I'll have something to review today or tomorrow. It's using a significant part of the draft by @jacobbien

capnrefsmmat:
Since we have a bunch of PRs piling up that are dependent on this one, might I suggest we go ahead and merge, and if there are remaining items you think should be done, you can file additional issues to keep track of them?

ryantibs:
@capnrefsmmat @dajmcdon @jacobbien Actually, I think there's still significant number of things that need to be done before merging evalcast.  I've been trying to use it over the last couple days for comprehensive "real" forecast evaluation in the wild, and I've been running into some blockers.  Dan is steering the evalcast ship at this point and has a list of things that need to get done before it's main-ready (and just tagged a bunch of issues accordingly, I believe, with the "evalcast v1" label).

Alex: what's being held up other than #156?  This one, as I wrote on that PR, doesn't actually need evalcast and could be merged into main if we could just pick off the relevant commits.

capnrefsmmat:
With #156 in limbo, there are various other changes that will either have conflicts after merging or that are just waiting for #156 (such as some testing things I'd like to do).

What exactly does "main-ready" mean? It's much easier to develop if you can make numerous small, easily reviewed PRs, rather than accumulating many changes into one huge PR. Being on main need not imply that we're advertising to the world that the package is ready for widespread use.

ryantibs:
That’s a good point Alex, we were thinking of main-ready as production-ready but I guess it really doesn’t have to be that way.

The package should be functional at the last evalcast commit. But still lacking in some key ways that we’re trying to fix. So maybe we should just merge to main and continue development that brings it to production-ready on a branch (and shoot for multiple smaller PRs if needed). @dajmcdon @jacobbien what do you think?

dajmcdon:
Yeah, I don't think there's anything sacred about main at the moment. We want our production to have a "released" version with known features. And we do need many of the features tagged as v1. But there are other users as well. Since the forecast-pipeline isn't running yet, I don't really see any reason not to just do a PR against main now. Which branch is most full-featured is not clear to me. Maybe the one @ryantibs is currently using? 

Once that's done, we'll Christen a development branch and do PRs for the pipeline-blocking features (most of the issues I just made) against that.

ryantibs:
I think this PR on this branch (evalcast) should be the one to merge to main. 

I stashed my some of my fixup work in case we need it but I was just rushing to undo assert statements as they were halting my runs but I think it’s better for someone to do it more systematically from the start.

PR #86, opened 2020-09-22
evalcast talk

jacobbien:


ryantibs:
Nice job!  I just went through it and made bunch of small edits (mostly copyedits).

I tried to change script.html to allow you to specify that you want certain code chunks shown by default. I thought I got it working but then it looks like I broke something else, so I ditched this effort. 

This seems like it should be an easy job for @sgratzl: Sam, can you add functionality to script.html so that we can specify which code chunks we want *shown* initially? (With all of them hidden by default.)  I thought we should be able to wrap the code chunks in something like `<div class = "show">` ... `</div>` and then detect the "show" class in the JavaScript.

jacobbien:
Great, thanks Ryan!  Oh, yes, Sam, if you could do that, that would be
wonderful!

sgratzl:
I created https://github.com/cmu-delphi/covidcast/pull/88

with an improved script that allows you to expand a code block by:

```{r, fig.width=7, fig.height=5, class.source="show"}
plot(iris)
```

ryantibs:
Great, thank you @sgratzl.  I assume @jacobbien you'll incorporate this but lmk if you want me to take a look.

Jacob a few comments:

- The knitr options set at the start of the talk turn caching on globally, for each code chunk (so you don't need to specify this in individual chunk options).  Also I removed the cached files from the repo (but you can of course keep them locally).

- Looking again this morning, then one thing that's missing to me is more of a "real" demo of the evaluation metrics.  Comparisons where we can actually learn something.  As it is right now, I think people won't appreciate the depth of the tools here (the really amazing work you and others put it into it).  I know you chose CH baseline and ensemble to be neutral, but maybe we could ask UMass and YYG if they wouldn't mind being shown?  And you could actually walk through the results to infer something about their *behavior*?  After all their predictions are public; I don't see why there's anything wrong with it.  If you want me to, I can ask them directly.

ryantibs:
Hi @jacobbien, I'm going to go ahead and merge this now, just to be safe.  I'm out all day tomorrow.  If you make any more edits before Tues you can always push them to this branch and resubmit a PR.

Issue #76, opened 2020-09-11
County out of place in R bubble plots


capnrefsmmat:
There's a county being placed somewhere in Canada in the bubble plots (just under "confirmed"):

<img width="311" alt="Screen Shot 2020-09-11 at 7 29 31 PM" src="https://user-images.githubusercontent.com/711629/92980677-218d3d00-f465-11ea-883c-c4f93cb17546.png">

ryantibs:
That was an Easter egg to acknowledge my Canadian heritage!

Just kidding. I'll think I know why this is happening, I'll try to get to this sometime this weekend, as well as #73 and Andrew's comments in #43.

capnrefsmmat:
@ryantibs FYI I have half an attempt to fix #73 in progress, except in my fix, the legend scales by area and the actual plot still seems to scale by radius... not sure what's going on. So I guess feel free to make a separate attempt to fix it, because I'm currently stumped.

ryantibs:
So this turns out to be a bit different than what I thought originally.  It looks like a bug with `usmap::usmap_transform()`

That misplaced bubble looks to be Aleutians West:


  | STATE<chr> | CWA<chr> | COUNTYNAME<chr> | FIPS<chr> | FE_AREA<chr> | LON<dbl> | LAT<dbl>
-- | -- | -- | -- | -- | -- | -- | --
3297 | AK | AFC | Aleutians West | 02016 | NA | -111.1392 | 52.8883

The `usmap` package tucks Alaska in nicely in the bottom left (next to where it places Hawaii) but somehow gets this wrong.  I also think that the extra counties that are floating around the coast of Hawaii are could also be Alaska errors.  The Alaska counties generally look kind of misaligned.

![image](https://user-images.githubusercontent.com/732541/94395343-79969580-012d-11eb-95b0-3eb004924960.png)

capnrefsmmat:
It looks like the usmap package specifies a bounding box for Alaska and gives special treatment to everything in that box: https://github.com/pdil/usmap/blob/master/R/transform.R#L87

Is it possible that the Aleutians don't entirely fit inside that bounding box? If so, we could report a bug in their repo and give the example of the island that's not in the box.

JedGrabman:
I think the issue is actually in our county_geo dataset:

```
> covidcast::county_geo[covidcast::county_geo$FIPS == "02016",]
     STATE CWA     COUNTYNAME  FIPS FE_AREA       LON     LAT
[...]

Issue #73, opened 2020-09-10
Fix R bubble plots to scale by area instead of radius


capnrefsmmat:
In ggplot2's `geom_point`, `size` means radius, rather than area, so `plot_bubble` generates bubbles with radius proportional to value instead of area proportional to value. This exaggerates the size of the largest bubbles.

Currently `plot_bubble` manually constructs a size scale. We should see if we can use [`scale_size_binned_area()`](https://ggplot2.tidyverse.org/reference/scale_size.html) or similar to achieve the same things but with scaling by area.

capnrefsmmat:
Fixed in #91.

Issue #43, opened 2020-08-27
Add bubble map support for plotting


chinandrew:
Currently only choropleths are supported with plotting. Add support for bubble maps like what is available in the [R package](https://cmu-delphi.github.io/covidcast/covidcastR/articles/plotting-signals.html#bubble-maps)

chinandrew:
Few questions on this one:
1. Do we want points to scale by area or radius? i.e. for a 2x increase, does the radius or area increase by 2x? I figured doing area would be most accurate, but the R package (screenshot) looks like it scales by radius based on my super scientific inspection below. matplotlib scales by area by default.
    - Value of 1.94 of left, value of 0.89 on right, 700% scale
    ![Screenshot from 2020-09-09 21-59-56](https://user-images.githubusercontent.com/20176218/92683480-d4379100-f2e7-11ea-87ab-a2f84bb302c8.png)

2. Do we want the non-values to have the specific regions outlines (the R implementation, first screenshot), or is the state background sufficient (my current Python implementation, second screenshot)?
![Screenshot from 2020-09-09 22-01-01](https://user-images.githubusercontent.com/20176218/92683532-fb8e5e00-f2e7-11ea-9eed-bcff02f327ec.png)
![Screenshot from 2020-09-09 22-45-55](https://user-images.githubusercontent.com/20176218/92686433-7eb2b280-f2ee-11ea-859b-63000e4381e2.png)

3. @ryantibs One thing I noticed on the R code for these plots: the R discretization function seems to null any values < 0.1 (or whatever the lower bound happens to be), so they don't get a bubble. The region still gets un-greyed, so you know the region reported a value < 0.1 and wasn't missing. Is this intentional? I wasn't able to find the bubble plots on the covidcast website so didn't have reference there.
```
> breaks = c(0.1,0.2,0.3,0.4)
> dis_fun(c(0,0.05,0.1))
[1]  NA  NA 0.1
```
Empty values:

![Screenshot from 2020-09-09 20-17-07](https://user-images.githubusercontent.com/20176218/92687037-a5251d80-f2ef-11ea-854a-cc8fccb2ffc6.png)
![Screenshot from 2020-09-09 23-02-35](https://user-images.githubusercontent.com/20176218/92687562-a145cb00-f2f0-11ea-8aad-48d8d0adbfee.png)

[...]

PR #78, opened 2020-09-12
WIP Fix several R package bugs

capnrefsmmat:
Working on #74, #73, and miscellaneous issues discovered on the way.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970335796737) by [Unito](https://www.unito.io/learn-more)

PR #91, opened 2020-09-28
Plotting fixes

ryantibs:
Hi all, particularly @capnrefsmmat @chinandrew, some miscellaneous plotting fixes in response to comments raised by Andrew in #43.

I actually made the commit messages a bit more informative here, so hopefully you can tell what's going on.  This PR should address #73 but unfortunately not #76.  The latter appears to be a bug with `usmap::usmap_transform()`.  I'll write more on that issue.

Alex: I decided to number the vignette entries already, because I made a couple small edits to the vignettes anyway (and I know the correlations one takes a long time to rebuild).  And this will incentivize us (me) to write a third vignette sooner than later, so it doesn't look so barebones with just two, for too long.

### DIFF
--- a/R-packages/covidcast/R/covidcast.R
+++ b/R-packages/covidcast/R/covidcast.R
@@ -331,7 +331,8 @@ summary.covidcast_signal = function(object, ...) {
 #'   covidcast.cmu.edu.
 #' @param bubble_col Bubble color for the bubble map. Default is "purple".
 #' @param num_bins Number of bins for determining the bubble sizes for the
-#'   bubble map. Default is 6. These bins are evenly-spaced in between the min
+#'   bubble map (here and throughout, to be precise, by bubble size we mean
+#'   bubble area). Default is 8. These bins are evenly-spaced in between the min
 #'   and max as specified through the `range` parameter. Each bin is assigned
 #'   the same bubble size. Also, values of zero special: it has its own separate
 #'   (small) bin, and values mapped to the zero bin are not drawn.
@@ -351,6 +352,7 @@ summary.covidcast_signal = function(object, ...) {
 #' \item{`missing_col`}{Color assigned to missing or NA geo locations.}
 #' \item{`border_col`}{Border color for geo locations.}
 #' \item{`border_size`}{Border size for geo locations.}
+#' \item{`legend_position`}{Position for legend; use "none" to hide legend.} 
 #' \item{`legend_height`, `legend_width`}{Height and width of the legend.} 
 #' \item{`breaks`}{Breaks for a custom (discrete) color or size scale.  Note
 #'   that we must set `breaks` to be a vector of the same length as `choro_col`
@@ -358,7 +360,7 @@ summary.covidcast_signal = function(object, ...) {
 #'   choropleth maps, or the `i`th size for bubble maps, if and only if the
 #'   given value satisfies `breaks[i] <= value < breaks[i+1]`, where we take by 
 #'   convention `breaks[0] = -Inf` and `breaks[N+1] = Inf` for `N =
-#'   length(breaks)`.}
+#'   length(breaks)`.}   
 #' \item{`legend_digits`}{Number of decimal places to show for the legend
 #'   labels.}  
 #' }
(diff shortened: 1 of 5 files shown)


R-packages/covidcast/R/covidcast.R
```diff
@@ -391,23 +391,20 @@ plot.covidcast_signal <- function(x, plot_type = c("choro", "bubble", "line"),
...
-    if (is.null(attributes(x)$metadata)) {
-      warn(
-        paste0("Metadata for signal mean and standard deviation not available; ",
-               "defaulting to observed mean and standard deviation to set plot range."),
-        class = "covidcast_plot_meta_not_found"
-      )
-
+  # For the maps, set range, if we need to (to mean +/- 3 standard deviations,
+  # from metadata) 
+  if (is.null(range) && plot_type == "choro" || plot_type == "bubble") {
```
capnrefsmmat, line 408: `&&` has higher precedence than `||`, so this is

```r
if ((is.null(range) && plot_type == "choro") || plot_type == "bubble") {
```

But you probably want 

```suggestion
  if (is.null(range) && (plot_type == "choro" || plot_type == "bubble")) {
```

otherwise `range` will always be ignored for bubble plots.

R-packages/covidcast/R/plot.R
```diff
@@ -411,16 +414,22 @@ plot_line = function(x, range = NULL, col = 1:6, line_type = rep(1:6, each = len
     ggplot2::theme(legend.position = "bottom",
                    legend.title = ggplot2::element_blank())
 
-  # Create line and lim layers
-  line_layer = ggplot2::geom_line(ggplot2::aes(x = time_value, y = value,
+  # Create lim, line, and ribbon layers
+  lim_layer = ggplot2::lims(y = range)
```
capnrefsmmat, line 431: Do you really want `lims`? Values outside the range will be deleted entirely. That means that if, say, the user sets a range of [0, 10], and one value happens to be 11, you won't see the graph spike off the top and come back down -- it'll just connect the two surrounding points with a straight line.

You probably want to use [coord_cartesian](https://ggplot2.tidyverse.org/reference/coord_cartesian.html) to set the ylim.

(And if this annoys you, this was also Howard's biggest peeve about ggplot2, so you're not alone)

R-packages/covidcast/R/plot.R
```diff
@@ -272,22 +275,26 @@ plot_bubble = function(x, time_value = NULL, include = c(), range = NULL,
...
   }
 
-  # Create bubble sizes
+  # Max and min bubble sizes
   min_size = params$min_size
   max_size = params$max_size
   if (is.null(min_size)) min_size = ifelse(attributes(x)$geo_type == "county",
                                            0.1, 1)
   if (is.null(max_size)) max_size = ifelse(attributes(x)$geo_type == "county",
                                            4, 12)
```
capnrefsmmat, line 284: I got the impression these default sizes were carefully tuned to make the map look good by default. Do they need to change now that you're on the area scale (and ignoring the factor of pi)?

capnrefsmmat (changes requested):
Two bugfix suggestions and one question. I'll build #78 on this when you're done

R-packages/covidcast/R/plot.R
ryantibs: Good catch!

ryantibs: These actually still refer to radii, so the smallest and biggest points should look the same as before.

When I define the sequence of radii to use, I just define them to be equally-spaced on the squared scale, so that the effective areas are equally-spaced on the linear scale.  Check out line 289.

I should be clear: in the code, any reference to "size" (by variable name) still refers to a radius.  Because that's the way `geom_line()` and `scale_size_manual()` treat it.  This is perhaps confusing (and what may have confused you here), because in the documentation, I use "size" (in written language) to refer to an area.

R-packages/covidcast/R/covidcast.R
ryantibs: Good catch!

R-packages/covidcast/R/plot.R
ryantibs: Good catch!  That's pretty annoying alright.  However, not my biggest peeve: there are still plenty of other contenders for the top spot. 😉

PR #92, opened 2020-09-28
Update Python documentation before next release 

### LINKED ISSUE #90: Cut a new Python client release
- [x] Update changelog to describe changes since 0.0.9
- [x] Rebuild all documentation and proofread for typos
- [x] Change version number in setup.py and the Sphinx config file
- [x] Build new package dists
- [x] Upload to PyPI
- [ ] Announce to API mailing list

@chinandrew, any changes we need to make before we should release 0.0.10? Should we celebrate by calling it 0.1.0?

chinandrew:
Closes #90 

Summary of changes:
- increments version to 0.1.0 in setup.py and conf.py
- updates changelog
- various documentation changes

### DIFF
--- a/Python-packages/covidcast-py/docs/changelog.rst
+++ b/Python-packages/covidcast-py/docs/changelog.rst
@@ -1,19 +1,37 @@
 Changelog
 =========
 
-v0.0.10, TODO
+v0.1.0, TODO
 -------------
 
 - New feature: Functions :py:func:`covidcast.fips_to_name`,
   :py:func:`covidcast.cbsa_to_name`, and :py:func:`covidcast.abbr_to_name` are
-  provided to convert FIPS codes, CBSA IDs, and two-lettr abbreviations to
-  county names, meeteropolitan statistical area names, and state names,
+  provided to convert FIPS codes, CBSA IDs, and two-letter abbreviations to
+  county names, metropolitan statistical area (MSA) names, and state names,
   respectively.
 
 - New feature: Functions :py:func:`covidcast.name_to_fips`,
   :py:func:`covidcast.name_to_cbsa`, and :py:func:`covidcast.name_to_fips` are
   provided to convert human-readable names into FIPS codes, CBSA IDs, and state
   abbreviations, for use when querying the COVIDcast API.
+
+- New feature: Function :py:func:`covidcast.animate` can take a DataFrame returned by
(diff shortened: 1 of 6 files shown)


Python-packages/covidcast-py/docs/changelog.rst
```diff
@@ -1,20 +1,39 @@
...
+  multiple signals into a single DataFrame for convenient signal comparisons across
+  time.
+
+- Bug fix: The `time_value` argument to the :py:func:`covidcast.plot_choropleth`
+  function can now accept `datetime <https://docs.python.org/3/library/datetime.html>`_
+  `datetime` and `date` objects, instead of only `pandas Timestamps
+  <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Timestamp.html>`_.
+
+- Bug fix: :py:func:`covidcast.signal` no longer returns duplicate values if duplicate
+  `geo_values` are provided in the function call.
```
capnrefsmmat, line 35:
```suggestion
- Bug fix: The ``time_value`` argument to the :py:func:`covidcast.plot_choropleth`
  function can now accept `datetime <https://docs.python.org/3/library/datetime.html>`_
  ``datetime`` and ``date`` objects, instead of only `pandas Timestamps
  <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Timestamp.html>`_.

- Bug fix: :py:func:`covidcast.signal` no longer returns duplicate values if duplicate
  ``geo_values`` are provided in the function call.
```

Not sure if this is deliberate or not, but in reStructuredText, you need two backticks to make code, unlike Markdown. One backtick just makes italics

capnrefsmmat (commented):
Looks good. One suggestion. Also one other question: I notice, on viewing the rendered docs, that the videos don't have controls. (I think you can add the `controls` argument to turn this on.) I disabled autoplaying video in Firefox, so the videos do not play by default for me, and without the controls I can't tell the files are videos without right-clicking and hitting Play. Could we enable the controls? This will also let people play more with the videos.

chinandrew:
> Looks good. One suggestion. Also one other question: I notice, on viewing the rendered docs, that the videos don't have controls. (I think you can add the `controls` argument to turn this on.) I disabled autoplaying video in Firefox, so the videos do not play by default for me, and without the controls I can't tell the files are videos without right-clicking and hitting Play. Could we enable the controls? This will also let people play more with the videos.

Yup, added controls to show by default.

Issue #94, opened 2020-09-30
Make the package documentation serve as a better starter guide


capnrefsmmat:
For both the R and Python package documentation, we need to be a bit more welcoming. A new user will want to know

1. How can I find a certain type of data? What kind of data is available? The package docs refer you to the COVIDcast API docs, but don't give any overview and don't explain how to find the appropriate data source and signal names to plug into the packages.
2. What does it mean that this is an API? Explain how the data is fetched live from the server when you request it, rather than being distributed from the package.
3. How up-to-date is this data? Both generally discuss how the API is updated daily, and specifically how to use the metadata to get the most recent update date.
4. How do I interpret the data? We cover this a bit in the documentation of return values, but we may want to be more explicit in the vignettes/examples. (Also, again refer the user to the API docs when they want to know how to interpret a specific signal.)

chinandrew:
How do we think about what belongs in the covidcast documentation vs the epidata api documentation? It sounds like the intention is that the covidcast package is aimed at users across the spectrum of technical skills, and the documentation should be pretty self suffiicentn with the exception of the specifics of each signals. Is that a good starting point?

capnrefsmmat:
Yes, I think that's about right.

brookslogan:
To add to 1., "What kind of data is available?", I think we should aim to at least have all signals findable via Ctrl-F from the [first page](https://cmu-delphi.github.io/delphi-epidata/api/covidcast.html) you get when googling covidcast docs, or maybe even have a long list of all signals (not just select examples) available near the beginning of the documentation.  It takes currently takes some effort to find out what the mask usage signals are called or whether they were documented at all, while already knowing that they exist (not findable searching for "mask" [here](https://cmu-delphi.github.io/delphi-epidata/api/covidcast.html) nor [here](https://cmu-delphi.github.io/delphi-epidata/api/covidcast_signals.html) nor [here](https://cmu-delphi.github.io/delphi-epidata/api/covidcast_changelog.html#fb-survey)); many potential users may not discover these signals' existence or may be frustrated.

capnrefsmmat:
That's a good point. The signal documentation is currently organized from our perspective: we have people who each work to import a data source and its signals, and the documentation is split up by source accordingly. But a user may instead have a topic of interest, and want to know which signals address that topic, regardless of source.

[...]

PR #111, opened 2020-10-06
Add source overview to documentation

chinandrew:
Addresses point 1 and 2 on #94

Summary of changes:
- Add section explaining the API
- Add an overview of sources and instructions on how to get the names to pass to covidcast.signal()
- Move installation info to "Getting Started"

I will say that the info duplication between this and the epidata api on sources (and the need to keep them in sync) is a bit messy, but probably ok for now. Also it probably looks better as a table, so I need to figure out how to get restructured tables to look good.

Once we finalize the language/structure, I can work on adding to the R docs as well.


krivard (approved):
Looks good to me!

PR #122, opened 2020-10-15
Fix misaligned counties

JedGrabman:
This removes a mapping dependency on `county_geo` or `state_geo`, and instead directly uses centroid data from usmap for bubble plots. Motivation for this change is #76 , where Aleutians West (Alaska) was misplaced.

Here is a replica of the map from that issue: jul_14

There is no longer a dot in Canada. This also appears to have resolved an issue where some points (I presume from Alaska) were showing up near Hawaii.

Here is the location of just Aleutians West Aleutians West


R-packages/covidcast/R/plot.R
```diff
@@ -346,22 +350,32 @@ plot_bubble = function(x, time_value = NULL, include = c(), range = NULL,
...
-    cur_lon = g$LON
-    cur_lat = g$LAT
+    centroid_file = system.file("extdata",
+                                "us_states_centroids.csv",
+                                package = "usmap")
+    centroids = utils::read.csv(centroid_file,
+                                colClasses = centroid_col_classes)
+    centroids$abbr = tolower(centroids$abbr)
+    centroids = centroids[centroids$abbr %in% map_geo, ]
+    centroids = centroids[match(map_geo, centroids$abbr), ]
```
krivard, line 376: What does this do?

JedGrabman: The first line is the equivalent of the previous `g = county_geo[county_geo$FIPS %in% map_geo, ]` and reduces the dataframe to the counties that are being mapped.

The 2nd line is superfluous and has been removed. I had an issue with data being mapped in the wrong spot and this was a fix. However, the root cause had to do with loading the fips column as numeric instead of character so it's no longer necessary after I fixed that.

krivard (approved):
Looks good!

R-packages/covidcast/R/plot.R
```diff
@@ -346,22 +350,27 @@ plot_bubble = function(x, time_value = NULL, include = c(), range = NULL,
...
 
-  # Set the lats and lons for counties
+  # Retrieve coordinates for mapping
+  # Reading from usmap files to ensure consistency with borders
   if (attributes(x)$geo_type == "county") {
-    g = county_geo[county_geo$FIPS %in% map_geo, ]
-    cur_geo = g$FIPS
-    cur_lon = g$LON
-    cur_lat = g$LAT
+    centroid_file = system.file("extdata",
```
ryantibs, line 356: @JedGrabman Instead of reading something from csv each time this gets called, you could do this one, back out in the data-raw/make.R file, and overwrite the `county_geo` and `state_geo` data frames that get defined there, and that get loaded when the `covidcast` package is loaded.  

I imagine this is the preferred route, but do you see any reasons not to do it the way I'm proposing?  Also tagging @capnrefsmmat.

capnrefsmmat: Yeah, I think the best options would be (a) get these into a dataset provided with the package, like county_geo or state_geo, or (b) if usmap exports this as a data frame instead of just a CSV, import that from usmap.

JedGrabman: @ryantibs 
Agreed, that seems like the correct approach.

@capnrefsmmat 
a.) Sounds good  
b.) Unfortunately, I couldn't find anywhere that usmap makes these available as a dataframe. That would be preferable, because then we wouldn't be relying on the internal implementation of usmap. Without that, it seems our choice is manually curating the location data vs. relying on usmap's csv. Creating the geo files in make.R should mitigate, but not entirely eliminate, the risk from reading from the csv.

ryantibs: After you update make.R, you only need to run it once to population the data files (currently `county_geo`, `state_geo`, etc.) that go into the package.  (That is, I'm just clarifying to @JedGrabman that it's not dynamically run each time on package load.)  So the only risk is that if usmap updates its shape files for drawing the map, and we don't catch it and don't rerun make.R, then the given `county_geo` and `state_geo` data sets in covidcast will be stale.  But once we catch it, we can just rerun make.R.  Since this shouldn't happen very often, this seems like a low-risk approach to me.

JedGrabman: Sounds good. I've replaced the `county_geo` and `state_geo` files. This change is not backwards compatible if somebody is directly using those files.

ryantibs: Looks good to me.  Thanks!  I will merge the PR now.

capnrefsmmat:
Does the documentation in [data.R](https://github.com/cmu-delphi/covidcast/blob/main/R-packages/covidcast/R/data.R) need to change to account for this change to the data frames? If so, please make another PR (and run roxygen and pkgdown so the generated docs are updated).

ryantibs:
@JedGrabman @capnrefsmmat I started a [new branch for doc updates](https://github.com/cmu-delphi/covidcast/pull/130), Jed you can just commit the data.R doc update to that PR.

PR #126, opened 2020-10-16
Change stopifnot to asserthat.

sarah-colq:


R-packages/evalcast/R/baseline.R
```diff
@@ -25,7 +25,8 @@ baseline_forecaster <- function(df,
                                 incidence_period,
                                 ahead,
                                 geo_type) {
-  stopifnot(incidence_period %in% c("epiweek", "day"))
+  assert_that(incidence_period %in% c("epiweek", "day"),
+              msg="incidence_period must be epiweek or day.")
```
capnrefsmmat, line 29: Here I'd actually have `incidence_period` have a default value of `c("epiweek", "day")`, then have the function open with

```r
incidence_period <- match.arg(incidence_period)
```

This is the standard way R functions handle an argument that can be one of several values; check out, say, `help(cor)` for an example (the `method` argument). And `match.arg` will raise an appropriate error if the argument is the wrong type.

R-packages/evalcast/R/evaluate.R
```diff
@@ -31,7 +31,7 @@
 #'
 #' @param predictions_cards a list of prediction cards from the same forecaster
 #' that are all for the same prediction task, meaning they are for the same
-#' response, incidence_period, ahead, and geo_type. Each should
+#' response, incidence_period, ahead, and geo_type. Each shouldts.
```
jacobbien, line 34: Looks like a typo in the comment here.

sarah-colq:
Done!

PR #130, opened 2020-10-20
Doc update

ryantibs:
A few small doc updates. 

@JedGrabman You can commit your doc updates for data.R to this PR, [as Alex pointed out is needed](https://github.com/cmu-delphi/covidcast/pull/122#issuecomment-713019702).

@capnrefsmmat Look for my comment below about another doc update that's needed before merging.

### DIFF
--- a/R-packages/covidcast/R/covidcast.R
+++ b/R-packages/covidcast/R/covidcast.R
@@ -91,11 +91,11 @@ COVIDCAST_BASE_URL <- 'https://api.covidcast.cmu.edu/epidata/api.php'
 #'   issued in this range. There may be multiple rows for each observation,
 #'   indicating several updates to its value. If `NULL`, the default, return the
 #'   most recently issued data.
-#' @param lag Integer. If, for example, `lag=3`, fetch only data that was
-#'   published or updated exactly 3 days after the date. For example, a row with
-#'   `time_value` of June 3 will only be included in the results if its data was
-#'   issued or updated on June 6. If `NULL`, the default, return the most
-#'   recently issued data regardless of its lag.
+#' @param lag Integer. If, for example, `lag = 3`, then we fetch only data that
+#'   was published or updated exactly 3 days after the date. For example, a row
+#'   with `time_value` of June 3 will only be included in the results if its
+#'   data was issued or updated on June 6. If `NULL`, the default, return the
+#'   most recently issued data regardless of its lag.
 #' 
 #' @return Data frame with matching data. Each row is one observation of one
 #'   signal on one day in one geographic location. Contains the following
--- a/R-packages/covidcast/R/covidcast.R
+++ b/R-packages/covidcast/R/covidcast.R
@@ -35,8 +35,8 @@ COVIDCAST_BASE_URL <- 'https://api.covidcast.cmu.edu/epidata/api.php'
 #' returns the most recent issue available for every observation. The `as_of`,
 #' `issues`, and `lag` parameters allow the user to select specific issues
 #' instead, or to see all updates to observations. These options are mutually
-#' exclusive; if you specify more than one, `as_of` will take priority over
-#' `issues`, which will take priority over `lag`.
+#' exclusive, and you should only specify one; if you specify more than one, you
+#' may get an error or confusing results.
 #'
 #' Note that the API only tracks the initial value of an estimate and *changes*
 #' to that value. If a value was first issued on June 5th and never updated,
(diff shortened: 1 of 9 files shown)


R-packages/covidcast/R/covidcast.R
```diff
@@ -91,11 +91,11 @@ COVIDCAST_BASE_URL <- 'https://api.covidcast.cmu.edu/epidata/api.php'
 #'   issued in this range. There may be multiple rows for each observation,
 #'   indicating several updates to its value. If `NULL`, the default, return the
 #'   most recently issued data.
-#' @param lag Integer. If, for example, `lag=3`, fetch only data that was
-#'   published or updated exactly 3 days after the date. For example, a row with
-#'   `time_value` of June 3 will only be included in the results if its data was
-#'   issued or updated on June 6. If `NULL`, the default, return the most
-#'   recently issued data regardless of its lag.
+#' @param lag Integer. If, for example, `lag = 3`, then we fetch only data that
```
ryantibs, line 94: @capnrefsmmat In reading the documentation for `as_of`, `issue`, `lag`, I realized we don't explain what happens if we attempt to specify more than one?  It looks to me like they don't actually intersect, and maybe `issues` takes priority over `lag` (by fiddling around with an example)?  But we should clarify this.

capnrefsmmat: In practice, `issues` takes precedence over `lag, which takes precedence over `as_of`, as [currently implemented](https://github.com/cmu-delphi/delphi-epidata/blob/7c3b3c6d0dbaf2bd6628907498e0098dcaee02a8/src/server/api.php#L975). If you specify two of them, one will win over the other, which will be ignored.

But I think that's not a guarantee of the API. The safe thing is to say that you can only specify one, and if you specify two or more, you're doing it wrong. (Maybe even raise an error and say that doesn't make sense, since I could easily see someone misreading the docs, trying it, and being confused by the results.)

ryantibs: Agree that it makes sense to throw an error if they try to specify more than one.  Seems like a pretty obscure use case if you want more than one (so we can eliminate this use case until a user complains for it).

Question is: do this on the client side or the API side?  Curious to hear opinions of @krivard @undefx.

krivard: From [the docs](https://cmu-delphi.github.io/delphi-epidata/api/covidcast.html#optional):

> You should specify only one of these three parameters in any given query.

Definitely open to rewrites that make this more apparent, since at least two very smart people skipped right over it. cc @avannort

As for where the error should be thrown: Upcoming work on deletion/missingness encoding will include a richer set of error messages the API may return, so server-side would be appropriate.

dfarrow0: Wouldn't hurt to have the error raised on both the client and the server. The client because it shortcuts a request that's going to fail anyway, and the server because it's possible to bypass the client altogether with a HTTP GET to the API.

For what it's worth, we already raise an error for invalid requests like this in the Epidata client: https://github.com/cmu-delphi/delphi-epidata/blob/5bc4131384f4d234d741997fa77700ae10568def/src/client/delphi_epidata.R#L515

dfarrow0: A more elaborate way of avoiding this _in the futue_ is to make the API unable to represent such a request (e.g. with both `lag` and `issues`). I don't know if there's an official name for it, but it's a design principle where you make invalid states impossible to represent. Trivial example: if you want to prevent passing a negative number to a function, you make the parameter type unsigned.

So we could make the API look something like this (in python since that's easier to write):

```python
def covidcast_signal(data_source, signal, time_spec=None):
   # call the API here...
  return data

class TimeSpec:

  class Type(enum):
    AS_OF, ISSUES, LAG

  def asOf(date):
    return TimeSpec(Type.AS_OF, date)

  def forIssues(issues):
    return TimeSpec(Type.ISSUES, issues)

  def atLag(lag):
    return TimeSpec(Type.LAG, lag)

  def __init__(type, value):
    self.value = value
    self.type = type
```

Then it would be called like this:

```python
data = covidcast_signal('source', 'signal', TimeSpec.atLag(7))
```

And it would be impossible for example to call with both `lag` and `issues`.

But that might be overkill, and it would break the API for existing callers, so we'd have to be backwards-compatible. Which means we're back to the same issue of detecting the invalid state...

ryantibs: @capnrefsmmat Actually, I suggest for now you just put a line in the doc saying "don't specify all three", and we handle the general problem on another branch.  This was supposed to be a quick-and-easy doc updates branch, we shouldn't let it get blocked by this.

capnrefsmmat: Done. Turns out there was already a note about this, except it lied about the order of precedence, so I've switched it to just say that specifying more than one is a bad idea

ryantibs: (Thanks by the way @krivard @undefx for the super helpful replies.)

ryantibs:
@capnrefsmmat Due to my unreasonable impatience (actually ... due to the fact that I'm currently working across way too many branches and I don't trust my weak git stash skills to hold up much longer!) ... I just added the documentation for the new data frames now. I built the documentation but not the pkgdown, since I think you'll have to build the pkgdown after merging #136 anyway.  So, please go ahead and merge this one first, then merge #136.  Fingers crossed for no conflicts.

Issue #32, opened 2020-08-21
Create basic dashboards for our signals


ryantibs:
In writing the Facebook blog post, I wanted to know a bunch of basic things about the Facebook survey, and found myself writing code to find the answers.  Then I thought we should just build a very basic dashboard so that we can constantly check these things.  Dashboard is probably too generous, I'm just talking about a notebook that gets run automatically each day.  I put one up [here](https://cmu-delphi.github.io/covidcast/R-notebooks/fb_dashboard.html).

This was pretty easy to build with the functionality in our covidcast R package.  If people agree it's useful, then we should make one of these for each of our signals, and write a script that runs these each morning.

Note: I guess we should discuss what this is for and who would use it.

* I think it's mainly for us, and it allows us to keep basic tabs on our indicators just by browsing a notebook.
* We could assign one person from the engineering team to take a look at one of these notebooks, once per day (which ideally would update itself).
* For example, I can see from the notebook linked above that the Facebook survey has been consistently trending down in responses per day and counties per day, over the last month ... if this keeps up we might go and tell Facebook.
* We could also use it to do some very basic anomaly detection, nothing fancy---we'd have to add a few plots for this.  This might be similar to what we're already getting from Larry (Pickett?) but could be worth just implementing something very basic on our side as well. 

Bottom line, I would say, is that: nothing fancy here.  I think we want to keep these notebooks simple to maintain and simple to look at.

Tagging krivard capnrefsmmat jsharpna because they might have ideas of who could tackle this. It's relatively low priority but would be a good way for somebody to onboard themselves both with respect to getting to know our R client and our signals.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970455381965) by [Unito](https://www.unito.io/learn-more)

benjaminysmith:
I am adding basic coverage and sampling dashboards for all of the signals, similar to what we already have for symptom surveys.

- [ ] Doctor Visits
- [ ] Google Health Trends
- [ ] Hospital Admissions
- [ ] Indicator Combination
- [ ] JHU Cases and Deaths
- [ ] Quidel
- [ ] SafeGraph Mobility
- [ ] Symptom Surveys
- [ ] USAFacts Cases and Deaths

benjaminysmith:
Ideas from @sangwon-hyun as extension to basic coverage:

- In the coverage graph, it might be helpful to visualize more than a snapshot e.g. which counties we used to have data in, but no longer do.

[...]

PR #133, opened 2020-10-21
Add Quidel, Doctor Visits, Hospital Admissions, GHT R dashboards to the example notebooks

benjaminysmith:
Per #32, this adds a basic coverage and sample size dashboard for a number of indicators, using our own API.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970301663177) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/R-notebooks/dashboard_functions.R
+++ b/R-notebooks/dashboard_functions.R
@@ -1,9 +1,28 @@
-plot_28_day_frequency <- function(df_to_plot, geo_type) {
-  # These are all required for rendering using covidcast.plot
-  df_to_plot$time_value = "2020-04-15"
-  df_to_plot$issue = "2020-04-15"
-  attributes(df_to_plot)$geo_type = geo_type
-  class(df_to_plot) = c("covidcast_signal", "data.frame")
+make_covidcast_signal <- function(destination, source, geo_type) {
+  destination$time_value = source$time_value[1]
+  destination$issue = source$issue[1]
+  attributes(destination)$geo_type = geo_type
+  class(destination) = c("covidcast_signal", "data.frame")
+  return(destination)
+}
+
+plot_28_day_frequency_state  <- function(df_to_plot) {
+  states_present = df_to_plot %>%
+    group_by(geo_value) %>%
+    summarize(value = n())
   
-  plot(df_to_plot, range=c(0,28))
+  covidcast_signal_to_plot = make_covidcast_signal(states_present, df_to_plot, "state")
+  
+  plot(covidcast_signal_to_plot, range=c(0,28))
+}
+
+plot_28_day_frequency_county  <- function(df_to_plot) {
+  counties_present = df_to_plot %>%
+    group_by(geo_value) %>%
+    summarize(value = n()) %>% ungroup() %>%
+    filter(substr(geo_value, 3, 5) != "000")
+  
+  covidcast_signal_to_plot = make_covidcast_signal(counties_present, df_to_plot, "county")
+  
+  plot(covidcast_signal_to_plot, range=c(0,28))
 }
--- a/R-notebooks/dashboard_functions.R
+++ b/R-notebooks/dashboard_functions.R
@@ -1,0 +1,9 @@
+plot_28_day_frequency <- function(df_to_plot, geo_type) {
+  # These are all required for rendering using covidcast.plot
+  df_to_plot$time_value = "2020-04-15"
+  df_to_plot$issue = "2020-04-15"
+  attributes(df_to_plot)$geo_type = geo_type
+  class(df_to_plot) = c("covidcast_signal", "data.frame")
+  
+  plot(df_to_plot, range=c(0,28))
+}
--- a/R-notebooks/dashboard_functions.R
+++ b/R-notebooks/dashboard_functions.R
@@ -14,7 +14,10 @@ plot_28_day_frequency_state  <- function(df_to_plot) {
   covidcast_signal_to_plot = make_covidcast_signal(states_present, df_to_plot, "state")
   
   plot(covidcast_signal_to_plot,
-       title = "State frequency in last 28 days",
+       title = sprintf(
+         "State frequency in last 28 days (start date: %s)",
+         covidcast_signal_to_plot$time_value[1]
+       ),
        range = c(0, 28))
 }
 
@@ -26,7 +29,12 @@ plot_28_day_frequency_county  <- function(df_to_plot) {
   
   covidcast_signal_to_plot = make_covidcast_signal(counties_present, df_to_plot, "county")
   
-  plot(covidcast_signal_to_plot,
-       title = "County frequency in last 28 days",
-       range = c(0, 28))
+  plot(
+    covidcast_signal_to_plot,
+    title = sprintf(
+      "County frequency in last 28 days (start date: %s)",
+      covidcast_signal_to_plot$time_value[1]
+    ),
+    range = c(0, 28)
+  )
 }
--- a/R-notebooks/dashboard_functions.R
+++ b/R-notebooks/dashboard_functions.R
@@ -13,7 +13,9 @@ plot_28_day_frequency_state  <- function(df_to_plot) {
   
   covidcast_signal_to_plot = make_covidcast_signal(states_present, df_to_plot, "state")
   
-  plot(covidcast_signal_to_plot, range=c(0,28))
+  plot(covidcast_signal_to_plot,
+       title = "State frequency in last 28 days",
+       range = c(0, 28))
 }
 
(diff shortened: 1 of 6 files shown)


ryantibs:
@benjaminysmith Nice start!  

From Slack: 

> Katie  4 hours ago
> I’ve tagged Ryan since he’s the correlations czar but we should probably get him an understudy. Daniel James Sharpnack thoughts?
> 
> James Sharpnack  3 hours ago
> No one comes to mind, although this should probably be thought of as a DAP.  Any of the senior statisticians can act as advisors in that case, Larry Daniel Ryan Valerie Ventura Jacob Bien...

I don't see any correlations being computed, I assume that was intentional @benjaminysmith?  Was the idea to minimize overlap with the correlations notebook?  If not and it was just an oversight I suggest adding it (and I also don't see the downside to having it in both places, as I did with the FB survey dashboard).

More broadly I don't think this is data analysis per say so I wouldn't treat this as a DAP.  I think it's just: what are the right dimensions to slice, to get some basic info & sanity checks on the behavior of our signals?  In that sense I think it'd be good to a get a light review from two people, one person who knows the technical aspects of the signals well and the other to give higher-level guidance about what we could show.  Say @brookslogan @zackchase @capnrefsmmat (and two of three would be great).

benjaminysmith:
>> I don't see any correlations being computed, I assume that was intentional @benjaminysmith?

Yeah, that's intentional. The purpose here is to get some simple, skeleton dashboards across all of our signals so that: 1) we can do a quick visual sanity check if things are weird; 2) serve as simple examples for accessing the signals.

I think there are likely other useful things to include in each (and there is also refactoring we can do to share common patterns, but those are easier to spot once we have a handful of examples we like) so the things that would be helpful from a review are:

1. Anything obviously wrong or things we can simplify

ryantibs:
Sounds good thanks for the update.  All makes sense.  Re 1: I do think correlations are valuable: seeing correlations plummeting seems like a telltale sign that something bad (or unexpected) is happening.

Let me amend my suggestions for reviewers a bit.  Any of @capnrefsmmat @brookslogan @capolitsch @JedGrabman @nmdefries @sangwon-hyun @nloliveira should be able to review it quickly and painlessly.  (Removing Zack since he's a Python guy 😱!)

sangwon-hyun:
I looked at just the facebook dashboard; looks great, but some comments:
- In the coverage graph, it might be helpful to visualize more than a snapshot e.g. which counties we used to have data in, but no longer do.
- In the first two panels (under "sampling") the decrease in FB counties and number of response seems significant. Should we be worried?
image

- It could be helpful to put a basic anomaly detection on these two graphs:
   + Whether the most recent 30 days are trending down, according to a best fit line's slope. 
   + Whether there is an abrupt drop (2 or 3 standard deviations?) compared to the most recent 10 day average?

benjaminysmith:
Thanks for the review Sangwon!

I really like your suggestions. In general I am targeting with starting with simple coverage dashboards at first (so we have somewhere quick to look at the basic stats) and then we can add more interesting stuff on top of those. I'm particularly interested in having some simple anomaly detection shown visually.

What do you think about checking in the simple ones first and then tracking the extensions you suggested as extensions in the main issue? I'll add them there now so we don't lose them. That will make it easier for others to add code since we will have a base to start with.

benjaminysmith:
Oh, and in regards to coverage drop for FB surveys, I know @krivard is aware of this, but not sure if we have a separate tracking issue.

krivard:
Yep, coverage drop is known and being tracked by both me and @capnrefsmmat. Hypothesis is some combination of population saturation and decline in general interest; there's nothing amiss in Facebook's sampling machinery that anyone over there can tell.

capnrefsmmat:
Yes, I like the idea of deploying simple dashboards now and filing issues to track subsequent improvements. Incremental changes are easier to work with than one big huge pull request.

sangwon-hyun:
Sounds good to me!

capolitsch:
These look great to me, @benjaminysmith. Are there plans to unify the simple dashboards (e.g. tabs)? I would personally also find it helpful to be able to adjust the time window with a slider (i.e. allowing for more than just 28-day).

R-notebooks/dashboard_functions.R
```diff
@@ -0,0 +1,9 @@
+plot_28_day_frequency <- function(df_to_plot, geo_type) {
```
JedGrabman, line 1: It looks like you do a lot of the same processing steps before passing it to this function. Since plot.covidcast works on covidcast_signal objects, it might be cleaner to have a function that does the data manipulation, returns the data.frame and then just plot that.

R-notebooks/dashboard_functions.R
```diff
@@ -0,0 +1,9 @@
...
+  # These are all required for rendering using covidcast.plot
+  df_to_plot$time_value = "2020-04-15"
```
JedGrabman, line 3: Right now, all plots are labeled with 2020-04-15. You can use the data from the covidcast_signal objects to set an appropriate value.

R-notebooks/ght_dashboard.Rmd
```diff
@@ -0,0 +1,92 @@
...
+  date_scale
+```
+
+## State Coverage Map
+
+This visualizes the state coverage -- how frequently does each state show up in the data over the last 28 days?
+
+```{r, fig.width = 10, fig.height = 8}
+
+states_present = df_ght_states %>%
```
JedGrabman, line 86:
```
group_by(geo_value) %>%
summarize(value = n())
```
seems like a common pre-processing step that can be moved to the function.

benjaminysmith:
Adding tabs is a fantastic idea for a usability improvement. It looks like there is some support for this using R markdown (https://bookdown.org/yihui/rmarkdown-cookbook/html-tabs.html). Since we can also hack together something from the individual HTML outputs I'm leaning to starting with that first and then exploring using tabs in R when we are happy with the individual ones.

R-notebooks/make.R
```diff
@@ -1,7 +1,6 @@
 file_names = list.files(".", pattern="*.Rmd")
 file_names = substr(file_names, 1, nchar(file_names)-4)
 
-file_name = "fb_dashboard"
 for (file_name in file_names) {
   cat(sprintf("Rendering %s ... ", file_name))
```
JedGrabman, line 5: fyi, fb_dashboard took a while to run and broke when I tried this (but I may not have had a clean environment). Doesn't look like you touched it though.

JedGrabman (commented):
Analysis looks good to me, except that the plot titles have an inaccurate date.

I also made a few comments about doing the plotting in a cleaner way by having a function do the data processing instead of the plot. But, also ok with the quick version of just passing in the date, since this isn't production code.

R-notebooks/make.R
capolitsch: I had the same issue

sangwon-hyun:
FYI Using tabs in R markdown is https://bookdown.org/yihui/rmarkdown-cookbook/html-tabs.html done using {.tabset} to the section header that is one level higher than the headers to be converted to tabs.

R-notebooks/make.R
benjaminysmith: I can reproduce this -- will follow up separately since it's not affected by this PR.

benjaminysmith: Tracking in #211

R-notebooks/dashboard_functions.R
benjaminysmith: I refactored more common code in here based on later comment -- is this what you meant, or do you have another refactoring idea?

R-notebooks/dashboard_functions.R
benjaminysmith: Done (though honestly I think the issue date is meaningless after grouping -- would be nice to consider support plotting without an issue date set)

R-notebooks/ght_dashboard.Rmd
benjaminysmith: Done.

R-notebooks/dashboard_functions.R
JedGrabman: I think my idea was slightly different, but this looks good

R-notebooks/dashboard_functions.R
JedGrabman: Ah, my intent was mainly to update the date in the plot titles. Could you add it in addition to "last 28 days", so that it's easy to see how old the graphs are?

R-notebooks/make.R
JedGrabman: Thanks for creating the bug

JedGrabman (approved):
All my comments are resolved, so looks good to me.

PR #140, opened 2020-10-22
Fix weighted_interval_score assert statement

JedGrabman:
Ensure assert_that argument has length 1. 
Fail instead of succeed when a quantile is NA,


jacobbien (commented):
Thanks for fixing this.

PR #143, opened 2020-10-22
Add custom warning to be more informative when getting data

### LINKED ISSUE #139: Make covidcast.signal give more specific warnings when data is not available
See https://github.com/cmu-delphi/covidcast-indicators/pull/343#discussion_r510188822

We should augment the warnings with source, signal, date, and geo_type, so anyone logging can find this out. (We already did this in the R package by raising a custom warning condition that can be handled appropriately with a logger.)

chinandrew:
Closes #139 

Summary of changes:
- Implement NoDataWarning custom warning for when no results are returned
- Change any other warnings to runtime warnings and raise them separately
- Add tests, and also fix two previous tests that weren't testing the right thing (oops)


capnrefsmmat (approved):
Looks good

Issue #112, opened 2020-10-06
data corrections within evalcast


jacobbien:
Incorporate the Delphi data corrections into evalcast.  This should probably always only be offered as an option that can be turned on/off rather than always being on.

dajmcdon would be the best person to do/direct this.

Data correction is relevant at various stages:

1. when predictions are made by a forecaster using `evalcast::get_predictions`
1. when predictions are being evaluated using `evalcast::evaluate_predictions`.

Both of these functions could have a flag indicating whether data corrections should be used.

One approach would be to integrate data correction directly into [`evalcast::download_signals`](https://github.com/cmu-delphi/covidcast/blob/evalcast/R-packages/evalcast/R/download_signals.R), so that whenever data is downloaded from the API, it comes with corrections (with a flag determining whether this occurs or not).

Something to consider in implementing this: An argument can be made that the "as of" principle should apply to the data corrections principle.  So in making predictions, a forecaster shouldn't have access to a correction that wouldn't have been available until a later date.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970439260183) by [Unito](https://www.unito.io/learn-more)

dajmcdon:
Yeah, I'll take this one.

dajmcdon:
I think this interacts with #107 . Ideally there, we want to memoise the `download_signals()` call. That will cache the api data. Then for corrections, we make them immediately after based on the flags in `get_predictions()` and `evaluate_predictions()`. We memoise those too (so they can be retrieved later if desired, say with another function).

As for the as_of, I'm not quite sure how to impose the right guardrails. In the past, corrections were little more than `dplyr::mutate()` applied to the signals, and my intention was to stay the same. Since `download_signals()` obeys `as_of`, the corrections would as well, unless the user intentionally passed a function that was based on future data. 

I'll proceed along these lines if that's ok with you.

jacobbien:
Hi @dajmcdon.  Yes this sounds good to me.  I agree with what you said about the corrections process (as you're doing it) not violating the as of principle.  If a user is determined to break the as of guardrails, it can certainly be done in many ways... but my main concern is that the basic user using the basic functionality and options will not be led to violate this principle.

Issue #107, opened 2020-10-06
cache previously downloaded data


jacobbien:
To avoid excessive numbers of repetitive calls to covidcast (and for improved performance), we can cache data that has been downloaded by evalcast.  We just need to be sure that it's an identical query to covidcast (not just the same signal and start/end dates but also -- and this is important -- the same as_of).

To do this, one would modify [`evalcast::download_signal`](https://github.com/cmu-delphi/covidcast/blob/evalcast/R-packages/evalcast/R/download_signals.R).

This feature was suggested by jsharpna.  The person implementing this might want to talk with bnaras about how best to implement caching in R.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970444952578) by [Unito](https://www.unito.io/learn-more)

dajmcdon:
I'll take this

ryantibs:
@jacobbien @dajmcdon @bnaras Some comments from me, as I'm working through evalcast now ... and starting to use it for some simple analyses.

First, [this explanation](https://www.r-bloggers.com/2020/04/caching-in-r-2/) for caching options in R seems pretty helpful.  I could imagine a few ways to use `memoise()` or `mustashe()` it in the context of evalcast.

Second, let me express my enthusiasm for caching being implemented sooner than later.  As a user of evalcast 🙂 my experiments would benefit from a speedup (currently taking a while to run).

Third, even apart from caching, let me point out that a bigger deal could be to change the order of operations.  **I'm not suggesting we do this now**, but I want to point it out because it's relevant.  Maybe it's worth pursuing after caching gets implemented, as we can assign a separate issue to this.  Here goes:

Most interesting forecast analysis involve comparisons between forecasters.  Right now `get_predictions()` and `evaluate_predictions()` each only operate on **a single forecaster at a time**.  The way it's implemented now, we could easily extend `get_predictions()` or `evaluate_predictions()` so that could take multiple forecasters (would just be a simple wrapper over the current implementations).  But this would be inefficient, because if I had a lot of forecast dates, there would either be a ton of duplicated data fetching. Assuming we're always considering the same forecast dates across forecasters (the most common case I believe) ... for `get_predictions()`, there would be duplication if my forecasters used highly overlapping data sets for training, and for `evaluate_predictions()`, there would always be duplication.  

[...]

Issue #144, opened 2020-10-22
Problem when using import dependency on covidcast


bnaras:
The variable `state_census` is not available to `covidcast::name_to_abbr()` when the package is imported rather than attached via `library()` (reprex below). The fix is to use `R/sysdata.rda` for internal use. But it appears these datasets are exported? If they have to be so, perhaps some convenience functions for access?

```{r}
> covidcast::name_to_abbr("Penn")
Error in eval(lhs, parent, parent) : object 'state_census' not found
```

capnrefsmmat:
Yes, `state_census` (and `county_census` and similar) is exported, since we imagine covidcast users may want the data in convenient form for other analyses.

Is it possible to use something like [this trick](https://stackoverflow.com/a/16355495) to ensure the data gets into the package's namespace, so its functions can use it, while still exporting it? It seems quite odd that R doesn't have a mechanism for datasets used both internally and exported.

bnaras:
Yes, that trick will work for now as it is only sticking data objects in the package namespace although I rarely see it in packages. (I generally don't like to muck around with parent environments etc. in `.onLoad`.) The alternative would be to have functions for each exported data set, which is not ideal. So the trick is ok, I'd say, with some comments thrown in case problems arise in the future. The documentation for `data()` states it was designed before lazy loading became common, so it has warts.

ryantibs:
@capnrefsmmat @bnaras Am I missing something ... isn't this fixed by just referring to the data objects within the covidcast package code itself as in `covidcast::state_census`, etc.?

Or will this create some kind of weird recursion issue?

bnaras:
@ryantibs @capnrefsmmat I like your solution: simple and not much work. Apparently it is also CRAN approved. https://coolbutuseless.github.io/2018/12/10/r-packages-internal-and-external-data/

bnaras:
@ryantibs @capnrefsmmat I'll make a PR in the next 10 min to fix this.

ryantibs:
Closed by #148.

Issue #98, opened 2020-10-02
Make predictions cards and score cards S3 objects with print and plot functions


jacobbien:
Currently they are tibbles, i.e. of class

`"tbl_df"     "tbl"        "data.frame"`.

Instead, we'd have them inherit from the above but be their own class.  For example, a predictions card would have class

`"predictions_card"     "tbl_df"     "tbl"        "data.frame"`

and then

* `print.predictions_card` would print out some of the attribute information (for example, name of the forecaster, `forecast_date`, `ahead`, etc.)
* `plot.predictions_card` would make the trajectory plots

Same idea for scorecards.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970448218457) by [Unito](https://www.unito.io/learn-more)

jacobbien:
To clarify, `plot.predictions_card` would probably just call the function described here: #101.

bnaras:
I can take this too.

jacobbien:
Great!

jacobbien:
Hi @bnaras!  I'm responding to the part of what you wrote in #150 that has to do with the S3 methods.  In my opinion, the main impetus for this would be the print method.  In particular, I find myself printing a list of predictions cards (which currently prints a list of tibbles) and then I want to know what the forecaster, ahead, and geo_type of these are.

For example, consider the following code:
```
forecast_dates <- evalcast::get_forecast_dates("CMU-TimeSeries")[1:4]
pc_cmu <- evalcast::get_covidhub_predictions("CMU-TimeSeries", forecast_dates)
```
This gives a list of 20 predictions cards, and printing `pc_cmu` gives me little information.  It'd be helpful if in addition to the tibble, I get to see something like
```
forecaster name: CMU-TimeSeries; ahead: 2; incidence_period: "epiweek"; geo_type: "state"; forecast_date: 2020-10-01
```
And for a score card, we'd like to know the as_of date.

jacobbien:
I'm responding to the S3 aspect of this discussion in issue #98 just to keep things organized.

jacobbien:
[...]

PR #145, opened 2020-10-23
Evalcast corrections

dajmcdon:
This should fix both #107 and #112 . 

I added a function argument to `get_predictions()` that implements corrections. For the moment, it's not clear to me that a similar method would work at evaluation time since we won't necessarily have access to most of the data. I put a rather dumb example in the examples (probably shouldn't run at check time).

Also implemented caching via `memoise`. Currently just to memory. The alternatives seem rather diverse and hard to anticipate (at least to me). Also put a unit test to make sure that it worked on two consecutive calls to `get_predictions()` with the same arguments.

Everything passes local checks with some warnings. Most are for undocumented function arguments. Plus all the stupid dplyr notes.


R-packages/evalcast/tests/testthat/test-caching.R
```diff
@@ -0,0 +1,32 @@
...
+        signal = c("deaths_incidence_num","confirmed_incidence_num"),
+        start_day=lubridate::ymd("2020-09-15")),
+      lubridate::ymd("2020-10-01"),"epiweek", 1L, "state", "mi"
+    )
+  )
+)
+
+
+test_that("we are caching COVIDCAST calls", {
+  expect_true(t1[1] > t2[1])
```
jacobbien, line 31: This does seem like a reasonable way to test that caching is working.  But I'm wondering if there could be cases where caching occurs but ends up being slower for some reason.  Is there a more direct way to verify that caching occurred? @bnaras, do you have thoughts on this?

R-packages/evalcast/data-raw/county_label_lookup.R
```diff
@@ -0,0 +1,5 @@
+## code to prepare `county_label_lookup` dataset goes here
+county_label_lookup <- paste(
```
jacobbien, line 2: Where is `county_label_lookup` used?

R-packages/evalcast/R/get_predictions.R
```diff
@@ -88,6 +117,10 @@ get_predictions_single_date <- function(forecaster,
                       geo_type = geo_type,
                       geo_values = geo_values)
     })
+  if(!is.null(apply_corrections)){
+    corrected <- apply_corrections(df)
+    df$value = corrected$corrected
```
jacobbien, line 122: If a user-supplied `apply_corrections` were to reorder the rows, then this could be a problem.  Perhaps a `left_join` would be safer?

R-packages/evalcast/data-raw/county_label_lookup.R
dajmcdon: I'm not sure where this came from. I must have screwed up the git. I think this script and the associated data file below should just be deleted.

R-packages/evalcast/tests/testthat/test-caching.R
jacobbien: Also, perhaps you could add a test that the predictions cards from the two runs are identical?

jacobbien:
Thanks, @dajmcdon, this is very nice.  I have a few more comments:
1. I agree that we'd want to add the corrections within the `get_target_response` function, but can you explain what the difficulty you allude to would be with this?
2. The attributes for a predictions card should have all the information needed to reproduce the call.  I believe we should therefore add an attribute that is the `apply_corrections` function that was used with `NA` (I suppose) in the case that no corrections were used.
3. Likewise, when we add corrections to the evaluate part, we should add an attribute `apply_corrections` to the scorecard
4. Then when downstream functions that take in a list of scorecards, we will need to make sure that the same corrections were used for scorecards in the list.  The one difficulty I could foresee here is if `apply_corrections` involves randomness.  I think one should discourage that since we don't want forecasters compared with each having its own ground truth data.  Actually, the simplest thing might be to just check that the response column in each scorecard being compared is in perfect agreement.

R-packages/evalcast/R/get_predictions.R
dajmcdon: I made a number of changes to address this, adding checks and a test. Currently all tests fail due to the `covidcast::name_to_abbr()` bug (see #144, fixed by @ryantibs on #146 ).

R-packages/evalcast/tests/testthat/test-caching.R
dajmcdon: I added a test to check the prediction cards (both the attributes and the values). There's a note there about the attribute check for @bnaras and #98 . As for the timing check, I added a direct test to make sure that the caching is enabled. Waiting to push until I can try it locally.

dajmcdon:
@jacobbien, some comments on your comments.

1. I guess I was imagining (perhaps incorrectly), that if one wanted to correct the evaluation data, the desired outcome would be the same as in the training data (or input data). For example, we use a rolling window. But based on the download scheme in `get_target_period()`, it doesn't seem like we could (since only the response over the forecast period is downloaded).  But maybe I'm wrong about how it works. We would also be unable to make corrections that depend on any signals except the response. So, it seems like there are some structural constraints on the types of corrections that can be made here that don't exist at the prediction stage. We could try to explain all the constraints to the user, or we could make more elaborate alterations. Say there's just one forecast date, download up to that date  for two `as_of`s, apply the corrections to everything and then load the memoised second `as_of` at evaluation. This all seemed fairly complicated, so I figured I'd ignore it for now. But maybe we need it?
2. I think I fixed this. But only in `get_predictions()`. I've been unable to find if they get set elsewhere checked against (in which case, I broke something).
3. The corrected prediction card attribute seems to propagate, but point 1 above.
4. Our own corrections ARE random (I abandoned the seed setting some time back). So this would pose a problem. Though, could be fixed by referencing the same corrected data for all forecasters rather than downloading separately. 

Relatedly, we have been storing the actual corrected values, not just the function to generate them. I was thinking that in our pipeline, I would make the `apply_corrections` function we pass in store the results for visualization, etc after the run. But maybe we actually want to "download and store" all the data necessary at once (though this sounds like just building upstream_df). That would allow the corrections but break lots of other things.

R-packages/evalcast/tests/testthat/test-caching.R
bnaras: I pushed a fix to #144 so that you can try your tests.  I am not sure building the caching directly into the function is the right approach as that might screw up experiments where real-time changes may take place to the data. It seems it would be better if the user is given the option of using caching by means of a configuration setting...

bnaras: @jacobbien To answer your question about checking if caching occured,  if one uses the file system, there must be objects in the file system one can look at for the timestamp...

dajmcdon: I haven't been able to get it to work. There are two commented out tests that use `memoise::has_cache()` to actually examine the cache. One issue with the current approach (supporting @bnaras's point) is that the in-memory version is hard to access. But it greatly simplifies the package development. We could prompt on load to warn of the behavior and suggest caching locally instead. This `memoise` creates unintuitive file names though.

dajmcdon:
I'm not sure where this got left. For the moment, we need the corrections if we're going to run our pipeline off of this.

Alternatively, I can put a fork somewhere.

jacobbien:
1.  Thanks for this explanation.  I had forgotten that `apply_corrections()` needs to take in the history of the variable (and even other variables) in order to decide what corrections to make.  However, the good news is that if I'm not mistaken things may be easier than you are describing.  When downloading data within `evaluate_predictions()`, we don't have to worry about the "as of principle" ...our goal is to get the ground truth data as best we know it as of the current time we are running `evaluate_predictions` (not as of the forecast date).  This means that the `apply_corrections()` function could on its own download whatever it wants internally.  To be clear, this corrections function would operate by different rules from the `apply_corrections()` function we would use in the `get_predictions()` so for clarity we would want to call this one something different.  It could take in that data frame in order to know what target period dates/locations we care about and then do its own thing using all the information it wants to make the best corrections possible.

2. Thanks.  This is why we need more unit tests.  I've written some, but we should have more for cases like this.

3. Right.  Can't address this part until we decide on 1.

4.  For reproducibility it seems that setting a seed would always be desirable, but I'm probably missing something here.

Yeah, I can understand the logic of wanting to store the corrections themselves.

PR #147, opened 2020-10-23
Add CI for R package

### LINKED ISSUE #142: Enable CI for our R packages
We should extend the CI workflows (in `.github/workflows/`) to build and test our R packages. This should be pretty straightforward (just do `R CMD BUILD` and also `CHECK`), except for the covidcast package we probably want to disable the vignettes, since they are quite slow.

This will be particularly useful after #124, since the CI can run our tests and tell us if PRs fail the tests, but it'd also be useful before then to ensure the packages build cleanly and there aren't check warnings.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970432617506) by [Unito](https://www.unito.io/learn-more)

chinandrew:
Closes #142

Summary of Changes:

* Add CI for R `covidcast` package (no evalcast yet)

Notes

* The check currently only fails if there are errors, and will not fail on warnings. There is currently one warning in the code.
* The CI config was pulled from the github actions premade template, which was mostly taken from the [r-lib](https://github.com/r-lib/actions/tree/master/examples#quickstart-ci-workflow) repo. I've changed it to not build vignettes, run on just one version of R, and on linux instead of mac (which is 10x cheaper).
* [RESOLVED by caching dependencies] It takes 6 minutes to install dependencies, which is quite a long time, and the check itself is <1min. I'm not sure if this can be optimized. That amount of time is a bit annoying to merge things fast and also drives up [compute minutes (and costs once we use up the free tier)](https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions) by almost an order of magnitude. I don't know how close we are to using up all the free CI time, but this would certainly add up. It shouldn't be hugely expensive if we do start paying (see pricing in the previous link), but still worth calling out.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970431945681) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/.github/workflows/r_ci.yml
+++ b/.github/workflows/r_ci.yml
@@ -39,5 +39,5 @@ jobs:
         shell: Rscript {0}
       - name: Check
         run: |
-          rcmdcheck::rcmdcheck(args = c("--no-manual", "--ignore-vignettes"), build_args = c("--no-build-vignettes"), error_on = "error")
+          rcmdcheck::rcmdcheck(args = c("--no-manual", "--ignore-vignettes", "--as-cran"), build_args = c("--no-build-vignettes"), error_on = "error")
         shell: Rscript {0}
--- a/.github/workflows/r_ci.yml
+++ b/.github/workflows/r_ci.yml
@@ -34,7 +34,7 @@ jobs:
         run: sudo apt-get install libcurl4-openssl-dev
       - name: Install dependencies
         run: |
-          install.packages(c("remotes", "rcmdcheck"))
+          install.packages(c("remotes", "rcmdcheck", "devtools"))
           remotes::install_deps(dependencies = TRUE)
         shell: Rscript {0}
       - name: Check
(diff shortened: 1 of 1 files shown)


chinandrew:
On installs, looks like `stringi` takes the longest to install at nearly a minute, but the sheer volume of packages (~70, most taking 2-10s to install) is causing the problem. Github has some [caching functionality](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows) to save dependencies, though not sure what happens when dependencies get updated, we may have to manually recache. Will look into it

chinandrew:
Build times brought down to <2 min by caching dependencies.

capnrefsmmat (approved):
Looks good to me. Look forward to preventing some bugs this way (and adding tests!)

PR #148, opened 2020-10-23
Issue 144

bnaras:
- Addressed issue #144 using Ryan's suggestion by prefixing all internal uses of data objects with `covidcast::`.
- Added file `.Rbuildignore` to prevent problems when building package.
- Bumped minor version to 0.3.1


ryantibs (approved):
Looks good to me, think you got them all!

Issue #63, opened 2020-09-03
Add flexible warning option for production pipeline in the R package


bnaras:
The upstream data generation pipeline relies on accurate warnings to detect anomalies. Currently, if the process generates warning, I have to explicitly request to see the warnings at the end of the run.  It would really help to send the warnings to a logger as they happen.

The code below, which can be placed in `covidcast/R/zzz.R` as is, allows a user to set an option `covidcast.warning`. If not set, it falls back to the standard `base::warning`. In addition, each call to `warning("bad", "stuff", ...)` has to be replaced with `getOption("covidcast.warning")("bad", "stuff", ...)`, which is a simple search and replace.  Example below. Happy to provide a PR if needed.

## File `zzz.R`
```{r}
## Set up flexible logging options on load
.onLoad <- function(libname, pkgname) {
    opts <- options()
    opts.covidcast <- list(
        ## Default warning function is the base warning function
        covidcast.warning = base::warning
    )
    toset <- !(names(opts.covidcast) %in% names(opts))
    if (any(toset)) {
        options(opts.covidcast[toset])
    }

    invisible()
}
```
## Example

```{r}
> library(covidcast)
We encourage COVIDcast API users to register on our mailing list:
https://lists.andrew.cmu.edu/mailman/listinfo/delphi-covidcast-api
We'll send announcements about new data sources, package updates,
server maintenance, and new features.
> m  <- covidcast_signal("fb-survey", "raw_cli", geo_type = "msa",
                  geo_values = name_to_cbsa("Pittsburgh"))

Fetched day 20200406: 1, success, num_entries = 1
...
Fetched day 20200901: 1, success, num_entries = 1
Warning message:
In single_geo(data_source, signal, start_day, end_day, geo_type,  :
  Fetching raw_cli from fb-survey for 20200412 in geography '38300': no results
```

Now set option.

```{r}
 options(covidcast.warning=logger::log_warn)
> m  <- covidcast_signal("fb-survey", "raw_cli", geo_type = "msa",
                  geo_values = name_to_cbsa("Pittsburgh"))
Fetched day 20200406: 1, success, num_entries = 1
Fetched day 20200407: 1, success, num_entries = 1
Fetched day 20200408: 1, success, num_entries = 1
Fetched day 20200409: 1, success, num_entries = 1
Fetched day 20200410: 1, success, num_entries = 1
Fetched day 20200411: 1, success, num_entries = 1

[...]

PR #151, opened 2020-10-26
Make use of new support for multiple geo_values in one request

JedGrabman:
Send multiple geo_values in comma delimited string
Request is returned as successful if any data retrieved, so treat day/geo_value combos with non-retreived data as warnings

Also, update error message to use human readable dates

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970431397793) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/R-packages/covidcast/R/covidcast.R
+++ b/R-packages/covidcast/R/covidcast.R
@@ -499,12 +499,18 @@ covidcast_days <- function(data_source, signal, start_day, end_day, geo_type,
                     nrow(dat[[i]]$epidata)))
     if (dat[[i]]$message == "success") {
       returned_geo_values <- dat[[i]]$epidata$geo_value
-      missed_geos <- setdiff(geo_value, returned_geo_values)
+      missed_geos <- setdiff(tolower(geo_value), tolower(returned_geo_values))
       if (length(missed_geos) > 0) {
         missed_geos_str <- paste0(missed_geos, collapse = ", ")
         warn(message =
                sprintf("Data not fetched for some geographies on %s: %s",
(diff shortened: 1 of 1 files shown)


R-packages/covidcast/R/covidcast.R
```diff
@@ -628,38 +626,49 @@ summary.covidcast_meta = function(object, ...) {
...
-                  " in geography '", geo_value, "': ", dat[[i]]$message),
+    if (dat[[i]]$message == "success") {
+      returned_geo_values <- dat[[i]]$epidata$geo_value
+      missed_geos <- setdiff(geo_value, returned_geo_values)
+      if (length(missed_geos) > 0) {
+        missed_geos_str <- paste0(missed_geos, collapse = ", ")
+        warn(message =
+               sprintf("Data not fetched for some geographies on %s: %s",
+                       query_day, missed_geos_str)
+             )
```
capnrefsmmat, line 663: Just like the `warn` call on line 666, could you augment this one with extra parameters stored in the warning object? Plus a distinct class, like `covidcast_missing_geos` or something?

See #63 for motivation. The forecaster code will be able to catch the warning, extract appropriate data from it, and log the problem so the forecasters know what is missing.

capnrefsmmat (changes requested):
Looks fine to me, just one request about the warning so it's easier to log.

R-packages/covidcast/R/covidcast.R
JedGrabman: Done.

capnrefsmmat (approved):

PR #152, opened 2020-10-26
Reorder Python requirements in makefile

chinandrew:
Summary of changes:
- Have dev requirements install before package dependencies. 

I can't reproduce this anymore, but one time I had a warning or error thrown when installing dependencies because wheel wasn't installed. Kinda weird but this should solve it.


krivard (approved):
lgtm

PR #155, opened 2020-10-27
Create Evalcast Vignette

JedGrabman:
A vignette for evalcast modeled on a draft from jacobbien

Rendered version is available [here](https://htmlpreview.github.io/?https://github.com/JedGrabman/covidcast/blob/vignette/R-packages/evalcast/vignettes/intro_vignette.html)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970428645954) by [Unito](https://www.unito.io/learn-more)


dajmcdon (approved):
Looks good @JedGrabman . I'm going to approve and get this off the to-do list.

R-packages/evalcast/vignettes/intro_vignette.Rmd
```diff
@@ -0,0 +1,132 @@
...
+signals
+```
+
+### Run the forecaster on some dates
+
+Within the `evalcast` framework, forecasters do not directly retrieve historic data. Instead, `get_predictions` provides *only the data that would have been available at the time* to the specified forecaster in order to generate retrospective predictions.
+
+```{r}
+library(evalcast)
+library(covidcast)
```
jacobbien, line 36: I think we can remove this now.  (In an earlier version of covidcast this was needed, but that issues has been resolved.)

R-packages/evalcast/vignettes/intro_vignette.Rmd
```diff
@@ -0,0 +1,132 @@
...
+library(lubridate)
+
+forecast_dates <- as_date(get_forecast_dates("CMU-TimeSeries"))
+forecast_dates_july <- forecast_dates[forecast_dates >= "2020-07-01" & forecast_dates <= "2020-07-31"]
```
jacobbien, line 40: I think this would be cleaner:
```
forecast_dates_july <- str_subset(forecast_dates, "^2020-07")
```
Also, we might consider going more recent at this point than July.  Perhaps September or October?

R-packages/evalcast/vignettes/intro_vignette.Rmd
```diff
@@ -0,0 +1,132 @@
...
+
+# Retrieve past predictions from CovidHub
+predictions_cards_yyg <- get_covidhub_predictions("YYG-ParamSearch", forecast_dates_july)
```
jacobbien, line 43: Instead of "YYG-ParamSearch" (which is no longer submitting forecasts), I would suggest comparing to "COVIDhub-ensemble" and "COVIDhub-baseline".

R-packages/evalcast/vignettes/intro_vignette.Rmd
```diff
@@ -0,0 +1,132 @@
...
+
+```{r}
+predictions_cards[[1]]
+```
+
+Forecasts are probabilistic, meaning that rather than predicting a single value, a distribution of values is generated. For example, the baseline forecaster would have predicted the following distribution in California for `r attributes(predictions_cards_cmu[[1]])$ahead` week(s) after `r attributes(predictions_cards_cmu[[1]])$forecast_date`
+```{r}
+pc <- predictions_cards_cmu[[1]]
+ca_fips = covidcast::name_to_fips("California")
+ca_code = substr(ca_fips, 0, 2) # States are identified by a 2 character code
```
jacobbien, line 66: For consistency, I'd use `<-` for assignment rather than `=`.

R-packages/evalcast/vignettes/intro_vignette.Rmd
```diff
@@ -0,0 +1,132 @@
...
+                                     incidence_period = "epiweek",
+                                     ahead = 3,
+                                     geo_type = "state")
+```
+
+`baseline_forecaster` created a list of `r length(predictions_cards)` predictions cards, (one for each `(forecast date, ahead)` pair).  Let's look at a predictions card:
+
+```{r}
+predictions_cards[[1]]
+```
```
jacobbien, line 60: This would be a good place to demonstrate the function `plot_trajectory`.  Currently this is giving me an error when I try it on `predictions_cards`.  However, it works on `predictions_cards_cmu`.

R-packages/evalcast/vignettes/intro_vignette.Rmd
```diff
@@ -0,0 +1,132 @@
...
+
+Forecasts are probabilistic, meaning that rather than predicting a single value, a distribution of values is generated. For example, the baseline forecaster would have predicted the following distribution in California for `r attributes(predictions_cards_cmu[[1]])$ahead` week(s) after `r attributes(predictions_cards_cmu[[1]])$forecast_date`
+```{r}
+pc <- predictions_cards_cmu[[1]]
+ca_fips = covidcast::name_to_fips("California")
+ca_code = substr(ca_fips, 0, 2) # States are identified by a 2 character code
+pc[pc$location == ca_code, ]$forecast_distribution[[1]]
```
jacobbien, line 67: This is just style, but I personally prefer this:
```
pc %>% filter(location == ca_code) %>% pull(forecast_distribution)
```

R-packages/evalcast/vignettes/intro_vignette.Rmd
```diff
@@ -0,0 +1,132 @@
...
+scorecards[[1]]
+```
+
+**Technical note:** What does `backfill_buffer = 10` do?  When we evaluate a forecaster in backtesting we are assuming that we know what actually occurred.  However, in light of backfill, we may not trust the data for some period of time.  The argument `backfill_buffer` allows us to specify how many days until we believe that the data has "settled down" and is unlikely to be updated further.  The choice of this argument will depend on the particular signal you are forecasting.
+
+### Plots of performance measures
+
+`evalcast` provides a number of tools to visually assess a forecaster (including comparisons with other forecasters):
+
+```{r}
```
jacobbien, line 110: These plots all look truncated or squished.  I think you could use the `fig.width` and `fig.height` code chunk options to make this look better.

R-packages/evalcast/vignettes/intro_vignette.Rmd
```diff
@@ -0,0 +1,132 @@
...
+```
+
+
+## Cumulative forecasting
+
+While the above describes incident forecasting, the same `evalcast` functions can also be used for cumulative forecasting.
+
+- For example, for `k`-day-ahead cumulative forecasting, choose a cumulative signal from `covidcast` (e.g., `deaths_cumulative_num`), set `incidence_period = "day"` and `ahead = k`.
+
+- For example, for `k`-week-ahead cumulative epiweek forecasting, do the same as above but with `ahead = 7 * k`.
```
jacobbien, line 128: It'd be good if we actually carry out examples showing how this would work. (@dajmcdon, this is what I was referring to [here](https://github.com/cmu-delphi/covidcast/issues/351#issuecomment-737514603).)

jacobbien:
Looks good @JedGrabman .  I left all my comments above.  Thanks!

Issue #131, opened 2020-10-21
Derivative functionality


ryantibs:
Direction isn't being supported in the API anymore, and we decided to move all direction calculations over to the client side.

First, we should nix the references to direction in the R package entirely (and eliminate any functionality that uses it, like the plotting functionality).  I believe this was done for the Python package in #128.

We should offer functionality in the covidcast package(s) to compute _derivatives_ of signals, and then simple functionality to turn derivatives into _direction_ calls.  Here's my thoughts about how this should look.  This would be a fun issue to tackle because it's very concrete, and contains a nice mix of signal processing and design.  Tagging sgsmob dshemetov huisaddison chinandrew because they might be interested.  cc jsharpna capnrefsmmat krivard

At the outset, whether it should be done in both R and Python (or just R, and not Python): I'm not sure.  I'm going to write everything below in R notation but the main ideas should transfer of course to Python.

## Simple example
Let's start with a simple example, of what I would want to be able to do.  If I have a signal data frame `df`, then I should be able to call:

```
df_deriv = compute_deriv(df, method = "linear-reg", n = 14)
```

This would run, separately for each location in `df`, and each t = `time_value`, a linear regression of y = `value` on x = `time_value`, using last n = 14 days of training data (where x = t - 13, ... x = t).  It would then populate a new column called (say) `deriv` and append this to `df`, where the `deriv` column contains the fitted slope (the coefficient multiplying the dependent variable x in the regression).

## Available methods
For available methods, as specified through the `method` parameter, I would start by just allowing two: "linear-reg" (the default) for linear regression, and "smooth-spline" for a smoothing spline.  For the smoothing spline, the `deriv` column would be populate by the derivative of the fitted smoothing spline at x = t.  You can use [`smooth.spline()`](https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/smooth.spline), and then evaluate its derivative via [`predict.smooth.spline()`](https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/predict.smooth.spline).

Additional methods to consider for fun: if you implement trend filtering, we can celebrate with a cup of coffee or a glass of scotch.  If you implement wavelet smoothing, I will reject your PR.

## Other arguments
I would allow for the following arguments in `compute_deriv()`:

* `n`: the number of trailing days to use to train the model used to estimate the derivative. Default is (say) n = 14.
[...]

Issue #48, opened 2020-08-27
Add tools to rearrange data for analysis


capnrefsmmat:
For example, suppose I want to do a regression where I predict cases using various other signals. I need to obtain the signals, join them together, maybe lag some variables, and produce a data frame with one row per location and time, containing all measurements at that location/time.

We should provide a function, name to be determined, that

* takes a bunch of signal data frames
* joins them together and names the `value` columns after the signals
* optionally lags some variables, e.g. if I want to predict cases with lagged signals
* returns one big data frame

Then we should put a few examples in the package docs showing how we can use this to conduct a data analysis.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970461908005) by [Unito](https://www.unito.io/learn-more)

ryantibs:
@chinandrew I'm sorry to ask because it's a bit lazy but ... can you point me to exactly where you implemented this in the Python package so I can see the naming and argument specifications you chose?

I already have similar functionality sitting around in some analyses/demo notebooks (like the [forecasting demo](https://delphi.cmu.edu/blog/2020/09/21/can-symptoms-surveys-improve-covid-19-forecasts/)), so it would be easy for me to add this to the R package.

chinandrew:
This is in Python-packages/covidcast-py/covidcast.py in `aggregate_signals()` with documentation at the bottom of the first section here https://cmu-delphi.github.io/covidcast/covidcast-py/html/getting_started.html. 

We had a few discussions on the right interface (e.g. pandas has a multiindex DF) in #62, which might be Python specific but potentially worth a read (and you'll have to excuse my horrible late night writing). I don't think we ended up very opinionated about the implementations, so if you end up with good ideas let us know.

capnrefsmmat:
Yeah, I think in R we should be able to do this with just `rbind` and `dplyr::pivot_wider` and a few extra steps to clean up the metadata and column names.

chinandrew:
Closed by #156 , not sure why it didn't auto-close

PR #156, opened 2020-10-28
Multi signals

### LINKED ISSUE #10: Finish the covidcast_signals function
It's currently in the R package, but not exported. Remaining issues:

* [x] How should the sources/signals be specified? Currently it accepts a data frame, but Ryan thinks separate vectors with recycling would be convenient, so you could specify things like `data_source = "jhu-csse"` , and `signal = c("confirmed_incidence_num", "confirmed_incidence_prop")`.
* [x] Should it be possible to select signals with different `geo_types` in one call?
* [x] How should `plot.covidcast_signal()` handle data frames containing multiple signals? It could make a grid of plots, perhaps; faceting doesn't work easily because different signals are on different scales.

I also think we should provide a convenience function that wraps `pivot_wider` to convert a data frame with multiple signals to have one row per date and one column per signal, named after the signal, instead of one row per observation per signal. People might like such a data frame (it'd be useful for making feature matrices for models).

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970310857497) by [Unito](https://www.unito.io/learn-more)

ryantibs:
Hi all, particularly capnrefsmmat @chinandrew, this finishes off the `covidcast_signals()` function, and provides related signal aggregation functionality.

* Closes #10 and #48.
* Discussion of the design decisions for signal aggregation can be found in the comments on #10. 
* I also did a bunch of clean-up work, like strike direction from the record everywhere (documentation, `covidcast_signal` object, plotting functionality, etc.), which I suggested we should do in #131.

Because I branched this off of evalcast-review (which was branched off of evalcast), the commit log here is huge.  My first "real" commit for this branch starts at b42b353.  Once evalcast gets merged into main #85 (should happen within a couple days), the commit log here will look a lot cleaner, and we can merge this branch.  chinandrew has already provided an informal (?) review, maybe he can take one more look.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970430348989) by [Unito](https://www.unito.io/learn-more)


chinandrew:
Changed the based of this PR to evalcast so 1) if it gets merged it doesn't accidently merge in all of evalcast to main, and 2) makes the diff and commit log a bit easier to read. It's easy to switch back if needed
Screenshot from 2020-10-28 11-23-29

R-packages/covidcast/R/wrangle.R
```diff
@@ -0,0 +1,295 @@
...
+#'   `covidcast_signal` data frames, and `dt` is either a vector of shifts---to
+#'   apply the same shifts for each signal in `x`, or a list of vector of
+#'   shifts---to apply different shifts for each signal in `x`.
+#'
+#' @return Data frame of aggregated signals in "wide" or "long" form, depending 
+#'   on `format`. In "long" form, an extra column `dt` is appended to indicate
+#'   the value of the time-shift. In "wide" form, only the latest issue of data
+#'   is retained, and the columns `data_source`, `signal`, `issue`, `lag`,
+#'   `stderr`, `sample_size` are all dropped from the returned data frame. Each
+#'   unique signal---defined by a combination of data source name, signal name,
```
chinandrew, line 128: Might be worth noting that these are full joins on everything in the list and not some other join type

chinandrew:
As mentioned in #10 the interface here looks good to me, though someone more familiar with R paradigms should have the final say.

R-packages/covidcast/R/wrangle.R
ryantibs: Thanks, done.

ryantibs:
@chinandrew @capnrefsmmat Few more updates.  First, I filled out the vignette so this branch should be good to go.  Andrew approves but suggests somebody more familiar with R takes a look.  To repeat this doesn't need to be merged until evalcast gets merged, which shouldn't be for a few days at least.  If Alex you're too busy then maybe @JedGrabman can take a look.

Second, a spotted a discrepancy in the way I defined `dt` in `aggregate_signals()` and `covidcast_cor()`.  The definition in the former function is more natural; the latter has it flipped.  So I changed everything surrounding `covidcast_cor()` to match `aggregate_signals()` and looked around for code that would break; all I could find was the correlations vignette and the intro-api talk, both of which I fixed.

To summarize: 
- This should be merged into main after evalcast gets merged.
- This requires us to rebuild the documentation and the vignettes.
- Maybe we should record as an update that `covidcast_cor()` now interprets `dt_x` and `dt_y` in the *opposite* way it did before, so that negative values mean lagging and positive values mean leading (which is more natural).

capnrefsmmat:
> Maybe we should record as an update that covidcast_cor() now interprets dt_x and dt_y in the opposite way it did before, so that negative values mean lagging and positive values mean leading (which is more natural).

That can go into NEWS.md, and we can bump the minor version of the package accordingly. Note that this branch doesn't contain #148, so there'll be a conflict if you do this.

ryantibs:
@capnrefsmmat @chinandrew Upon reflection there was really no reason for me to have branched this off of the evalcast branch.  I was just working on something else that I needed the evalcast package for, and then I started implementing these multi signal functions because I also needed them.  But my "real" commits starting with b42b353 have noting to do with evalcast.  

Since the evalcast branch is actually held up with more development needed, what's the best way to jut pick off the commits starting with b42b353 through the end, and move them to a separate branch, which is created off of main?  Then we could just merge that new into main.

capnrefsmmat:
We'd have to carefully cherry-pick, since there *are* changes to evalcast in this series of commits, namely in 
1e674d8. Those already seem to have a merge conflict, so we'd want to cherry-pick every other commit into a new branch.

That's doable, although whether or not we do it, I think it still makes sense to merge evalcast and continue further development of it in new PRs targeting specific features/fixes.

chinandrew:
You should either be able to do a rebase like `git rebase --onto main evalcast multi-signals`, or creating a new branch and cherry-picking the commit range you want: `git cherry-pick start_commit^..end_commit`

ryantibs:
Thanks both of you! I guess I could try to cherry-pick but it might be easiest just to merge evalcast to main as it stands now (Alex I just left a message agreeing with your logic on #85).

ryantibs:
@capnrefsmmat Well I snuck in a few other really small light housekeeping things.  See the details of my last commit.  I managed to do it before you merge this into main (but apparently not before you merged main into this!).

R-packages/covidcast/R/cor.R
```diff
@@ -1,13 +1,18 @@
...
-#'   before computing correlations. These must both be nonnegative, and only one
+#'   before computing correlations. Negative shifts translate into in a lag
+#'   value and positive shifts into a lead value; for example, if `dt = -1`,
+#'   then the value on June 2 that gets reported is the original value on June
+#'   1; if `dt = 0`, then the values are left as is.
+#'
+#' These must both be nonnegative, and only one
 #'   can be positive. If `dt_x = 1`, for example, then data for `x` is shifted
 #'   forward 1 day in time (so, data on June 1 becomes data on June 2, and so
 #'   on). Default is 0 for both.
```
capnrefsmmat, line 19: Is this documentation accurate anymore? It still says both must be nonnegative.

R-packages/covidcast/R/wrangle.R
```diff
@@ -0,0 +1,296 @@
...
+    dplyr::relocate(dt, .before = value)
+
+  # Change class and return
+  class(x) = c("covicast_signal_long", "data.frame")
+  return(x)
+}
+
+#' Pivot covidcast_signal_long object to "wide" format
+#
+#' @export
```
capnrefsmmat, line 270: Is this sufficient documentation for this function (and its counterpart `covidcast_longer`)?

capnrefsmmat (changes requested):
Noticed two issues while preparing to merge

R-packages/covidcast/R/cor.R
ryantibs: Oops that was a paragraph that was intended to be deleted. Good catch, thanks.

R-packages/covidcast/R/wrangle.R
ryantibs: Thanks, I added a bit more.  This doesn't need extensive documentation because it just acts as if `aggregate_signals()` had been called with a different option, which I now explained.

Issue #158, opened 2020-10-29
Create a notebook that studies mask wearing by case velocity


tildechris:
There have been regional changes in mask wearing: 

![mask-wearing-changes](https://user-images.githubusercontent.com/46794497/97576105-1ade6900-19c4-11eb-8b40-0b12fcda7591.png)

Note that the largest increase in mask wearing, is where cases are currently spiking (upper great plains states).  Where mask wearing has gone down, however, there are more modest changes in cases:

![case_velocity_by_mask_change](https://user-images.githubusercontent.com/46794497/97576122-216ce080-19c4-11eb-811c-2da4a3237994.png)

Create a notebook that studies this correlation.

PR #159, opened 2020-10-29
Study mask wearing by case rates

tildechris:
Closes: #158


tildechris:
The regional correlation is striking:

image

Issue #160, opened 2020-10-29
Minor improvements for evalcast trajectory plots


jsharpna:
- [x] add pagination for when number of locations exceed threshold
- [x] tune the size of trajectory plots
- [ ] tune the font size

(If you have access) you can see what I did for the [QA plots here](https://github.com/cmu-delphi/covidcast-forecast/blob/main/forecaster_predictions/submit_predictions/code/trajectories.R), basically I just used `facet_wrap_paginate`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970294899032) by [Unito](https://www.unito.io/learn-more)

jsharpna:
want to assign to @yelselmiao

dajmcdon:
Closed with #213 and #247

PR #213, opened 2020-10-30
Update plot_trajectory.R

yelselmiao:
Minor improvement on pagination regarding issue #160 

* I replaced `facet_wrap` with `facet_wrap_paginate` in the **plot_trajectory.R** so that it can accommodate plots with more than 200 locations by breaking the facets into pages.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198975694234112) by [Unito](https://www.unito.io/learn-more)


yelselmiao:
@dajmcdon

R-packages/evalcast/R/plot_trajectory.R
```diff
@@ -28,7 +28,8 @@
 plot_trajectory <- function(list_of_predictions_cards,
                             first_day = "2020-07-01",
                             last_day = Sys.Date(),
-                            alpha = .2)
+                            alpha = .2,
+                           page_facet = 1)
```
dajmcdon, line 32: Let's also pass some extra arguments here `ncol=4` and `nrow=6` or some reasonable defaults to be used below.

R-packages/evalcast/R/plot_trajectory.R
```diff
@@ -97,7 +98,22 @@ plot_trajectory <- function(list_of_predictions_cards,
   response_df <- response_df %>% dplyr::mutate(
     location_abbr = if(geo_type == "state") covidcast::fips_to_abbr(paste0(location,"000")) else covidcast::fips_to_name(location)) %>%
     dplyr::filter(!is.na(location_abbr))
-  ggplot(preds_df, aes(x = .data$reference_period)) +
+    
+    #set the page number 
+  fc_round <- function(x){  #round to integer
+    idx = rep(c(FALSE,TRUE), length.out=length(x))
+    floor(x)*idx + ceiling(x)*(!idx)
+  }
```
dajmcdon, line 106: We don't actually need this at all I don't think.

R-packages/evalcast/R/plot_trajectory.R
```diff
@@ -97,7 +98,22 @@ plot_trajectory <- function(list_of_predictions_cards,
...
+  
+  if (length(unique(preds_df$location_abbr))/45 < 1) {
+    page_facet = page_facet
+  } else 
+    page_facet = fc_round(length(unique(preds_df$location_abbr))/45)
+
```
dajmcdon, line 112: Here, we do
```
nfacets = length(unique(preds_df$location_abbr))
if(ncol*nrow*page_facet < nfacets) page_facet = ceiling(nfacets / (ncol*nrow))
```
We have to figure out how to let one or both of ncol or nrow be `NULL` so we can just let `facet_wrap()` do it's thing in those cases.

dajmcdon (changes requested):

R-packages/evalcast/R/plot_trajectory.R
```diff
@@ -100,40 +102,46 @@ plot_trajectory <- function(list_of_predictions_cards,
...
+            labs(x = incidence_period,
+                 y = paste0(response$data_source,": ",response$signal),
+                 colour = "forecaster_name") +
+           theme_bw() + #tune font size for axis and facet title
+           theme(axis.text = element_text(size = 8), 
+                 strip.text = element_text(size = 10,face = "bold"))
+ 
+ # The breakpoint is 52 so that all states can be rendered in one plot
+ if (nfacets <= 52 ) {
+   print(p + ggplot2::facet_wrap(~.data$location_abbr, scales = "free", ncol = 6,nrow=NULL))
```
dajmcdon, line 137: I'd rather not set this explicitly. I don't want to override the user's choices without reason. So let's remove this check.

R-packages/evalcast/R/plot_trajectory.R
```diff
@@ -100,40 +102,46 @@ plot_trajectory <- function(list_of_predictions_cards,
...
+ } else { 
+   for (i in 1: page_facet) { 
+# if facet > 52, pagination will be implemented
+     print(p + ggforce::facet_wrap_paginate(~.data$location_abbr, scales = "free_y", ncol = ncol,nrow=nrow, page=i))
+   }
```
dajmcdon, line 142: What happens if we want to save these plots with `ggsave()`? Or if we don't? I imagine this would work inside an Rmarkdown document, but what about outside of one?

dajmcdon (commented):
See comments in the code.

yelselmiao:
Commit [f9518a3](https://github.com/cmu-delphi/covidcast/pull/213/commits/f9518a3136f76bbfb3b82a1716962811204dbe4e)
Minor change mades:

- **Line 33**: default `ncol=6` is changed to `ncol=4` for better appearance. 
- **Line 35**: new argument `cutoff= 52` added.
- **Line 137**: explicit number 52 is replaced with `cutoff`

R-packages/evalcast/R/plot_trajectory.R
```diff
@@ -28,7 +28,11 @@
 plot_trajectory <- function(list_of_predictions_cards,
                             first_day = "2020-07-01",
                             last_day = Sys.Date(),
-                            alpha = .2)
+                            alpha = .2,
+                            nrow = 6,
+                            ncol = 4,  
+                           page_facet = 1，
+                           cutoff = 52)
```
dajmcdon, line 35: Ok, last thing I think.

Go through and fix the indenting where you see it.

We want to clean this up so it is easily readable.

R-packages/evalcast/R/plot_trajectory.R
```diff
@@ -97,25 +101,49 @@ plot_trajectory <- function(list_of_predictions_cards,
...
+  nfacets = length(unique(preds_df$location_abbr)) 
+  if(ncol*nrow*page_facet < nfacets) page_facet = ceiling(nfacets / (ncol*nrow))
+  
+
+  
+#facetting
+  p <- ggplot(preds_df, aes(x = .data$reference_period)) +
+            geom_line(aes(y = .data$point,
+                          col = .data$forecaster_name, 
+                          group=.data$forecast_date),
```
dajmcdon, line 115: More indenting problems

R-packages/evalcast/R/plot_trajectory.R
```diff
@@ -97,25 +101,49 @@ plot_trajectory <- function(list_of_predictions_cards,
   response_df <- response_df %>% dplyr::mutate(
     location_abbr = if(geo_type == "state") covidcast::fips_to_abbr(paste0(location,"000")) else covidcast::fips_to_name(location)) %>%
     dplyr::filter(!is.na(location_abbr))
```
dajmcdon, line 103: This is also pretty ugly. It runs over a lot and is hard to read. Can we clean this up?

dajmcdon (changes requested):
Ok, last try. A few more minor things to fix.

R-packages/evalcast/R/plot_trajectory.R
yelselmiao: Okay, I'll look over the script and fix the indenting issue:)

yelselmiao:
More fixes to improve readability
- Loops are broken into lines
- Arguments are aligned

dajmcdon (approved):
I'm going to make a few more fixes. It's probably worth checking them out, but I'd like to have you move on to the next thing.

PR #216, opened 2020-10-30
Add HRR and MSA to choropleth plots 

sarah-colq:
Shapefiles from python package are used and are now under data.  Description has been updated to include sf library.  Alaska, PR and Hawaii are manually moved.  There is a background grey U.S. for the MSA plot (but also used for HRR).  This solves half of issue #216


R-packages/covidcast/R/plot.R
```diff
@@ -8,8 +8,11 @@ plot_choro = function(x, time_value = NULL, include = c(), range,
                       title = NULL, params = list()) {
   # Check that we're looking at either counties or states
   if (!(attributes(x)$geo_type == "county" ||
-                     attributes(x)$geo_type == "state")) {
-    stop("Only 'county' and 'state' are supported for choropleth maps.")
+                     attributes(x)$geo_type == "state" ||
+                     attributes(x)$geo_type == "hrr" ||
+                     attributes(x)$geo_type == "msa")) {
```
capnrefsmmat, line 13: At this point we may as well do `attributes(x)$geo_type %in% c("county", "state", "hrr", "msa")`, since there are so many

R-packages/covidcast/R/plot.R
```diff
@@ -104,6 +107,80 @@ plot_choro = function(x, time_value = NULL, include = c(), range,
   geo = df$geo
   names(val) = geo
 
+  # Make background layer for MSA and HRR maps which are incomplete
+  if ((attributes(x)$geo_type == "msa") |
```
capnrefsmmat, line 111:
```suggestion
  if ((attributes(x)$geo_type == "msa") ||
```

just a nitpick; `|` is vector elementwise OR, while `||` is logical OR with short-circuiting behavior

R-packages/covidcast/R/plot.R
```diff
@@ -104,6 +107,80 @@ plot_choro = function(x, time_value = NULL, include = c(), range,
...
+
+    alaska_centroid = sf::st_centroid(sf::st_geometry(alaska_df))
+    print('hi')
+    alaska_scale = (sf::st_geometry(alaska_df) - alaska_centroid) * 0.35 + alaska_centroid
+    alaska_df = sf::st_set_geometry(alaska_df, alaska_scale)
+    alaska_shift = sf::st_geometry(alaska_df) + c(-2e+6, -2.6e+6)
+    alaska_df = sf::st_set_geometry(alaska_df, alaska_shift)
+    alaska_df = sf::st_set_crs(alaska_df, 102003)
+
+    print('hi')
```
capnrefsmmat, line 148: whoops

R-packages/covidcast/R/plot.R
```diff
@@ -104,6 +107,80 @@ plot_choro = function(x, time_value = NULL, include = c(), range,
...
+    geom_args$fill = pr_col
+    geom_args$data = pr_df
+    back_pr_layer = do.call(ggplot2::geom_sf, geom_args)
+    geom_args$fill = hawaii_col
+    geom_args$data = hawaii_df
+    back_hawaii_layer = do.call(ggplot2::geom_sf, geom_args)
+    geom_args$fill = alaska_col
+    geom_args$data = alaska_df
+    back_alaska_layer = do.call(ggplot2::geom_sf, geom_args)
+    }
```
capnrefsmmat, line 182: incorrect indentation on this `}`

R-packages/covidcast/R/plot.R
```diff
@@ -104,6 +107,80 @@ plot_choro = function(x, time_value = NULL, include = c(), range,
...
+    hawaii_df = sf::st_transform(hawaii_df, 102007)
+    alaska_df = sf::st_transform(alaska_df, 102006)
+    pr_df = sf::st_transform(pr_df, 102003)
+
+    hawaii_shift = sf::st_geometry(hawaii_df) + c(-1e+6, -2e+6)
+    hawaii_df = sf::st_set_geometry(hawaii_df, hawaii_shift)
+    hawaii_df = sf::st_set_crs(hawaii_df, 102003)
+
+    alaska_centroid = sf::st_centroid(sf::st_geometry(alaska_df))
+    print('hi')
```
capnrefsmmat, line 141: whoops

R-packages/covidcast/R/plot.R
```diff
@@ -104,6 +107,80 @@ plot_choro = function(x, time_value = NULL, include = c(), range,
   geo = df$geo
   names(val) = geo
 
+  # Make background layer for MSA and HRR maps which are incomplete
+  if ((attributes(x)$geo_type == "msa") |
+      (attributes(x)$geo_type == "hrr")) {
+    map_df = sf::st_read('../data/shapefiles/state/cb_2019_us_state_5m.shp')
```
capnrefsmmat, line 113: Hmm. I think this (and the similar call below) relies on the R working directory being exactly right. This may not work when the user installs the package themselves. Should we be putting the shapefiles in `inst/` and using `system.file(path, package="covidcast")` or something like that?

capnrefsmmat (changes requested):
Looking like a good start. AA few nitpicking comments; the only substantive one is ensuring we put the shapefiles somewhere that gets installed with the package, since I think the current way of loading `data/` won't work when the package is installed.

ryantibs:
@sarah-colq @capnrefsmmat Make sure not to rely on `attributes(x)$geo_type` but instead use  `attributes(x)$metadata$geo_type`, since #156 eliminates the former!

capnrefsmmat:
Assigning @sgsmob for review (assuming this is ready for re-review), so I stop trying to do everything myself

R-packages/covidcast/R/plot.R
```diff
@@ -252,6 +253,80 @@ plot_choro = function(x, time_value = NULL, include = c(), range,
...
+  # Make background layer for MSA and HRR maps which are incomplete
+  if ((attributes(x)$metadata$geo_type == "msa") ||
+      (attributes(x)$metadata$geo_type == "hrr")) {
+    map_df = sf::st_read(system.file(
+                         "shapefiles/state",
+                         "cb_2019_us_state_5m.shp",
+                         package = "covidcast"))
+    background_crs = sf::st_crs(map_df)
+    map_df$STATEFP <- as.character(map_df$STATEFP)
+    map_df$is_alaska = map_df$STATEFP == '02'
```
sgsmob, line 265: I think it is possible and cleaner to wrap many of these column creations into one call to `dplyr::mutate()`.

R-packages/covidcast/R/plot.R
```diff
@@ -252,6 +253,80 @@ plot_choro = function(x, time_value = NULL, include = c(), range,
...
+    map_df$is_pr = map_df$STATEFP == '72'
+    map_df$STATEFP <- as.numeric(map_df$STATEFP)
+    map_df$is_state = map_df$STATEFP < 57
+    map_df$color = missing_col
+
+    pr_df = map_df %>% filter(.$is_pr)
+    hawaii_df = map_df %>% filter(.$is_hawaii)
+    alaska_df = map_df %>% filter(.$is_alaska)
+
+    main_df = map_df %>% filter(!map_df$is_alaska)
```
sgsmob, line 276: These 3 lines and probably every other place where you have the same left hand side on 2+ consecutive lines probably can be rewritten using the pipe operator (`%>%`).

R-packages/covidcast/R/plot.R
```diff
@@ -286,15 +361,156 @@ plot_choro = function(x, time_value = NULL, include = c(), range,
...
-  # Create the polygon layer
-  aes = ggplot2::aes
-  geom_args = list()
-  geom_args$color = border_col
-  geom_args$size = border_size
-  geom_args$fill = map_col
-  geom_args$mapping = aes(x = x, y = y, group = group)
-  geom_args$data = map_df
-  polygon_layer = do.call(ggplot2::geom_polygon, geom_args)
+  else if (attributes(x)$metadata$geo_type == "msa") {
```
sgsmob, line 364: Basically all my comments from above apply here too:

- Make into a separate function
- Use `mutate`
- Chain calls with pipes where possible

sgsmob (changes requested):

krivard:
It looks like the build is failing -- @chinandrew any ideas?

sarah-colq:
Oh I didn't mean to request that review, I pulled the cmu-delphi directory into my own fork to update everything before doing those changes that were requested.  I will investigate the build failure as well.

chinandrew:
Hmm I've gotten two errors, one on the dependency install and now one on the R setup. looking into it
```
Error in data.frame(package = deps, installed = inst_ver, available = cran_ver,  : 
  row names contain missing values
Calls: <Anonymous> ... dev_package_deps -> package_deps -> structure -> data.frame
Execution halted
Error: Process completed with exit code 1.
```

krivard:
@sarah-colq no worries; I get auto-tagged for "miscellaneous" changes, like the update to `r_ci.yml` (which really shouldn't have broken anything, so, glad i was here)

chinandrew:
Fix for the CI is in #290 , some combination of the r version that was being built and resetting the cache.

krivard:
@sarah-colq if you merge main into this, the CI should start passing

R-packages/covidcast/R/plot.R
```diff
@@ -252,6 +253,80 @@ plot_choro = function(x, time_value = NULL, include = c(), range,
...
 
+  # Make background layer for MSA and HRR maps which are incomplete
+  if ((attributes(x)$metadata$geo_type == "msa") ||
+      (attributes(x)$metadata$geo_type == "hrr")) {
+    map_df = sf::st_read(system.file(
+                         "shapefiles/state",
+                         "cb_2019_us_state_5m.shp",
+                         package = "covidcast"))
+    background_crs = sf::st_crs(map_df)
+    map_df$STATEFP <- as.character(map_df$STATEFP)
```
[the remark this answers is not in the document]
sarah-colq: Yeah, it wasn't working without it.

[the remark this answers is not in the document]
sarah-colq: Yes, it wasn't working without it.

R-packages/covidcast/R/plot.R
```diff
@@ -252,6 +253,80 @@ plot_choro = function(x, time_value = NULL, include = c(), range,
   geo = df$geo
   names(val) = geo
 
+  # Make background layer for MSA and HRR maps which are incomplete
+  if ((attributes(x)$metadata$geo_type == "msa") ||
```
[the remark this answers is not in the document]
sarah-colq: I've put repeated sections with shifting and scaling into helper functions.  I have not done it further because I will be replacing the state and county sections in my next PR, which will eliminate a lot of repetition because they will be doing the same thing.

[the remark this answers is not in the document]
sarah-colq: @sgsmob let me know if the changes work.

sarah-colq:
@chinandrew The build is still failing ( I did merge in main before pushing my changes to this PR).  Any ideas?

chinandrew:
Hm that's weird...maybe something to do with it being a fork? 

Try incrementing  line 39 in .github/workflows/r_ci.yaml  from
`          key: ${{ runner.os }}-r-2-
`
to 
`          key: ${{ runner.os }}-r-3-
`
to reset the dependency cache

sarah-colq:
@chinandrew Still not working. :(

R-packages/covidcast/DESCRIPTION
```diff
@@ -46,8 +46,10 @@ Imports:
   maptools,
   purrr,
   rlang,
+  sf,
   tidyr,
   usmap
+
```
chinandrew, line 52: @sarah-colq I'd be surprised if this was the issue, but is it this newline?

Seems like the error is may be due to[ improperly specified packages](https://github.com/r-lib/devtools/issues/1262#issuecomment-232352601), and I wonder if it's looking at the newline.

.github/workflows/r_ci.yml
```diff
@@ -30,13 +30,16 @@ jobs:
...
+        run: |
+            sudo apt-get install libcurl4-openssl-dev
+            sudo apt-get install libudunits2-dev
+            sudo apt-get install libgdal-dev
       - name: Cache R packages
         uses: actions/cache@v2
         with:
           path: ${{ env.R_LIBS_USER }}
-          key: ${{ runner.os }}-r-2-
+          key: ${{ runner.os }}-r-3-
```
sgsmob, line 42: @chinandrew does this need to be reverted?

R-packages/covidcast/DESCRIPTION
sarah-colq: Looks like that fixed it!

R-packages/covidcast/R/plot.R
```diff
@@ -566,3 +710,63 @@ plot_line = function(x, range = NULL, title = NULL, params = list()) {
...
+
+
+shift_main = function(map_df){
+  main_df = map_df %>% dplyr::filter(
+      !.$is_alaska) %>% dplyr::filter(
+        !.$is_hawaii) %>% dplyr::filter(!.$is_pr)
+  main_df = sf::st_transform(main_df, 102003)
+  return(main_df)
+}
+
```
sgsmob, line 772: It's asking for a newline here, but perhaps you need to delete the whitespace at the end of this file

R-packages/covidcast/inst/shapefiles/county/cb_2019_us_county_5m.cpg
```diff
@@ -0,0 +1 @@
+UTF-8
```
sgsmob, line 1: Is this an extraneous output file?  If so we should probably mask it in a `.gitignore`

R-packages/covidcast/inst/shapefiles/county/cb_2019_us_county_5m.prj
```diff
@@ -0,0 +1 @@
+GEOGCS["GCS_North_American_1983",DATUM["D_North_American_1983",SPHEROID["GRS_1980",6378137,298.257222101]],PRIMEM["Greenwich",0],UNIT["Degree",0.017453292519943295]]
```
sgsmob, line 1: Same comment as above

sgsmob (commented):
Nearly ready to approve, but still have some questions about stray files.

.github/workflows/r_ci.yml
chinandrew: Don't think it matters (both caches are now created), but I reverting it makes sense since it's not needed right now.

R-packages/covidcast/inst/shapefiles/county/cb_2019_us_county_5m.cpg
capnrefsmmat: Shapefiles are a bizarre collection of several files in one folder; I think both these files are probably necessary. (For example, the .prj defines the geographic projection used for the shapefile's coordinates)

sarah-colq: No, All these files would need to be used if we call the county shp file.  This hasn't happened yet, but I'm planning on using those to add Puerto Rico in a subsequent PR.

R-packages/covidcast/inst/shapefiles/county/cb_2019_us_county_5m.prj
sarah-colq: Same as above.

.github/workflows/r_ci.yml
sarah-colq: I changed it back and it still works.

sgsmob (approved):

sarah-colq:
@krivard @capnrefsmmat Could you do a last check and give approval so this can be merged?

capnrefsmmat (approved):
I think you only need me, since I requested changes before. Looks good now. We should definitely follow up to add tests using the vdiffr stuff I set up in #217; I'll file an issue

Issue #134, opened 2020-10-21
Choropleth map plot does not work if all the state values are the same (R library)


benjaminysmith:
When trying to plot data that has the same value for every state I get: "Error: Breaks and labels are different lengths". Changing one value slightly allows the plot. I suspect this is because of the way that the legend breaks vs labels are calculated.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970434114006) by [Unito](https://www.unito.io/learn-more)

ryantibs:
@benjaminysmith Can you produce example code?  
@JedGrabman maybe you want to take a quick look?

benjaminysmith:
Example:

```
library(covidcast)

example_df = data.frame("geo_value"= tolower(c(state.abb)))

example_df$time_value = "2020-04-15"
example_df$issue = "2020-04-15"
example_df$value = 10
attributes(example_df)$geo_type = "state"
class(example_df) = c("covidcast_signal", "data.frame")

plot(example_df)
```

Fails with:

`Error: Breaks and labels are different lengths`

Adding `range=c(0,10)` to plot command is workaround.

JedGrabman:
```
In addition: Warning message:
Metadata for signal mean and standard deviation not available; defaulting to observed mean and standard deviation to set plot range.
```

Issue looks like it probably starts [here](https://github.com/cmu-delphi/covidcast/blob/5ca16a9e111b82f23c399549a8e7d5f4c910bcfd/R-packages/covidcast/R/covidcast.R#L414)


`stdev_value <- sd(x$value)`
...
`range <- c(mean_value - 3 * stdev_value, mean_value + 3 * stdev_value)`

If we just check if stdev is 0 or set it to atleast some small positive value, it should take care of it.

JedGrabman:
Adding to myself for now so it doesn't fall through the cracks, but treating as low-priority

statsmaths:
The older counterexample doesn't run for a new reason now (I think due to some naming changes in the package). Here's a way to reproduce on r-pkg-devel:

```{r}
library(covidcast)
library(tibble)

example_df <- structure(
  tibble(
    geo_value = tolower(state.abb),
    time_value = "2020-04-15",
    issue = "2020-04-15",
    data_source = "example",
    signal = "example",
    value = 10
  ),
  metadata = list(geo_type = "state"),
  class = c("covidcast_signal", "data.frame")
)

plot(example_df)
```

[...]

PR #217, opened 2020-10-31
Try some basic unit tests

capnrefsmmat:
WIP. Will write some more tests Soonish to see what happens (and how well our CI works).


capnrefsmmat:
Apparently there's a nice package for testing plotting, so we can also do that to avoid regressing our plotting features: https://github.com/r-lib/vdiffr

capnrefsmmat:
Ok, I've successfully set up mockery and vdiffr to do tests of functions that use the network (like `covidcast_signal`) and tests of our plotting/mapping functions. They work great on simple examples.

I also added `DEVELOP.md` with instructions on what to run and a PR checklist.

I think it'll be easy to make a few different plot examples (by saving a few real data frames) and save those as the key plotting tests. We could also add tests for issues like #134 when they're fixed.

capnrefsmmat:
There are a lot more tests we should write, but I likely won't have time to write more for a while. So it's probably best if we merge this now, and then file an issue with a list of things we should add tests for. Then the testing framework will be in place for everyone to use for their new PRs, and we can fill in the testing gaps over the next few weeks.

ryantibs:
This looks great, thanks for starting this off!. I won't be able to review this in any level of detail in the next few days, too bogged down with other things. Are any of @nmdefries @JedGrabman @sarah-colq @sgsmob @undefx @sangwon-hyun @nloliveira @capolitsch @brookslogan available?

sgsmob:
I can't assign myself to review this but I will

R-packages/covidcast/DESCRIPTION
```diff
@@ -52,5 +52,9 @@ RoxygenNote: 7.1.1
 Suggests: 
     knitr,
     rmarkdown,
-    gridExtra
+    gridExtra,
+    testthat,
```
sgsmob, line 56: Please alphabetize these lists

R-packages/covidcast/tests/testthat/test-cor.R
```diff
@@ -0,0 +1,85 @@
...
+  expect_equal(covidcast_cor(foo, bar, by = "geo_value"),
+               expected)
+
+  # Try permuting order
+  expect_equal(covidcast_cor(foo, bar[c(2, 1, 4, 3), ],
+                             by = "geo_value"),
+               expected)
+})
+
+test_that("basic lagged correlations", {
```
sgsmob, line 33: Please add a comment explaining the purpose of this test.

R-packages/covidcast/tests/testthat/test-cor.R
```diff
@@ -0,0 +1,85 @@
...
+    geo_value = "pa",
+    value = 1
+  )
+  expect_equal(covidcast_cor(foo, bar, dt_x = 1),
+               expected)
+  expect_equal(covidcast_cor(foo, bar, dt_y = 1),
+               expected)
+})
+
+test_that("lags are in correct direction", {
```
sgsmob, line 60: Please add a comment explaining the purpose of this test.

R-packages/covidcast/tests/testthat/test-covidcast.R
```diff
@@ -0,0 +1,26 @@
+library(covidcast)
+library(mockery)
+
+test_that("covidcast_meta", {
```
sgsmob, line 4: Please add a comment explaining this test

R-packages/covidcast/tests/testthat/test-covidcast.R
```diff
@@ -0,0 +1,26 @@
...
+                   max_time = as.Date(c("2020-01-02", "2020-10-03")),
+                   max_issue = as.Date(c("2020-04-04", "2020-11-01"))
+                 ),
+                 class = c("covidcast_meta", "data.frame")
+               ))
+
+  stub(covidcast_meta, ".request",
+       list(message = "argle-bargle"))
+
+  expect_error(covidcast_meta())
```
sgsmob, line 25: Does this allow specification of which error is expected?

R-packages/covidcast/tests/testthat/test-plot.R
```diff
@@ -0,0 +1,81 @@
+library(testthat)
+library(vdiffr)
+library(dplyr)
+
+# Contexts are no longer required or recommended in testthat, but vdiffr still
+# wants one to place the figure files correctly. See
+# https://github.com/r-lib/vdiffr/issues/71
+context("plot")
+
+test_that("simple line graph", {
```
sgsmob, line 10: Please leave a comment describing the purpose of this test

R-packages/covidcast/tests/testthat/test-plot.R
```diff
@@ -0,0 +1,81 @@
...
+    line_params = list(
+      xlab = "Day",
+      ylab = "Penguinocity",
+      stderr_bands = TRUE,
+      stderr_alpha = 0.3
+    )
+  ))
+})
+
+test_that("state line graphs", {
```
sgsmob, line 35: Please leave a comment describing the purpose of this test

R-packages/covidcast/tests/testthat/test-plot.R
```diff
@@ -0,0 +1,81 @@
...
+                      plot(filter(fb_state, geo_value %in% c("pa", "tx", "ny")),
+                           plot_type = "line",
+                           line_params = list(stderr_bands = TRUE)))
+
+  expect_doppelganger("state line graph with range",
+                      plot(fb_state, plot_type = "line",
+                           range = c(0, 10)))
+})
+
+test_that("simple state choropleths", {
```
sgsmob, line 51: Please leave a comment describing the purpose of this test

R-packages/covidcast/tests/testthat/test-wrangle.R
```diff
@@ -0,0 +1,82 @@
+library(covidcast)
+library(dplyr)
+
+test_that("aggregate and shift", {
```
sgsmob, line 4: Please leave comments describing the purpose of the tests in this file

R-packages/covidcast/tests/testthat/test-wrangle.R
```diff
@@ -0,0 +1,82 @@
...
+               structure(data.frame(
+                 data_source = c(rep("bar", 3), rep("foo", 3)),
+                 signal = c(rep("bar", 3), rep("foo", 3)),
+                 geo_value = rep(c("pa", "ri", "tx"), 2),
+                 time_value = as.Date("2020-01-01"),
+                 dt = 0,
+                 value = c(4, 6, 5, 1, 3, 2)),
+                 class = c("covidcast_signal_long", "data.frame")))
+
+  # Now try it long in the first place. Currently fails because wide format does
```
sgsmob, line 78: If these tests are not working, they should just be removed from this file entirely.

sgsmob (changes requested):
This will be a great asset to this repo!

Not entirely sure how to validate an initial commit of the charts but I think they are probably ok.  Will approve once some more comments are added.

R-packages/covidcast/tests/testthat/test-covidcast.R
capnrefsmmat: Not if `covidcast_meta()` uses `stop()`, which does not specify a specific error type -- unless you regexp-match against the error message.

I've just changed `covidcast_meta()` to use `abort()` with a specific error class, so we can catch and handle it, and the test can verify it was raised correctly.

capnrefsmmat:
Pushed fixes. Rather than adding a bunch of comments to the tests, I tried to change their names so that when you read them, they're meaningful, such as

```r
test_that("covidcast_meta raises error when API signals one", {
```

capnrefsmmat:
Also, regarding the commented-out test code in `test-wrangle.R`: I can use `skip()` to skip that instead of commenting out, but that also prints a message every time you run the tests. The issue is that I want to change the code to make that test pass, but it doesn't yet; would you prefer the code be removed entirely and returned when the issues is fixed, leave it commented out, or have it skipped and noisily complain to encourage people to fix the wrangling code?

sgsmob:
> Also, regarding the commented-out test code in `test-wrangle.R`: I can use `skip()` to skip that instead of commenting out, but that also prints a message every time you run the tests. The issue is that I want to change the code to make that test pass, but it doesn't yet; would you prefer the code be removed entirely and returned when the issues is fixed, leave it commented out, or have it skipped and noisily complain to encourage people to fix the wrangling code?

A TODO in the comments at the top would be good.

sgsmob (approved):

PR #232, opened 2020-11-01
Evalcast tests

jacobbien:
Adding more tests to the evalcast package (and fixing a bug involving geo_values).  The new tests:
- test that backfill buffer works
- test that additional arguments to forecaster works
- test that geo_values works
- test that start_day within signals works

I've verified that all tests pass and R CMD check runs with no errors, warnings, or notes.

### DIFF
--- a/R-packages/evalcast/R/get_predictions.R
+++ b/R-packages/evalcast/R/get_predictions.R
@@ -88,7 +88,7 @@ get_predictions_single_date <- function(forecaster,
                                         geo_type,
                                         geo_values,
                                         ...) {
-  if (length(geo_values) > 1) geo_values <- list(geo_values)
+  #if (length(geo_values) > 1) geo_values <- list(geo_values)
   forecast_date <- lubridate::ymd(forecast_date)
   # compute the start_day from the forecast_date, if we need to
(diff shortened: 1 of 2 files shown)


jacobbien:
Two things to consider:
- I use `<<-` within the first test_that so that these objects can be used in other tests.  Not sure if that's bad practice.
- all tests are currently in a single file; perhaps we should consider splitting these up

bnaras:
@jacobbien The use of `<<-`, which should be limited, can almost always be avoided by using an environment variable, the one object in R that has referential semantics. So, I'd create a variable 
```{r}
common_objects <- new.env(parent = emptyenv())
```
and from then on, stick everything you need in it via
```{r}
common_objects$foo <- foo_object_to_save
```
In fact, this is also a good general strategy for global variables: create an environment and use that throughout. Such global variables may also be used for configuration variables in a package that the user may change.

jacobbien:
Thanks @bnaras. I followed your advice in the latest commit.  (Unfortunately my commit message, which was "Remove `<<-` from tests." was changed by git to "Remove from tests.")

dajmcdon (approved):
This looks good to me. I'll approve and check with @bnaras before merging.

jacobbien:
Thanks!

PR #233, opened 2020-11-01
Small R package fixes and improved documentataion

capnrefsmmat:
This fixes a check warning and improves the discoverability of the documentation.

@ryantibs If this looks good to you, I'll just regenerate the HTML and merge.


capnrefsmmat:
If you don't want to wait for the pkgdown HTML to build, here's what the top part of the reference page looks like now:

Screen Shot 2020-11-01 at 2 31 14 PM

capnrefsmmat:
54bd8a5 is a good example of why we need to finish up #217 and add a bunch of tests...

R-packages/covidcast/R/covidcast.R
```diff
@@ -301,13 +301,16 @@ summary.covidcast_signal = function(object, ...) {
 #'
 #' @details The argument structure is just as in `covidcast_signal()`, except
 #'   the first four arguments `data_source`, `signal`, `start_day`, `end_day`
-#'   are permitted to be vectors. The returned object is a list of
-#'   `covidcast_signal` data frames, of length `N = max(length(data_source),
-#'   length(signal))`. The first two arguments `data_source`, `signal` are
+#'   are permitted to be vectors. The first two arguments `data_source`, `signal` are
 #'   recycled appropriately, in the calls to `covidcast_signal()`; see example
 #'   below. The next two arguments `start_day`, `end_day`, unless `NULL`, must
 #'   be either length 1 or N.
```
ryantibs, line 307: This might be a bit confusing because `N` hasn't been defined in the details section (since you cut the line where it's defined). Though it is defined in the return section; and IIRC, the details are always printed after the return paragraph?

R-packages/covidcast/R/wrangle.R
```diff
@@ -281,7 +281,7 @@ covidcast_longer = function(x) {
 #' @rdname covidcast_longer
 #' @export
 covidcast_wider = function(x) {
-  if (!("covidcast_signal_long" %in% class(x))) {
+  if (!inherits(x, "covidcast_signal_long")) {
```
ryantibs, line 284: Good call, and can you please change the other two appearances of this logic: `if ("covidcast_signal" %in% class(x))`, in wrangle.R (I found them just now with grep).

R-packages/covidcast/_pkgdown.yml
```diff
@@ -8,3 +8,43 @@ home:
...
+
+reference:
+  - title: Fetch data
+    desc: Fetch data from the COVIDcast API
+  - contents:
+      - covidcast_signal
+      - covidcast_signals
+      - covidcast_meta
+  - title: Analyze and plot data
+    desc: Plot and summarize signals
```
ryantibs, line 20: I might suggest: 
title: Plot and summarize data
desc: Plot and summarize COVIDcast signals

R-packages/covidcast/_pkgdown.yml
```diff
@@ -8,3 +8,43 @@ home:
 
     - text: View the COVIDcast map
       href: https://covidcast.cmu.edu/
+
+reference:
+  - title: Fetch data
+    desc: Fetch data from the COVIDcast API
```
ryantibs, line 14: I might suggest: 
title: Fetch data
desc: Fetch signals from the COVIDcast API

R-packages/covidcast/_pkgdown.yml
```diff
@@ -8,3 +8,43 @@ home:
...
+  - title: Analyze and plot data
+    desc: Plot and summarize signals
+  - contents:
+      - plot.covidcast_signal
+      - print.covidcast_signal
+      - print.covidcast_meta
+      - summary.covidcast_signal
+      - summary.covidcast_meta
+  - title: Wrangle and correlate multiple signals
+    desc: Calculate correlations and move signals between formats for analysis
```
ryantibs, line 28: I might suggest: 
title: Wrangle and correlate multiple signals
desc: Calculate correlations, aggregate signals, and move between aggregated data formats

ryantibs:
Looks good @capnrefsmmat thanks for doing this. I just made a bunch of very small comments.

R-packages/covidcast/_pkgdown.yml
ryantibs: By the way do you think it's OK we go back and forth between "data" and "signal" (tend to use these two synonymously)?  On the one hand this is natural, on the other hand calling our signals "data" downplays the amount of work that we put into creating them.

R-packages/covidcast/R/covidcast.R
capnrefsmmat: Right, in pkgdown you get this:

Screen Shot 2020-11-01 at 4 14 38 PM

capnrefsmmat:
Tip: For small suggestions, the "Insert a suggestion" button lets you edit the original text directly, and then lets me apply your suggestion just by pressing a button on GitHub.

Screen Shot 2020-11-01 at 4 19 48 PM

capnrefsmmat:
Pushed fixes to your comments. I sidestepped the data/signal problem by writing "signals and metadata", so data refers to both.

In general I don't think we need to be persnickety about always using "signals"; the lay user is not going to respect whatever careful distinction we set up, and will tend to refer to things from the API as "data". But we can try to use "signals" when we specifically mean it.

ryantibs:
Looks good to me, thanks!

PR #234, opened 2020-11-01
Further R documentation fixes

capnrefsmmat:
See commit message for the first commit.

Also regenerates the HTML so the deployed docs site will update.


ryantibs (approved):
Looks good to me.

ryantibs:
Looks good to me.

PR #235, opened 2020-11-02
Add JHU, USA Facts and Safegraph example R coverage dashboards

benjaminysmith:
Per #32, add 3 more coverage dashboards. These are only basic coverage dashboards meant as an easy way to sanity check the API and can be enhanced later.


capolitsch:
What does a gray-colored county signify on the coverage maps? No data at all for the last 28 days? If this is the case the color ramp suggests it should be light yellow.

benjaminysmith:
> What does a gray-colored county signify on the coverage maps? No data at all for the last 28 days? If this is the case the color ramp suggests it should be light yellow.

It looks like grey = not present, vs. yellow = present but with 0 value.

It looks like this logic is in the covidcast R plot API, I can see an argument for distinguishing between not present vs present-but-0. If you feel strongly that it would help improve these dashboards we could try to massage the data a bit before plotting in the API (or switching to raw plots to avoid the covidcast behavior).

capolitsch:
> > What does a gray-colored county signify on the coverage maps? No data at all for the last 28 days? If this is the case the color ramp suggests it should be light yellow.
> 
> It looks like grey = not present, vs. yellow = present but with 0 value.
> 
> It looks like this logic is in the covidcast R plot API, I can see an argument for distinguishing between not present vs present-but-0. If you feel strongly that it would help improve these dashboards we could try to massage the data a bit before plotting in the API (or switching to raw plots to avoid the covidcast behavior).

Oh okay, so gray means we don't have data for that signal + location at *any* point in the past (not just the last 28 days)? If so, I agree not present vs present-but-0 is an interesting distinction. But personally I think we should provide a label for the gray that says that somewhere, or at least state it, if we are making these dashboards available for more than just internal checks.

benjaminysmith:
>> Oh okay, so gray means we don't have data for that signal + location at any point in the past (not just the last 28 days)? If so, I agree not present vs present-but-0 is an interesting distinction. But personally I think we should provide a label for the gray that says that somewhere, or at least state it, if we are making these dashboards available for more than just internal checks.

This is only for the last 28 days. I made this more clear by updating the title of the plots, and changing the range on the legend to start at 1 (rather than 0).

(fwiw, these are not public dashboards -- yet, though the notebooks are public).

capolitsch:
These look great, Ben.

benjaminysmith:
@krivard for merging

benjaminysmith:
(also fwiw, @capolitsch it looks like github wants you to formally approve -- I see: "At least 1 approving review is required to merge this pull request")

capolitsch (approved):
approved

benjaminysmith:
Thanks!

krivard (approved):
lgtm

PR #236, opened 2020-11-02
Memoize calls to `download_signal()` and `get_covidhub_prediction()`

### LINKED ISSUE #218: Cache calls to API and downloads from COVIDforecast hub
Current behavior:

`download_signals()` and `get_covidhub_predictions()` load data from, respectively, the COVIDcast API and the COVID19forecasthub Github site. These calls are slow. 

Desired behavior:

Fine to do these once, but repeated calls should use already downloaded data.

Suggested fix:

Use `{memoise}` to cache both of these calls. We need to address two use cases:
1. It should do this automatically without requiring setup. This means adding the caching of both functions `.onload`. For this version partially implemented, see https://github.com/cmu-delphi/covidcast/blob/2e730871f4278c629c91de68d80ecd3438fbb60b/R-packages/evalcast/R/package.R#L21.
2. Alternatively, a user may wish to cache to disk. We should issue a message onload that warns of the caching behavior and prompts a possible fix if users wish to alter this behavior.

Previous issues with potentially relevant discussion: #150, #107.

sgsmob:
Fixes #218.  Wraps `download_signal()` and `get_covidhub_prediction()` in in-memory `memoise` wrappers.

Please let me know if there is more I am supposed to be doing here.


dajmcdon:
I think this is a good start, but we also want to address @bnaras's concern about differential user behavior. 

So
- [ ] On load should also display a message that warns the user what's happening. See for example [in covidcast](https://github.com/cmu-delphi/covidcast/blob/4bd43f348dbc388071cd0cfd68ba25f5937376db/R-packages/covidcast/R/covidcast.R#L4)
- [ ] The message should also provide a workaround. I'm thinking: "If you want to cache to disk, run these two lines of code". I'm not quite sure what the appropriate lines are at the moment. Possibly also, "Run these lines to turn it off"

sgsmob:
> I think this is a good start, but we also want to address @bnaras's concern about differential user behavior.
> 
> So
> 
> * [ ]  On load should also display a message that warns the user what's happening. See for example [in covidcast](https://github.com/cmu-delphi/covidcast/blob/4bd43f348dbc388071cd0cfd68ba25f5937376db/R-packages/covidcast/R/covidcast.R#L4)
> * [ ]  The message should also provide a workaround. I'm thinking: "If you want to cache to disk, run these two lines of code". I'm not quite sure what the appropriate lines are at the moment. Possibly also, "Run these lines to turn it off"

Done (?)

PR #242, opened 2020-11-03
Use metadata$geo_type in example dashboards

benjaminysmith:
In https://github.com/cmu-delphi/covidcast/pull/156 we switched to using the metadata$geo_type, so update dashboard code to set this instead.


capnrefsmmat (approved):
If you've verified the dashboards run correctly with this change, it looks good to me, and you can merge

benjaminysmith:
Yep, confirmed that this fixes the breakage. I don't think I have permissions to merge, though.

benjaminysmith:
@krivard could you please merge?

PR #249, opened 2020-11-04
Remove unnecessary newlines by only emitting messages when not empty 

### LINKED ISSUE #245: Empty messages emitted by covidcast_signal() when queries fail to return data
See @capnrefsmmat 's comment below: happens more generally when queries fail to return data.

Case 1: `as_of < start_day` covidcast_signal() does (properly) return an empty dataframe.  However, it first emits a series of newlines; it could instead automatically catch this edge case, immediately return (and also not emit a bunch of newlines).

To reproduce:

```
library(covidcast)

start_day = "2020-04-15"
end_day = NULL

covidcast_signal("doctor-visits", "smoothed_adj_cli",
                 start_day, end_day,
                 as_of='2020-01-01')
```

Case 2: `as_of = '2020-10-01'` -- fetches data up to `2020-09-27`, and then emits a bunch of newlines.  

```
[...]
Fetched day 20200910: 1, success, num_entries = 1880
Fetched day 20200911: 1, success, num_entries = 1884
Fetched day 20200912: 1, success, num_entries = 1865
Fetched day 20200913: 1, success, num_entries = 1796
Fetched day 20200914: 1, success, num_entries = 1789
Fetched day 20200915: 1, success, num_entries = 1902
Fetched day 20200916: 1, success, num_entries = 1987
Fetched day 20200917: 1, success, num_entries = 1972
Fetched day 20200918: 1, success, num_entries = 1954
Fetched day 20200919: 1, success, num_entries = 1931
Fetched day 20200920: 1, success, num_entries = 1851
Fetched day 20200921: 1, success, num_entries = 1838
Fetched day 20200922: 1, success, num_entries = 1908
Fetched day 20200923: 1, success, num_entries = 1869
Fetched day 20200924: 1, success, num_entries = 1828
Fetched day 20200925: 1, success, num_entries = 1794
Fetched day 20200926: 1, success, num_entries = 1727
Fetched day 20200927: 1, success, num_entries = 1619

































There were 33 warnings (use warnings() to see them)
```

Case 3: `as_of = '2020-05-01`, whereas `start_date='2020-04-15'` - also a bunch of empty newlines.

benjaminysmith:
Fixes #245.

In R, sprintf returns an empty string if any of the arguments are NULL (see https://bugs.r-project.org/bugzilla/show_bug.cgi?id=16915 for details). This means that excessive newlines are rendered if we receive a (correct) NULL value.

To hide this, add a check on the message length before rendering.


R-packages/covidcast/R/covidcast.R
```diff
@@ -494,11 +494,16 @@ covidcast_days <- function(data_source, signal, start_day, end_day, geo_type,
...
-                    dat[[i]]$message,
-                    nrow(dat[[i]]$epidata)))
+    summary <- sprintf(
+      "Fetched day %s: %s, %s, num_entries = %s",
+      query_day,
+      dat[[i]]$result,
+      dat[[i]]$message,
+      nrow(dat[[i]]$epidata)
+    )
+    if (length(summary) != 0) {
```
capnrefsmmat, line 504: Is it only the case when `message != "success"` that we have NULLs? If so, we can just move this `message()` down into the `if (dat[[i]]$message == "success")` block just below, and not need this separate length check.

benjaminysmith: For this specific case that looks true (message = "no results"). I did it this way since the same bug could be triggered by any of the parameters being null (or the message being changed to include different, null paramters), so this fixes the general case of empty message = spurious newlines. If you feel that those are unlikely and prefer the simpler move into the success block, I'm fine with that too. wdyt?

capnrefsmmat: I suppose we can be defensive and keep it as-is, in case of weird things (like Epidata changing how it returns data in specific error cases)

Issue #251, opened 2020-11-04
msg` has been renamed to `message` 


gtzwang:
Hi, I just installed covidcast package. Getting this error. Any idea what that means? Thanks!

> covidcast_signal(data_source = "fb-survey", signal = "smoothed_cli", start_day = "2020-05-01", end_day = "2020-05-07",geo_type = "county")
Fetched day 2020-05-01: 1, success, num_entries = 1052

Error: `msg` has been renamed to `message` and is deprecated as of rlang 0.3.0

capnrefsmmat:
Ah -- it looks like this is a consequence of #151 and a change in `rlang`.

Specifically, this warning: https://github.com/cmu-delphi/covidcast/blob/9bf67daf59926879289eaff22b585023597e2eba/R-packages/covidcast/R/covidcast.R#L502-L516

When `geo_values = "*"`, the default, this incorrectly issues a warning, because `"*"` just means "all available". And since we include an argument named `msg`, `rlang` complains that is deprecated, even though we're not using it to mean a message (just to store the error message from the API).

@JedGrabman, since this came from #151, could you fix the geo check to not check for missing geos when requesting `"*"`, and change the name of the argument to the warning?

Then we should add this warning to our list of tests to implement in #239, so the tests would have caught this issue. Sorry I didn't see it in review.

JedGrabman:
Fix seems simple enough, opened a PR.
As a side note from testing, we seem to allow searches for vectors with "*" in them, which searches for the literal (i.e. You can set `geo_values = c("AK", "*")` which only returns results for Alaska). I matched the behavior with the warning (e.g. `Data not fetched for some geographies on 2020-05-01: * `) and I think the behavior is reasonable for covidcast_signal (although maybe a warning is warranted), but it seemed like an edge case worth mentioning.

capnrefsmmat:
Closing this as fixed by #252.

PR #252, opened 2020-11-05
Don't warn when retrieving all geo_values, fix rlang error

JedGrabman:
See Issue #251 for more details

### DIFF
--- a/R-packages/covidcast/R/covidcast.R
+++ b/R-packages/covidcast/R/covidcast.R
@@ -501,19 +501,21 @@ covidcast_days <- function(data_source, signal, start_day, end_day, geo_type,
                     nrow(dat[[i]]$epidata)))
     if (dat[[i]]$message == "success") {
       returned_geo_values <- dat[[i]]$epidata$geo_value
-      missed_geos <- setdiff(tolower(geo_value), 
-                             c(tolower(returned_geo_values), "*"))
-      if (length(missed_geos) > 0) {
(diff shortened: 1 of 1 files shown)


R-packages/covidcast/R/covidcast.R
```diff
@@ -501,19 +501,21 @@ covidcast_days <- function(data_source, signal, start_day, end_day, geo_type,
...
-               sprintf("Data not fetched for some geographies on %s: %s",
-                       query_day, missed_geos_str),
-             data_source = data_source,
-             signal = signal,
-             day = query_day,
-             geo_value = geo_value,
-             msg = dat[[i]]$message,
-             class = "covidcast_missing_geo_values"
-             )
+      if (!identical("*", geo_value)){
```
capnrefsmmat, line 504:
```suggestion
      if (!identical("*", geo_value)) {
```

capnrefsmmat (approved):
Looks good to me. One minor suggestion. And you've tested this works for both `"*"` and specific geo_values queries, right?

Later we need to add some automated tests for this so it's harder for us to mess up...

JedGrabman:
> and you've tested this works for both `"*"` and specific geo_values queries, right?

Yes, it worked for me and gave the appropriate warnings, with the caveats mentioned in issue #251 ([direct link](https://github.com/cmu-delphi/covidcast/issues/251#issuecomment-722573636)). Basically, `covidcast` treats `*` as a literal when part of a vector and results in a warning from this code because it doesn't return any results.

PR #254, opened 2020-11-06
Fix comma mismatch

sgsmob:
One of arguments in `plot_trajectory()` had a trailing comma from a Chinese keyboard rather than the ASCII character.  This PR replaces it with the correct comma.


chinandrew (approved):
:shipit: :shipit: :shipit:

dajmcdon (approved):

Issue #239, opened 2020-11-03
Add additional tests to covidcast


capnrefsmmat:
#217 adds a starter set, but we need more. These can be added piecemeal over many PRs, so they're good starter tasks for anyone who'd like to try working on the R package.

- [ ] Bubble plots. Particularly the behavior when counties have missing data, have data present but have 0 counts, or have nonzero data. They should be tested for both county and state levels, and also when there are megacounties.
- [ ] More on correlations. We do not currently test `by = "time_value"` at all, nor do we test different correlation types (Spearman, etc.). And when #33 is fixed, we should test it.
- [ ] `aggregate_signals` takes several different formats for `dt` depending on how many data frames are provided; we should test all of these formats.
- [ ] `covidcast_signal` should be tested (using mocks) to ensure it formats its output correctly.
- [ ] `covidcast_signal`'s error messages (when a signal or day is missing) should be tested so they don't regress
- [ ] It should be possible to call `covidcast_signal` on signals with no metadata, provided you provide `start_day` and `end_day`.

@ryantibs you're welcome to add other test ideas to this list for others to take on.

ryantibs:
Looks good to me, thank you Alex.  Atm I can't think of much else to add.  But will leave comments here as I keep working on development to see if new things come to mind.

Nb @jacobbien @bnaras @dajmcdon on the evalcast side: you can also look at #217 to what Alex already did and how he set it up through testthat.

capnrefsmmat:
The only thing remaining here is to add additional correlation tests, so I'm renaming this.

PR #256, opened 2020-11-08
Second batch of R tests

capnrefsmmat:
This addresses a few of the items in #239.

The httptest package will be very helpful here. I was going to use `mockery::stub`, but it misbehaves when you want to repeatedly swap out the stub to return different values. Instead, httptest serves the API HTTP requests from a file instead of letting them go to the server. See the comment at the top of `test-covidcast.R` for more details.

It's a bit clunky (particularly keeping track of the JSON) but it is *much* better than having no tests at all.

cc @JedGrabman in case you'll be working R package features


R-packages/covidcast/R/covidcast.R
```diff
@@ -177,7 +177,9 @@ covidcast_signal <- function(data_source, signal,
                   .data$geo_type == given_geo_type)
 
   if (nrow(relevant_meta) == 0) {
-    relevant_meta <- list()
+    # even if no other metadata is available, we should set the geo_type so
+    # plotting functions can deal with this signal.
+    relevant_meta <- list(geo_type = geo_type)
```
sgsmob, line 182: How much does this change the previous behavior?  Is it needed for testing?  All else equal, I would prefer that this PR only add tests.

sgsmob (approved):

R-packages/covidcast/tests/testthat/test-covidcast.R
```diff
@@ -1,25 +1,42 @@
...
+# Many of these tests use mockery::with_mock_api. This replaces calls to the
+# live API server, instead returning static JSON files from disk. Three
+# weaknesses to be aware of:
+#
+# 1. The HTTP URL and parameters are hashed into a single filename to use. This
+# means that you can't have the query return one thing, then have the same query
+# return something else, for instance to test how the client handles different
+# types of errors.
+#
+# 2. Once you've written a test, it can be difficult to find the file storing
```
sgsmob, line 14: Can you also add comments in the json files to indicate to which query they belong?

capnrefsmmat: what's your preferred way for adding comments in JSON files? a dummy key/value pair?

R-packages/covidcast/R/covidcast.R
capnrefsmmat: See the commit message in 85c1ca2. It addresses the last point in #239. This behavior has regressed several times, and regressed again in #156.

I could separate the fix into another PR, but basically I got halfway through writing the test when I realized the code had regressed again, so I decided to include the fix. Separating it out may be more trouble than it's worth for a one-liner.

R-packages/covidcast/tests/testthat/test-covidcast.R
sgsmob: Ah didn't realize json didn't support comments.  Ignore it then.

R-packages/covidcast/R/covidcast.R
sgsmob: Sounds good to leave it in as is

capnrefsmmat:
@sgsmob so given the comments above, is this ready to merge?

sgsmob:
yes

Issue #222, opened 2020-10-31
Print methods for cards


dajmcdon:
See @jacobbien 's discussion in #98.

Not clear if we want full S3 classes for pcards and ecards (see #150).

Minimum deliverables required:
- [ ] Both types of cards should have a print method to avoid the problem mentioned in #98 

Complete solution:
- [ ] Do we actually give these a class? 
- [ ] Perhaps the class only propagates to print and `aggregate_cards()` (#221 )
- [ ] Probably do not want plot methods for these (aggregating and then just passing to ggplot gives more freedom).
- [ ] How do we handle 1 set of cards versus a list of cards?

sgsmob:
Essentially should print out a formatted `all_attr`.

Issue #221, opened 2020-10-31
Aggregate cards into a tibble


dajmcdon:
Create a function `aggregate_cards()`.

- [x] Should appropriately handle prediction cards or evaluation cards
- [x] Should also accommodate lists of these, for instance the output from `get_predictions()` run twice with different arguments.
- [ ] Ensure that this function is used throughout `evalcast` (eg in `plot_trajectory()`)

See also #98

sgsmob:
I'm reading back through the previous discussions and I am not clear on the input/output types here.

On the input side am I correct that we should expect:
* A list of 2-column data frames with columns:
    * Geo code
    * 23 by 2 tibble with columns:
        * Quantile
        * Predicted value (prediction cards) or error (evalutait
?

On the output side, should there be a single data frame in the same format as above such that the predicted values are averaged over all elements of the list?  The same thing but the predicted values are a list of the values (i.e., go from a list of frames to a frame of lists)?  Something completely different?

dajmcdon:
Something a bit different. Basically the geo code should be repeated for every one of the 23 quantile/value pairs. One also needs the many of the attributes of the card. If you look from [here](https://github.com/cmu-delphi/covidcast/blob/9c5848152b966fb97cff666059ee8574e07b16e9/R-packages/evalcast/R/plot_trajectory.R#L34) until about line 48, that should be most of it. It may also help to see the [notebook](https://htmlpreview.github.io/?https://github.com/cmu-delphi/covidcast/blob/covidhub-eval/R-notebooks/covidhub_evaluation.html) Ryan posted this morning. He basically did it manually there.

sgsmob:
So the output is a single tibble with columns:

- Geo code
- Forecaster name
- Number of weeks ahead
- Quantile
- Predicted value or error

and there is a 1:1 correspondence between each row in the output and a row in one of the cards in the list?

dajmcdon:
that's right. Produce 1 tibble. But you will also, need some other things. If you look at line 162 of `get_predictions.R`, all the attributes are listed there. I don't think we need `forecaster` or `corrections_applied` since they are functions. Probably not `forecaster_params` either (maybe make that an option to the function). On the `score_cards` you need much of the same plus some of the other things. 

Imagine I run 2 forecasters each at 3 aheads on 5 different forecast dates. I want to put everything in one tibble that contains any information I would want to try to summarize or plot the output. Same for the score cards.

Issue #150, opened 2020-10-26
evalcast, caching, s3 method


bnaras:
jacobbien dajmcdon It seems to me that caching should always be available for the user to control. So any function we provide that can cache should take a boolean flag as argument (maybe default TRUE) along with a `cache_path`, which, if null, reverts to memory cache. This is because if such a function is being run in a production system at a time that straddles the instant when new covidcast data becomes available, making caching the default will ensure that the stale data remains in that session, silently. The user should have control.

If I am not mistaken, in `baseline_forecaster()` the use of `download_signal()` seems to do some computations to get the `location` variable. While we can use such a function internally, I don't think `baseline_forecaster()` (which I am thinking of as a template for other forecasters users might write) should hide it. If a common task, this functionality can directly provided in an exported function in covidcast.  Then, a vignette can clearly explain how to use those covidcast functions directly to produce the df that is the first arg to `baseline_forecaster()`.

As I am adding S3 plots methods, etc., I am not getting a sense of what is gained by it. Taking `lm()` as an example, there is a very obvious concept of the model object, its behavior, methods etc.  I can't seem to figure out an analogous reason to elevate prediction_cards to an object, other than the fact that they can be plotted and printed. They are tibbles after all, and tidyverse provides ways to slice and dice them any which way. I might be missing something, so please let me know.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970432225928) by [Unito](https://www.unito.io/learn-more)

dajmcdon:
I don't have a problem with forcing caching to be under user control, but it is a much more "quiet" fix to do it this way. It does lose flexibility though. The cache would be more readable if we did it differently. I think there are multiple ways to do this depending on how much control do we want to give the user. We could turn it off by default and give a message at load with a line of code to run. That would allow selecting between memory and disk. But via `memoise`, it's hard to access the cache for the user. Passing the argument through everywhere would allow more control of the name of the stored objects, but maybe be more painful to use.

dajmcdon:
[...]

PR #264, opened 2020-11-11
Add S3 class typing to cards

sgsmob:
Annotate an S3 class onto each tibble in the lists returned by `evalcast::get_covidhub_predictions()` and `evalcast::evaluate_predictions()` ("prediction_card" and "score_card", respectively).  This will allow us to have generic methods for printing (#222) and aggregating (#221) the cards while allowing specialized behavior for the two types of cards.

See also #150.


sgsmob:
Oops I need to sync this to `evalcast` and not `main`.

sgsmob:
probably going to be easier to delete this PR and try again.

Issue #224, opened 2020-10-31
Fail gracefully


dajmcdon:
From @ryantibs:

>  if anything goes wrong in evaluating predictions, at any of the forecast dates or any of the aheads, for example, then it quits (via an `assert_that()` statement).  I think we need to go through all the `assert_that()` statements and ask---is this really needed, or can we issue a warning, return an `NA` from that function, and rewrite the parent functions to proceed gracefully? 

- [x] Ensure that these sorts of errors/warnings happen only when necessarry.

ryantibs:
To be clear, I meant that this should be true overall, across the whole evalcast package, and not just within `evaluate_predictions()`.

bnaras:
I agree that `assert_that()` should catch errors, not warnings or situations where an `NA` return would make sense.

Issue #223, opened 2020-10-31
Appropriate behavior of `get_covidhub_predictions()`


dajmcdon:
- [x] Should cache appropriately (#218 )
- [x] Should also grab "point" forecasts (optionally, only these)
- [x] Clean up documentation throughout this file. All three functions should be exported for now.

ryantibs:
Tiny comment: we may as well expose `get_covid_hub_forecast_names()` to be a public function.  This would be useful to have.  And, for consistency, I would change `covid_hub` in this name to `covidhub`.  And, I would make that change throughout (several argument names appear to have `covid_hub` in them, rather than `covidhub`, which you can catch with `grep "covid_hub" evalcast/R/*.R`.

brookslogan:
I am confused by the output/documentation of this function in other aspects; `evalcast::get_covidhub_predictions("COVIDhub-baseline", as.Date("2020-09-07"))` produces an unnamed list of 20 cards, some with 7-row `forecast_distribution`s, some with 23-row `forecast_distribution`s.  Looking at the csv downloaded, there appear to be only 16 distinct `target`s though, or 12 "inc"-type `target`s; I am not sure why there are 20 cards or what these 20 cards are.  Not sure if it is my version of packages resulting in unnamed lists or other issues.

```
> sessionInfo()
R version 3.5.0 (2018-04-23)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS High Sierra 10.13.6

Matrix products: default
BLAS: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRblas.0.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] evalcast_0.1.1  testthat_3.0.0  covidcast_0.3.0 tibble_3.0.4   
[5] pipeR_0.6.1.3  

loaded via a namespace (and not attached):
 [1] zoo_1.8-8         tidyselect_1.1.0  remotes_2.2.0     purrr_0.3.4      
 [5] lattice_0.20-41   colorspace_1.4-1  vctrs_0.3.4       generics_0.1.0   
[...]

PR #265, opened 2020-11-11
Add S3 class typing to cards

sgsmob:
Annotate an S3 class onto each tibble in the lists returned by `evalcast::get_covidhub_predictions()` and `evalcast::evaluate_predictions()` ("prediction_card" and "score_card", respectively). This will allow us to have generic methods for printing (#222) and aggregating (#221) the cards while allowing specialized behavior for the two types of cards.

See also #150.

### DIFF
--- a/R-packages/evalcast/R/evaluate.R
+++ b/R-packages/evalcast/R/evaluate.R
@@ -159,6 +159,7 @@ evaluate_predictions_single_ahead <- function(predictions_cards,
          backfill_buffer = backfill_buffer,
          as_of = as_of)
   )
+  class(score_card) <- c("score_card", class(score_card))
   return(score_card)
 }
 
--- a/R-packages/evalcast/R/get_covidhub_predictions.R
+++ b/R-packages/evalcast/R/get_covidhub_predictions.R
@@ -82,7 +82,7 @@ get_covidhub_predictions <- function(covid_hub_forecaster_name,
                geo_values = NA,
                from_covidhub = TRUE)
         )
-
+        class(.x) <- c("prediction_card", class(.x))
         return(.x)
       })
   }


dajmcdon (approved):

dajmcdon:
Sorry, @sgsmob I approved this too quickly and had to revert. 

The "score_card" in `evaluate.R` is just one card, while the `prediction_cards` from covidhub may be 1 card, or more likely, many. 

Both aggregating and printing need to handle 1 or many. My instinct is that we should always expect "cards" and attach the class as plural to the list of lists rather than to each individual card. Your current version classes 1 card in `evaluate.R` and plural cards in `get_covidhub_predictions`.

Prediction cards (generally many) are also produced by `get_predictions()`, so we need the same result there.

sgsmob:
I think we want the cards to be labeled; the divergent behavior between the aggregation functions is actually (here)[https://github.com/cmu-delphi/covidcast/blob/9c5848152b966fb97cff666059ee8574e07b16e9/R-packages/evalcast/R/plot_trajectory.R#L43] where we are iterating over the cards.

dajmcdon:
My feeling is that the aggregation should begin on the line above your link (line 42). So we take in a list of cards, and implement the processing that goes from 43 to 58. Does that make sense? We want to turn all cards at once into one giant tibble.

sgsmob:
It does make sense but we don't have to decide what type of card we are dealing with until we unpack the cards in L43.  So the workflow would look like:

```
unpack_single_card <- function(card) {
    UseMethod("unpack_single_card", card)
}

unpack_single_card.prediction_card <- function(card) {
     # ...
}

unpack_single_card.evaluation_card <- function(card) {
    # ...
}

aggregate_cards <- function(list_of_cards) {
    return list_of_cards %>% purrr::map_dfr(unpack_single_card)
}
```

dajmcdon:
Ok. Works for me.

sgsmob:
Logan's example in #223 does not work because the dimensions of the cards are not the same.  I never looked closer into that because I assumed I was doing something wrong since I didn't understand the output.  I will look into what assumptions the snippit from `plot_trajectory.R` was making and why it rejects dimension-mismatched cards.

dajmcdon:
I think the trajectory plotting (and aggregation more generally) should accommodate the dimension mismatch. So we'll have to fix that up. This is a good example of the need to "fail gracefully" #224. Aggregating these shouldn't be an issue, nor should plotting.

PR #266, opened 2020-11-11
Faster R package vignettes

capnrefsmmat:
With some judicious optimization of `latest_issue`, we can make the correlation vignette several hundred times faster to build. I'm also switching it to be at the state level to save on data download time.

See the commit message in 4e9875c for an explanation of the speed improvements.

This is a work in progress; my goal is to make the vignettes fast enough that we can change the install instructions to build them. On my machine they take about 2 minutes to build, about two-thirds of which is waiting for data from the API; we can probably cut that down by using fewer datasets.


capnrefsmmat:
@ryantibs This isn't done, but you may like it just because it makes your correlation analyses ludicrously faster (along with anything using `latest_issue`). Maybe that will help some forecasting or modeling work that uses `covidcast_cor` for exploration.

ryantibs:
@capnrefsmmat Nice job.  Generally lgtm but just a few thoughts for you to consider. 

- This will return a `covidcast_signal` data frame with scrambled rows. Would it be worth reordering? This doesn't affect any of our downstream uses like plotting or correlations, but may confuse a user if they call `latest_issue()` directly.

- As far as I understand, `latest_issue()` doesn't check the class?  So it could in principle be applied to a `covidcast_signal_long` data frame?  And in this case it wouldn't return the right result, since you're masking by `geo_value` and `time_value` in the call to `dplyr::distinct()`.  Thus I would either check the class in `latest_issue()` ... or include `data_source` and `signal` as masking variables.

PR #267, opened 2020-11-11
Rename vignettes

ryantibs:
So that they will appear in the order we want them to.

### DIFF
--- a/R-packages/covidcast/_pkgdown.yml
+++ b/R-packages/covidcast/_pkgdown.yml
@@ -9,6 +9,16 @@ home:
     - text: View the COVIDcast map
       href: https://covidcast.cmu.edu/
 
+articles:
+  - title: Using the package
+    desc: Basic usage and examples.
+    navbar: ~
+    contents:
+      - covidcast
+      - plotting-signals
+      - correlation-utils
+      - multi-signals
+
 reference:
   - title: Fetch data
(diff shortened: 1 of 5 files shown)


ryantibs:
@capnrefsmmat I didn't mean to include the first commit f279444, that was a branch mistake (I branched from my covidhub-eval branch).  But that branch should be merged into main anyway, per #240.  Once Katie merges that, you can just merge this one, to save yourself a cherry-picking task.

ryantibs:
@capnrefsmmat I just fixed the correlation vignette to study case and death *rates* (rather than *counts*), since the former is the only thing that makes sense for "sliced by time".  (Oops! what a goof.)  And for "sliced by location", they're equivalent.  

Note: I force added the html file to save you from rerunning it, since it takes so long to knit.  You can see I rewrote the text a bit because the results changed.  You can delete the html file from the repo once you've looked at it.

capnrefsmmat:
FYI adding the HTML is unnecessary, since I have to rebuild the vignette to make the pkgdown site anyway. Soon #266 will make the vignette super-fast and this won't be a problem.

I'll rebuild the site, check the result, and merge this in.

ryantibs:
Thanks!  (I just meant to add the html so you could look at it right away, and if you spotted anything, you could reply without needing to rebuild the pkgdown site.)

PR #271, opened 2020-11-11
Use CSV format for API access in R package

capnrefsmmat:
These changes make the R package download data from the API using its CSV download format. This format is much more compact than JSON. This also improves tests related to the data fetch behavior, since these improvements make me more confident in this PR's correctness.

There would be nominal bandwidth improvements, but the main benefit is for #266: I plan to use httptest's ability to make vignettes use static data instead of requesting data from the API. The JSON files for big API downloads are *huge*, while the CSVs will be much more compact, preventing the repo from bloating.

This depends on https://github.com/cmu-delphi/delphi-epidata/pull/281; the servers should support compression for CSVs before we start directing clients to to the CSV downloads.


R-packages/covidcast/R/covidcast.R
```diff
@@ -437,18 +437,21 @@ covidcast_signals <- function(data_source, signal,
 #'
 #' @export
 covidcast_meta <- function() {
-  meta <- .request(list(source='covidcast_meta', cached="true"))
+  meta <- .request(
+    list(source = "covidcast_meta",
+         format = "csv"))
```
sgsmob, line 442: Are we purposefully removing caching behavior here?

R-packages/covidcast/R/covidcast.R
```diff
@@ -529,58 +532,59 @@ covidcast_days <- function(data_source, signal, start_day, end_day, geo_type,
...
-    dat[[i]] <- covidcast(data_source = data_source,
-                          signal = signal,
-                          time_type = "day",
-                          geo_type = geo_type,
-                          time_values = day_str,
-                          geo_value = geo_value,
-                          as_of = as_of,
-                          issues = issues,
-                          lag = lag)
+    res <- covidcast(data_source = data_source,
```
sgsmob, line 535: Not really keen on the variable name "res" but I don't have a better, more descriptive suggestion off the top of my head.

R-packages/covidcast/R/covidcast.R
```diff
@@ -529,58 +532,59 @@ covidcast_days <- function(data_source, signal, start_day, end_day, geo_type,
...
+           class = "covidcast_fetch_failed")
+
+      next
+    }
+
+    dat[[i]] <- res
+
     summary <- sprintf(
-      "Fetched day %s: %s, %s, num_entries = %s",
+      "Fetched day %s: num_entries = %s",
```
sgsmob, line 560: Does the change to this line represent a change to the semantics of the output or is it just a change in compatibility between the two object types?  That is to say, the old `dat[[i]]` has attributes `result`, `message`, and `epidata`--does the new `dat[[i]]` have something different?

R-packages/covidcast/R/covidcast.R
```diff
@@ -529,58 +532,59 @@ covidcast_days <- function(data_source, signal, start_day, end_day, geo_type,
...
-               data_source = data_source,
-               signal = signal,
-               day = query_day,
-               geo_value = geo_value,
-               api_msg = dat[[i]]$message,
-               class = "covidcast_missing_geo_values"
-               )
-        }
+
+    if (nrow(dat[[i]]) > 0 && !identical("*", geo_value)) {
```
sgsmob, line 568: Why are we using `identical()` and not just `==`?

R-packages/covidcast/R/covidcast.R
```diff
@@ -529,58 +532,59 @@ covidcast_days <- function(data_source, signal, start_day, end_day, geo_type,
...
+      returned_geo_values <- dat[[i]]$geo_value
```
sgsmob, line 569: Are there multiple geovalues here?  Should this just be renamed `returned_geo_value`?

R-packages/covidcast/R/covidcast.R
```diff
@@ -645,7 +651,16 @@ covidcast <- function(data_source, signal, time_type, geo_type, time_values,
   }
 
   # Make the API call
-  return(.request(params))
+  res <- .request(params)
```
sgsmob, line 654: Still not thrilled with `res`.

R-packages/covidcast/tests/testthat/api.covidcast.cmu.edu/epidata/api.php-32641f.csv
```diff
@@ -0,0 +1,3 @@
+geo_value,signal,time_value,direction,issue,lag,value,stderr,sample_size
+01001,bar,20200110,,20200111,1,91.2,0.8,114.2
+01002,bar,20200110,,20200111,1,99.1,0.2,217.8
```
sgsmob, line 3: Add a newline here.

R-packages/covidcast/tests/testthat/api.covidcast.cmu.edu/epidata/api.php-f49e8f.csv
```diff
@@ -0,0 +1,3 @@
+geo_value,signal,time_value,direction,issue,lag,value,stderr,sample_size
+31001,bar,20200112,,20200113,1,81.2,0.8,314.2
+31002,bar,20200112,,20200113,1,89.1,0.2,417.8
```
sgsmob, line 3: Newline

R-packages/covidcast/tests/testthat/test-covidcast.R
```diff
@@ -104,8 +105,51 @@ with_mock_api({
...
+          max_time = as.Date("2020-01-02"),
+          min_value = 0,
+          max_value = 10,
+          max_issue = as.Date("2020-04-04")),
+          class = c("covidcast_meta", "data.frame"))))
+  })
+
   test_that("covidcast_signal stops when end_day < start_day", {
-    # reusing api.php-da6974.json
+    # reusing api.php-d2e163.json for metadata
```
sgsmob, line 152: Is this still json?

sgsmob (changes requested):
A lot of comments here based on me not really knowing what is going on here.  But this is in pretty good shape.

R-packages/covidcast/R/covidcast.R
capnrefsmmat: Hysterical Raisins™. Caching behavior used to be opt-in, and if you didn't request the cached version, it would take forever to load. All requests now get the cached version: https://github.com/cmu-delphi/delphi-epidata/commit/ed19bc4f77f40e8695c04a13a2d20a2072cfab47#diff-a9c5ce7767493542e52e74713cbbeacdee649c91a5c933f3b9981f5173826a3b

So that parameter was just wasting a few bytes.

R-packages/covidcast/R/covidcast.R
capnrefsmmat: Good question, and perhaps we should document this more clearly in a comment. The old format would return error codes in the JSON. The CSV format does not have a field for error messages; instead it uses HTTP status codes, so the 

```r
 httr::stop_for_status(response, task = "fetch data from API")
```

line in `.request` will do the main error reporting if we don't get HTTP 200 back.

R-packages/covidcast/R/covidcast.R
capnrefsmmat: If `geo_value` is a vector, `"*" == geo_value` will be a vector of logicals instead of one boolean value.

R-packages/covidcast/R/covidcast.R
capnrefsmmat: Yes, there are multiple, since `dat[[i]]$geo_value` is a column of a data frame and may have multiple entries

R-packages/covidcast/R/covidcast.R
capnrefsmmat: How about `response`?

R-packages/covidcast/tests/testthat/test-covidcast.R
capnrefsmmat: Good catch, and the filename was wrong too (the tooling for httptest filename tracking is pretty clunky)

sgsmob (approved):

capnrefsmmat:
@sgsmob Planning question. This PR, #266, and #216 together represent a decent amount of improvements and new features. It may be worth designating these as a new release.

Since we have people use `devtools::install_github()`, merging these PRs makes them instantly available to new users. Do you think we should merge these PRs in a coordinated way? Just dribble them out? Get formal and have a development branch that we merge to main when we want to "release"?

sgsmob:
I don't have a good intuition around releases.  @benjaminysmith and @tildechris might have a better sense for it or at least some insight from the practices we have adopted in engineering.

capnrefsmmat:
I propose something like

1. I create a release branch
2. I merge these PRs into the release branch
3. We make a PR for it so CI can test
4. Merge in one big batch when it's done and we're ready (including changelog, etc.)

chinandrew:
> I propose something like
> 
> 1. I create a release branch
> 2. I merge these PRs into the release branch
> 3. We make a PR for it so CI can test
> 4. Merge in one big batch when it's done and we're ready (including changelog, etc.)

We were discussing branching strategies for the indicator repos and landed on something like gitflow, where everyone merges continuously into one branch, and releases are cut and merged into the production branch whenever we want. Could see a similar thing working here too.

R-packages/covidcast/R/covidcast.R
```diff
@@ -644,8 +650,19 @@ covidcast <- function(data_source, signal, time_type, geo_type, time_values,
...
+  # there are no results for our query.
+  response <- .request(params)
+  if (nchar(response) == 0) {
+    # empty if no results
+    return(NULL)
+  }
+
+  # geo_value must be read as character so FIPS codes are returned as character,
+  # not numbers (with leading 0s potentially removed)
+  return(read.csv(textConnection(response), stringsAsFactors = FALSE,
```
benjaminysmith, line 664: Are there any possible issues with large data sizes here? (which might work differently with CSV vs json)

benjaminysmith (commented):
Conceptually this makes sense, but I am a but nervous that this is a pretty substantial change to the way the API client works -- how can we be confident that it works as a drop-in replacement? Testing in our own code? (that might be fine -- just calling it out)

Another option would be to have this as an alternative method in parallel at first (maybe that can be enabled through a flag or parameter) and then switch over to using it by default once it has been tested in our code. What do you think?

R-packages/covidcast/R/covidcast.R
capnrefsmmat: `textConnection` does copy the response, and I don't know how jsonlite's performance compares to R's native CSV reader. Not sure how we'd test this, though; maybe benchmarking a call that returns a particularly large dataset?

My guess is that the smaller size of CSV responses outweighs any change in performance, but that's just a guess

capnrefsmmat:
> Conceptually this makes sense, but I am a but nervous that this is a pretty substantial change to the way the API client works -- how can we be confident that it works as a drop-in replacement? Testing in our own code? (that might be fine -- just calling it out)

Yeah, I'm nervous too. That's why I've added more tests here, and also verified the vignettes run. But since we're brand-new to testing for this package, it's entirely possible there's something I've missed.

I think I'd be most comfortable if we prepare a release branch and have some Delphi members use it for their work for a week and see if any bugs arise. Any objections to the idea of preparing a release branch for that purpose? We could move to that model for future work, where we always merge to a dev branch and beta-test before release. I think that's in line with @chinandrew's suggestion.

benjaminysmith (approved):
Looks good to me as sounds like you are considering safe rollout options. i think the main concerns are 1) detecting if there are problems when this goes out (I suspect this would rely on user feedback); 2) having a way for users to roll back. Since this is installed through devtools, is there an easy way to revert to the previous version? if not might be worth putting this behind  a parameter or ffalg.

capnrefsmmat:
`devtools::install_github` takes a `ref` argument, which can be a branch, commit, or tag, so we can always point back to a prior commit or to a different branch if needed.

I'll need to point this branch at `r-pkg-devel` and then fix some tests after it goes in, probably after #275.

capnrefsmmat:
I believe this is ready for re-review. Changes since last time:

* I rebased this on top of #275, which required some changes to tests and tweaks to the `covidcast_days` function now that it batches requests.
* I updated the NEWS file to describe this change (and #275).

cc @JedGrabman, since this touches your batching code

sgsmob (approved):

Issue #270, opened 2020-11-11
Signals without metadata fail to plot


tildechris:
```r
changehc = covidcast_signal(
  data_source = "changehc", 
  signal = "smoothed_adj_outpatient_covid", 
  start_day = "2020-11-04", end_day = "2020-11-04", 
  geo_type = "county")
plot(changehc)
```
Currently yields an error:

```
Error in seq.default(range[1], range[2], len = n) : 'from' must be a finite number
6. stop("'from' must be a finite number")
5. seq.default(range[1], range[2], len = n)
4. seq(range[1], range[2], len = n)
3. plot_choro(x, time_value = time_value, include = include, range = range, col = choro_col, alpha = alpha, direction = direction, dir_col = dir_col, title = title, params = choro_params)
2. plot.covidcast_signal(changehc)
1. plot(changehc)
```

Because meta-data isn't yet being returned by the API.  Manually setting the range to min/max of the value works around the error:

```
min_value <- max(changehc$value)
max_value <- min(changehc$value)

plot(changehc, range=c(max_value,min_value))
```

As per @capnrefsmmat, we should at least return a better error message in this case.

capnrefsmmat:
The problem is this block: https://github.com/cmu-delphi/covidcast/blob/f3e3b3c12936c8d65cdebdf18c6478b41c2da5f9/R-packages/covidcast/R/plot.R#L113-L123

Since we now set `attributes(x)$metadata$geo_type` for all data frames returned from `covidcast_signal`, even those without other metadata, the check does not detect the missing mean and stdev. The check should be updated and a unit test added to verify the warning is correctly raised in this case.

chinandrew:
This is fixed in #326

PR #275, opened 2020-11-12
Batch requests when requesting multiple days of data

JedGrabman:
Description:
Improve efficiency of covidcast_days by requesting multiple days of data when it can be determined that it is safe to do so.

Change Log:
- covidcast.R
  - Update covidcast_days to query multiple days at once
  - Create max_geo_values to find upper bounds on number of results returned
  - Create warning messages for missing dates or missing geo_values
- covidcast.R
  - Add tests for added batching functionality
  - Add basic documention for the use of `mock` and `stub`


R-packages/covidcast/tests/testthat/test-covidcast.R
```diff
@@ -118,3 +135,102 @@ with_mock_api({
...
+    covidcast_days(
+      data_source = "fb-survey",
+      signal = "raw_cli",
+      start_day = ymd("2020-10-01"),
+      end_day = ymd("2020-10-06"),
+      geo_type = "county",
+      geo_value = "*",
+      as_of = NULL,
+      issues = NULL,
+      lag = NULL
```
sgsmob, line 231: should be indented more

sgsmob (changes requested):

sgsmob:
Please add a description of this PR.

capnrefsmmat:
This will also run into conflicts with #271, mainly just in the unit tests (should be easy to deal with). We should plan the order of operations here when we merge.

(but thanks -- this should be a great benefit to download speed)

R-packages/covidcast/R/covidcast.R
```diff
@@ -515,32 +515,68 @@ summary.covidcast_meta = function(object, ...) {
...
 }
 
-##########
+# Find the maximum number of rows the API is expected to return per day of data
+max_geo_values <- function(data_source, signal, geo_type) {
+  meta_info <- covidcast_meta()
+  max_locations <- meta_info[meta_info$data_source == data_source &
+                              meta_info$signal == signal &
+                              meta_info$geo_type == geo_type, ]$num_locations
+  if (length(max_locations) == 0) {
```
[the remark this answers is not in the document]
krivard: I believe it's a pre-emptive fix for a similar effect of #270, but yeah, a comment would be good here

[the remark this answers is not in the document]
JedGrabman: That's right, it's a defensive measure against signals that aren't documented in `covidcast_meta` and using the max as an upper bound when it can't be found. Added a comment.

R-packages/covidcast/R/covidcast.R
```diff
@@ -515,32 +515,68 @@ summary.covidcast_meta = function(object, ...) {
...
+  if (identical(geo_value, "*")) {
+    ngeos <- max_geo_values(data_source, signal, geo_type)
+  } else {
+    ngeos <- length(geo_value)
+  }
+  if (length(issues) == 2) {
+    nissues <- as.numeric(issues[2] - issues[1]) + 1
+  } else {
+    nissues <- 1
+  }
```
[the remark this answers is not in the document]
JedGrabman: From `help("covidcast_signal")`:
>Provided as either a single Date object (or string in the form "YYYY-MM-DD"), indicating a single date to fetch data issued on, or a vector specifying two dates, start and end. In this case, return all data issued in this range. 

Added a comment.

R-packages/covidcast/R/covidcast.R
```diff
@@ -515,32 +515,68 @@ summary.covidcast_meta = function(object, ...) {
...
+  max_days_at_time <- floor(3649 / (ngeos * nissues))
```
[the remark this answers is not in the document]
JedGrabman: Done

R-packages/covidcast/R/covidcast.R
```diff
@@ -515,32 +515,68 @@ summary.covidcast_meta = function(object, ...) {
...
+
+  # In theory, we could exceed max rows with 1 day, but try anyway
+  if (max_days_at_time == 0) {
+    max_days_at_time <- 1
+  }
+  batch_days <- ceiling(ndays / max_days_at_time)
```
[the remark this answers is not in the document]
JedGrabman: Done

R-packages/covidcast/tests/testthat/test-covidcast.R
JedGrabman: Done

R-packages/covidcast/tests/testthat/test-covidcast.R
```diff
@@ -118,3 +137,101 @@ with_mock_api({
...
+        end_day = ymd("2020-10-06"),
+        geo_type = "county",
+        geo_value = "*",
+        as_of = NULL,
+        issues = NULL,
+        lag = NULL
+      ),
+      regexp = NA)
+  expect_called(m, 2)
+})
```
sgsmob, line 237: Add a newline here.

sgsmob (approved):
Just the minor newline nit, but otherwise this is good to go.

R-packages/covidcast/R/covidcast.R
```diff
@@ -515,32 +518,76 @@ summary.covidcast_meta = function(object, ...) {
   invisible(df)
 }
 
-##########
+# Find the maximum number of rows the API is expected to return per day of data
+max_geo_values <- function(data_source, signal, geo_type) {
+  meta_info <- covidcast_meta()
```
capnrefsmmat, line 523: `covidcast_signal()` already calls `covidcast_meta()`. Could it be more appropriate for `covidcast_signal()` to call `max_geo_values()` directly, providing it the necessary metadata, and then give `max_locations` as an argument to `covidcast_days`?

Otherwise each `covidcast_signal()` call would require two `covidcast_meta()` calls, which inflates API calls a fair bit.

JedGrabman: I combined `max_geo_values` with some filtering work that was already being done within `covidcast_signal` and renamed it to `specific_meta`. `max_geos` is now an optional parameter for `covidcast_days`, if not provided it assumes a large number of rows may be returned, so it queries one day at a time.

R-packages/covidcast/tests/testthat/test-covidcast.R
JedGrabman: Done

R-packages/covidcast/R/covidcast.R
```diff
@@ -549,30 +588,52 @@ covidcast_days <- function(data_source, signal, start_day, end_day, geo_type,
...
+          warn(missing_geo_array$warning,
                data_source = data_source,
                signal = signal,
-               day = query_day,
+               day = ymd(missing_geo_array$time_value),
                geo_value = geo_value,
-               api_msg = dat[[i]]$message,
-               class = "covidcast_missing_geo_values"
-               )
+               msg = dat[[i]]$message,
```
capnrefsmmat, line 625: This needs to go back to being called `api_msg`; maybe this changed back when you rebased on #252?

```suggestion
               api_msg = dat[[i]]$message,
```

R-packages/covidcast/DESCRIPTION
```diff
@@ -43,6 +43,7 @@ Imports:
   grDevices,
   httr,
   jsonlite,
+  lubridate,
```
capnrefsmmat, line 46: lubridate is only used in the tests, right? I think that means [it can go in Suggests instead of Imports](https://r-pkgs.org/description.html#dependencies)

capnrefsmmat (approved):
This looks good, thanks. Let's not merge yet -- I'd like to merge into `r-pkg-devel` instead of main, so let me adjust and get that working

R-packages/covidcast/DESCRIPTION
JedGrabman: `ymd` is a lubricate function which I used a few times in `covidcast.R`

R-packages/covidcast/R/covidcast.R
JedGrabman: I had some merge conflicts early on, seems likely I grabbed the wrong line there. Fixed.

capnrefsmmat:
Well, that was satisfying to test out:

```
> foo = covidcast_signal("fb-survey", "smoothed_whh_cmnty_cli", start_day="2020-10-01", end_day="2020-11-20", geo_type="state")
Fetched day 2020-10-01 to 2020-11-20: 1, success, num_entries = 2601
```

That would have been 50 separate API calls before, so this is very nice.

PR #278, opened 2020-11-13
Parallelize covidcast_signal()

huisaddison:
Proof of concept -- I noticed that covidcast_signal() can be slow; but the operation is easily parallelizable using a tool like `parallel::mclapply()`.

This PR draft is meant to sketch a suggestion, and if it is a good idea (reasons I can think of for it being a "bad idea": parallelism is finicky enough that maybe we prefer to defer it to an end user...) then I (or somebody else interested) can put some time into building, testing, etc.


huisaddison:
Superseded by batched / async(?) downloading

PR #281, opened 2020-11-14
Rollup of R package changes

capnrefsmmat:
This includes the following changes:

- #267
- The faster `latest_issue` from #266, plus the fixed link to source files in the docs
- An improvement to `_pkgdown.yml` to fix ordering of the articles in the rendered site.

I propose we merge this and leave the further changes from #266 for a subsequent release, in combination with #275.


capnrefsmmat:
Ryan checked the `latest_issue` stuff in #266, and we're anxious to get the fixed vignettes up, so I'm using Executive Privilege and merging.

PR #284, opened 2020-11-15
Document a Git Flow release process

capnrefsmmat:
I propose we follow this process for future covidcast releases. This will require us to work on a branch until we're ready to release, but it affords us the opportunity to beta-test before release.

### DIFF
--- a/R-packages/covidcast/DEVELOP.md
+++ b/R-packages/covidcast/DEVELOP.md
@@ -1,5 +1,15 @@
 # covidcast development guide
 
+We follow a Git Flow model for development. The `main` branch is for the
+"release" version of the package, since changes on `main` are immediately
+available for installation and immediately appear on the documentation website.
+New features are instead developed on the `r-pkg-devel` branch so they can be
+beta-tested before release.
+
+To make a change to the package, pull the latest version of `r-pkg-devel` and
+make a branch starting from there. When you open a pull request, set the base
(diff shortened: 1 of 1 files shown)


chinandrew:
What are your thoughts on using this for the Python package too, even though the actual deploy process there is different and doesn't have the same complications? Main benefits in my mind are 1) consistency and 2) ability to checkout the pypi version easily (slightly quicker than finding the commit sha). I don't feel super strongly either way. Seems fairly unnecessary process wise, but consistency isn't worthless.

capnrefsmmat:
I think the main benefit for the Python package would be the documentation -- the Sphinx site is deployed instantly when the HTML is pushed to `main`. We could also just be careful not to commit the HTML until we're ready to release. That's just waiting for a mistake to happen, though/

capnrefsmmat:
@chinandrew Just realized this process will also require us to change the R CI workflow to trigger on PRs to `r-pkg-devel` as well as `main`.

chinandrew:
>the Sphinx site is deployed instantly when the HTML is pushed to `main`

Did not realize that was the case, that seems to be a good reason to use this for Python too.

> @chinandrew Just realized this process will also require us to change the R CI workflow to trigger on PRs to `r-pkg-devel` as well as `main`.

Yup, that shouldn't be difficult to setup.

sgsmob (approved):

PR #289, opened 2020-11-16
[DRAFT] update R CI

chinandrew:
reopening now that I pulled the right commits

### DIFF
--- a/.github/workflows/r_ci.yml
+++ b/.github/workflows/r_ci.yml
@@ -10,9 +10,9 @@ name: R
 
 on:
   push:
-    branches: [ main ]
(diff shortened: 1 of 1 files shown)


chinandrew:
@capnrefsmmat check is still failing here, and I just pulled from r-pkg-devel. Anything I'm missing?

capnrefsmmat:
aaargh

I have R > 4.0, where `stringsAsFactors = FALSE` by default. CI has R 3.5, where `stringsAsFactors = TRUE` by default. So the data frames in the tests have factors, and the code produces strings, and... It should be easy to fix the options to disable `stringsAsFactors` in tests.

I have a PR I'll open momentarily

PR #290, opened 2020-11-16
Disable stringsAsFactors during unit testing

capnrefsmmat:
See comment. This meant the tests ran on R 4.0 (my machine) but not on 3.5 (CI), and I didn't notice because we need to land #289 to have `r-pkg-devel` be tested.

cc @chinandrew


chinandrew:
Just for context, CI is 3.5 since the DESCRIPTION file in covidcast has `Depends: R (>= 3.5.0)`. We can easily add more versions or adjust versions if needed.

capnrefsmmat:
Maybe we stick with 3.5 for now, and after we figure out why the dependency cache doesn't always work, we add 4.0 as well?

(Ubuntu 18.04 LTS is actually on 3.4, which we don't support. We probably don't want to bump the minimum past 3.5 for a while, since it's probably still in wide use on slightly newer distros.)

chinandrew:
I'm going to merge my branch into this one so the CI runs with the changes.

chinandrew:
Looks like we're good.

sgsmob (approved):

PR #303, opened 2020-11-18
Merge main into evalcast

sgsmob:
The main branch has seen a lot of development in parallel to evalcast.  Rather than waiting until evalcast is ready to launch to reconcile these branches, this PR merges main into evalcast now.


capnrefsmmat (approved):
Not sure I know what to review here, since I don't work with evalcast, but it looks like there's only one tiny change in an evalcast file here, so fine with me

dajmcdon (approved):
Yuk. I don't see anything we should be concerned about.

PR #306, opened 2020-11-18
Fix Python client shapefile packaging

chinandrew:
Summary of changes:
a88

All the shapefiles are in subfolders, e.g. shapefiles/state/* or shapefiles/msa/*, so needed to add additional directory layer


capnrefsmmat (approved):
lgtm

PR #313, opened 2020-11-19
Functions for printing prediction and score cards

### LINKED ISSUE #222: Print methods for cards

sgsmob:
Add functions that print out the attributes of prediction and score cards, in addition to their contents.

Fixes #222


dajmcdon (approved):

jacobbien:
Running `devtools::check()`, this creates the following warning:

> > checking S3 generic/method consistency ... WARNING
>   print:
>     function(x, ...)
>   print.prediction_card:
>     function(card, ...)
>   
>   print:
>     function(x, ...)
>   print.score_card:
>     function(card, ...)
>   
>   See section ‘Generic functions and methods’ in the ‘Writing R
>   Extensions’ manual. 
> 

Basically, I think it wants the first argument to be named `x`.

jacobbien:
It also complains that the argument `...` is not documented in both `print.prediction_card ` and `print.score_card`

PR #315, opened 2020-11-20
Add PR to county and state choropleth maps. 

sarah-colq:
Also implemented megacounties on state level as opposed to county level, and fixed includes for hrr, msa to represent that includes is state abbreviations.


sarah-colq:
There's a test error which I missed; I will fix this tomorrow, ignore this PR for now.

PR #316, opened 2020-11-20
Make covidhub_evaluation.Rmd more efficient when previous data exists

JedGrabman:
Description:
Checks for previously saved data before requesting / processing new data, leading to substantial speed improvements for weekly updates.

Change log:
- Checks for previous `predictions_cards` and `score_cards` before requesting or processing data.
- Only save `predictions_cards` and `score_cards`, not all values.
- Updated code for determining which dates to request (includes only desired dates, doesn't use NA).

### DIFF
--- a/R-notebooks/covidhub_evaluation.Rmd
+++ b/R-notebooks/covidhub_evaluation.Rmd
@@ -28,7 +28,6 @@ submissions to make useful comparisons. To do this, we can expose the function
 `get_covidhub_forecaster_names()`.
 
 ```{r}
-load(file = "covidhub_evaluation.rda")
 forecasters = c("CMU-TimeSeries", 
                 "YYG-ParamSearch", 
                 "UMass-MechBayes", 
@@ -57,19 +56,21 @@ for (i in 1:length(forecasters)) {
 }
 ```
 
-Now figure out "comparable" forecast dates: making a forecast on a Sunday or a 
-Monday of the same epiweek should be comparable. **TODO**: we should switch over 
-to using the Zoltar API, and soon it should have an "as of" parameter, so then 
-we shouldn't need to do this. 
-
+Load data from previous run so we don't have to reingest / process it. This
+data could end up out of date if a forecast is retrospectively updated, but in
+that case it's no longer a true prediction. We can always restart from scratch
+by deleting covidhub_evaluation.rda.
 ```{r}
+load(file = "covidhub_evaluation.rda")
 if (exists("predictions_cards")){
   seen_dates = list()
   for (i in 1:length(predictions_cards)){
     forecaster_predictions_cards = predictions_cards[[i]]
-    pred_dates = lapply(forecaster_predictions_cards, FUN = function(x) attributes(x)$forecast_date)
+    pred_dates = lapply(forecaster_predictions_cards, 
+                        FUN = function(x) attributes(x)$forecast_date)
     seen_dates[[i]] = unique(do.call("c", pred_dates))
-    name_of_forecaster = attributes(forecaster_predictions_cards[[1]])$name_of_forecaster
+    name_of_forecaster = attributes(
+                          forecaster_predictions_cards[[1]])$name_of_forecaster
     names(seen_dates)[[i]] = name_of_forecaster
   }
 } else {
@@ -78,14 +79,27 @@ if (exists("predictions_cards")){
   }
   names(seen_dates) = forecasters
 }
+```
+
+Now figure out "comparable" forecast dates: making a forecast on a Sunday or a 
+Monday of the same epiweek should be comparable. **TODO**: we should switch over 
+to using the Zoltar API, and soon it should have an "as of" parameter, so then 
+we shouldn't need to do this. 
+
+```{r}
 
 forecast_dates_cmu = forecast_dates[[1]]
 new_dates = list()
 for (i in 1:length(forecasters)) {
   given_dates = forecast_dates[[i]]
+  # If the dates match exactly, or the given date falls on a Sunday and the
+  # CMU date falls on a Monday of the same epiweek, then call it comparable...
   dates_comp = given_dates[(given_dates %in% forecast_dates_cmu | 
                             ((given_dates + 1) %in% forecast_dates_cmu) &
                             wday(given_dates) == 1)]
+  
+  # ...but if there is an exact match on dates, ignore predictions made on the
+  # previous day
   extra_dates = dates_comp[(dates_comp + 1) %in% dates_comp]
   dates_comp = as_date(setdiff(dates_comp, extra_dates))
   if(forecasters[[i]] %in% names(seen_dates)){
@@ -114,9 +128,11 @@ for (i in 1:length(forecasters)) {
 }
 old_forecaster_idxs = match(forecasters, names(seen_dates))
 
+# Combine old and new predictions cards
 predictions_cards_comb = vector("list", length = length(forecasters))
 for (i in 1:length(forecasters)){
   if (is.na(old_forecaster_idxs[i])){
+    #as.list needed to handle possible NULL values
     predictions_cards_comb[[i]] = as.list(predictions_cards_new[[i]])
   } else {
     predictions_cards_comb[[i]] = as.list(c(predictions_cards_new[[i]], 
@@ -159,7 +175,7 @@ Evaluate the predictions based on weighted interval score (WIS), absolute error
 make sure evalcast "fails gracefully" in as many places as possible, and doesn't
 throw an error (which would halt all execution) when it should be instead just
 throwing a warning and moving on. Basically I needed to artifically trim the
-forecast dates to have the latest one by "2020-09-28" so that I wouldn't get
+forecast dates to have the latest one by "2020-10-19" so that I wouldn't get
 errors with the aheads here. This is undesirable obviously ...
 
 Also, another **TODO**: just want to note that this is again very slow. Caching
@@ -168,7 +184,11 @@ same data something like 15 times in a row. If we simply changed the order of
 operations then we wouldn't even need caching at all.
 
 ```{r}
-scored_forecasters = unlist(lapply(score_cards, FUN = function(card) attributes(card[[1]])$name_of_forecaster))
+scored_forecasters = lapply(score_cards, 
+                            FUN = function(card) {
+                              attributes(card[[1]])$name_of_forecaster
+                              })
+scored_forecasters = unlist(scored_forecasters)
 
 ahead = 1:4
 last_date = as.Date("2020-10-19")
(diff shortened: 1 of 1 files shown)


R-notebooks/covidhub_evaluation.Rmd
```diff
@@ -135,49 +175,80 @@ Evaluate the predictions based on weighted interval score (WIS), absolute error
...
+```{r}
+scored_forecasters = lapply(score_cards, 
+                            FUN = function(card) {
+                              attributes(card[[1]])$name_of_forecaster
+                              })
+scored_forecasters = unlist(scored_forecasters)
+
 ahead = 1:4
-last_date = as.Date("2020-09-28")
+last_date = as.Date("2020-10-19")
```
krivard, line 194: Should this be computed dynamically?

R-notebooks/covidhub_evaluation.Rmd
```diff
@@ -49,57 +49,97 @@ forecasters = c("CMU-TimeSeries",
...
-    else forecast_dates_comparable[[i]][j] = NA
+  # If the dates match exactly, or the given date falls on a Sunday and the
+  # CMU date falls on a Monday of the same epiweek, then call it comparable...
+  dates_comp = given_dates[(given_dates %in% forecast_dates_cmu | 
+                            ((given_dates + 1) %in% forecast_dates_cmu) &
+                            wday(given_dates) == 1)]
+  
+  # ...but if there is an exact match on dates, ignore predictions made on the
+  # previous day
+  extra_dates = dates_comp[(dates_comp + 1) %in% dates_comp]
```
sgsmob, line 103: Took me a while to parse through this.  This variable stores all dates except for the last one in each continuous blocks?  I would like to see a more descriptive name, like `previous_{consecutive_}dates_with_forecasts`.

R-notebooks/covidhub_evaluation.Rmd
```diff
@@ -49,57 +49,97 @@ forecasters = c("CMU-TimeSeries",
...
-        (as.Date(forecast_dates_cmu[j]) - as.Date(given_date) == 1 &&
-         wday(forecast_dates_cmu[j]) == 2 && wday(given_date) == 1))) {
-      forecast_dates_comparable[[i]][j] = given_date
-    }
-    
-    # Otherwise call it NA
-    else forecast_dates_comparable[[i]][j] = NA
+  # If the dates match exactly, or the given date falls on a Sunday and the
+  # CMU date falls on a Monday of the same epiweek, then call it comparable...
+  dates_comp = given_dates[(given_dates %in% forecast_dates_cmu |
```
sgsmob, line 97: Would find this easier to read as `comparable_forecast_dates`

R-notebooks/covidhub_evaluation.Rmd
```diff
@@ -49,57 +49,97 @@ forecasters = c("CMU-TimeSeries",
...
 Monday of the same epiweek should be comparable. **TODO**: we should switch over 
 to using the Zoltar API, and soon it should have an "as of" parameter, so then 
 we shouldn't need to do this. 
 
-```{r, eval = FALSE}
-forecast_dates_comparable = vector("list", length = length(forecasters)) 
+```{r}
+
 forecast_dates_cmu = forecast_dates[[1]]
+new_dates = list()
```
sgsmob, line 92: Can you leave a comment explaining exactly what is "new" about the `new_dates`?  They are dates we don't have a comparable prediction for?

R-notebooks/covidhub_evaluation.Rmd
```diff
@@ -49,57 +49,97 @@ forecasters = c("CMU-TimeSeries",
...
+                            ((given_dates + 1) %in% forecast_dates_cmu) &
+                            wday(given_dates) == 1)]
+  
+  # ...but if there is an exact match on dates, ignore predictions made on the
+  # previous day
+  extra_dates = dates_comp[(dates_comp + 1) %in% dates_comp]
+  dates_comp = as_date(setdiff(dates_comp, extra_dates))
+  if(forecasters[[i]] %in% names(seen_dates)){
+    forecaster_idx = which(names(seen_dates) == forecasters[[i]])
+    dates_comp = dates_comp[!(dates_comp %in% seen_dates[[forecaster_idx]])]
```
sgsmob, line 107: Is this equivalent to `dates_comp = as_date(setdiff(dates_comp, seen_dates[[forecaster_idx]]))`?  If so, I think it would be more readable and probably more efficient the way I wrote it.

R-notebooks/covidhub_evaluation.Rmd
```diff
@@ -49,57 +49,97 @@ forecasters = c("CMU-TimeSeries",
...
                              geo_type = "state")
   },
   error = function(e) cat(e$message))
 }
+old_forecaster_idxs = match(forecasters, names(seen_dates))
+
+# Combine old and new predictions cards
+predictions_cards_comb = vector("list", length = length(forecasters))
+for (i in 1:length(forecasters)){
+  if (is.na(old_forecaster_idxs[i])){
```
sgsmob, line 134: Do you want `i %in% old_forecaster_idxs` instead?  My understanding is that `old_forecaster_idxs` is a list of index positions such as `c(1, 3, 7, 9)` corresponding to forecasters already seen.  So `old_forecaster_idxs[i]` asks for the `i`th member of the list rather than whether `i` is in the list.

R-notebooks/covidhub_evaluation.Rmd
```diff
@@ -49,57 +49,97 @@ forecasters = c("CMU-TimeSeries",
...
+    #as.list needed to handle possible NULL values
+    predictions_cards_comb[[i]] = as.list(predictions_cards_new[[i]])
+  } else {
+    predictions_cards_comb[[i]] = as.list(c(predictions_cards_new[[i]], 
+                                    predictions_cards[[old_forecaster_idxs[i]]]))
```
sgsmob, line 139: Same comment about `old_forecaster_idxs[i]`.

sgsmob (changes requested):

R-notebooks/covidhub_evaluation.Rmd
JedGrabman: >This variable stores all dates except for the last one in each continuous blocks

Yes, but the reason is a little bit more subtle.

CMU makes forecasts once a week. We consider a forecast date to be comparable to a CMU forecast if it was made on the same day or the previous day. So, if an organization makes weekly forecasts on a Sunday or Monday, we want to include it.

But, some organizations make forecasts daily and we don't want to include both the Sunday and Monday versions in that case, just the Monday version.

So, it is the last day in the chain, but it is really just excluding the Sunday version if the Monday version is also there.

R-notebooks/covidhub_evaluation.Rmd
JedGrabman: Agreed. That looks equivalent and clearer.

R-notebooks/covidhub_evaluation.Rmd
JedGrabman: The result from `match` is the same length as `forecasters` and specifies which value in the 2nd vector matches the value in the 1st vector (with NA representing no match).  

For example, `match(c('foo', 'bar', 'baz'), c('baz', 'foo'))` would be `c(2, NA, 1)`.

so, `old_forecaster_idxs[i]` being `NA` means there is no match.

R-notebooks/covidhub_evaluation.Rmd
JedGrabman: It should be, but it seemed out of scope for this PR and I'm not 100% on what the value should be (but I *think* I just need to check for 4 weeks back).

R-notebooks/covidhub_evaluation.Rmd
```diff
@@ -49,57 +49,97 @@ forecasters = c("CMU-TimeSeries",
...
+load(file = "covidhub_evaluation.rda")
+if (exists("predictions_cards")){
+  seen_dates = list()
+  for (i in 1:length(predictions_cards)){
+    forecaster_predictions_cards = predictions_cards[[i]]
+    pred_dates = lapply(forecaster_predictions_cards, 
+                        FUN = function(x) attributes(x)$forecast_date)
+    seen_dates[[i]] = unique(do.call("c", pred_dates))
+    name_of_forecaster = attributes(
+                          forecaster_predictions_cards[[1]])$name_of_forecaster
```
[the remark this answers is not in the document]
JedGrabman: `predictions_cards` is a list of lists. Each inner list (i.e. `predictions_cards[[i]]`) is specific to a single forecaster. The assumption is that an empty list will not be saved, and that all inner lists will have at least 1 element, with each element corresponding to the same forecaster. Therefore, checking the 1st element seems appropriate, because we cannot guarantee that there are any additional elements.

R-notebooks/covidhub_evaluation.Rmd
JedGrabman: Dates that we don't have data saved for. They are new in the sense that we will find any new forecasts that were made since the last time we ran the notebook. Technically, it includes old dates as well if they were non-comparable, because we didn't save that data. But, we filter those out when we find comparable dates.

I can add a comment.

R-notebooks/covidhub_evaluation.Rmd
JedGrabman: I will update

R-notebooks/covidhub_evaluation.Rmd
sgsmob: Please do

R-notebooks/covidhub_evaluation.Rmd
```diff
@@ -49,57 +49,97 @@ forecasters = c("CMU-TimeSeries",
...
+                            ((given_dates + 1) %in% forecast_dates_cmu) &
+                            wday(given_dates) == 1)]
+  
+  # ...but if there is an exact match on dates, ignore predictions made on the
+  # previous day
+  extra_dates = comparable_forecast_dates[(comparable_forecast_dates + 1) %in% comparable_forecast_dates]
+  comparable_forecast_dates = as_date(setdiff(comparable_forecast_dates, extra_dates))
+  if(forecasters[[i]] %in% names(seen_dates)){
+    forecaster_idx = which(names(seen_dates) == forecasters[[i]])
+    comparable_forecast_dates = comparable_forecast_dates[!(comparable_forecast_dates %in% seen_dates[[forecaster_idx]])]
```
sgsmob, line 107: can be clearer as `as_date(setdiff(comparable_forecast_dates, seen_dates[[forecaster_idx]]))`

sgsmob (approved):
After addressing the few remaining comments, should be good to go.

R-notebooks/covidhub_evaluation.Rmd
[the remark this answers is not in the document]
sgsmob: Oops nevermind, I misread this. It is ok as is.

R-notebooks/covidhub_evaluation.Rmd
JedGrabman: Sorry, forgot to push before. Updated.

JedGrabman:
@krivard I think this is good to go (pending your approval).

PR #321, opened 2020-11-23
R custom API endpoint

### LINKED ISSUE #317: Expose access to the API endpoint used for R covidcast client queries
It is possible to direct the Python covidcast client at an alternative API endpoint like so:

```
import covidcast
covidcast.covidcast.Epidata.BASE_URL = "https://[something_else]/epidata/api.php"
```

However R places tighter controls on library innards. The closest I could get was the following, which errors out:

```
> library(covidcast)
> environment(covidcast_signal)$COVIDCAST_BASE_URL <- "https://[something-else]/epidata/api.php"
Error in environment(covidcast_signal)$COVIDCAST_BASE_URL <- "https://[something-else]/epidata/api.php" : 
  cannot change value of locked binding for 'COVIDCAST_BASE_URL'
```

For purposes of giving folks advanced access to prototype signals that are not yet available from the public API, it would be nice if the R client could accept an alternative API endpoint address.

capnrefsmmat:
Fixes #317.

With this, you should be able to use `options(covidcast.base_url = "https://some-staging-server/...")` to have the package use a different endpoint.

### DIFF
--- a/R-packages/covidcast/R/covidcast.R
+++ b/R-packages/covidcast/R/covidcast.R
@@ -667,9 +667,10 @@ covidcast <- function(data_source, signal, time_type, geo_type, time_values,
 
 # Helper function to request and parse epidata
 .request <- function(params) {
-  # API call
-  response <- httr::GET(COVIDCAST_BASE_URL, httr::user_agent("covidcastR"),
-                        query=params)
+  # API call. Allow base API URL to be replaced, e.g. to use a staging/testing
+  # server when needed.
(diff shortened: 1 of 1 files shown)


sgsmob (commented):
Is this documented anywhere?  Should it be?

Code change looks fine.

capnrefsmmat:
I think the idea is that this will be used by the engineering team while they're testing out new indicators on a staging API server, but that there's no reason for ordinary users to be changing endpoints. Hence the lack of public documentation.

sgsmob (approved):

PR #323, opened 2020-11-23
Migrate `plot_trajectory()` onto `aggregate_cards()`

sgsmob:
The functionality of `aggregate_cards()` was modeled after a subset of `plot_trajectory()`.  To avoid code reuse, we replace the duplicated code in `plot_trajectory()` with a call to `aggregate_cards()`.

This PR also fixes a bug in `plot_trajectory()` where the `forecast_date` column was unavailable and cleans up some readability issues.

### DIFF
--- a/R-packages/evalcast/R/plot_trajectory.R
+++ b/R-packages/evalcast/R/plot_trajectory.R
@@ -204,12 +204,11 @@ get_target_period_num <- function(forecast_date,ahead,incidence_period){
   )
   if(incidence_period == "day")
   {
-    return(forecast_date + ahead)
+    forecast_date + ahead
   } else {
-    x = ifelse(wday(forecast_date) <= 2, 
+    ifelse(wday(forecast_date) <= 2, 
            MMWRweek::MMWRweek(forecast_date)$MMWRweek + ahead - 1,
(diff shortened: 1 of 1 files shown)


dajmcdon (approved):

jacobbien:
When I run the following, I get the error `Error: Invalid index: out of bounds`:
```
signals <- tibble::tibble(data_source = "jhu-csse",
                  signal = c("deaths_incidence_num", "confirmed_incidence_num"),
                  start_day = "2020-06-15")
forecast_dates <- lubridate::ymd(c("2020-07-20", "2020-08-10"))
pc <- get_predictions(baseline_forecaster,
                        name_of_forecaster = "baseline",
                        signals = signals,
                        forecast_dates = forecast_dates,
                        incidence_period = "epiweek",
                        ahead = 3,
                        geo_type = "state")
aggregate_cards(pc)
```
Is this not an appropriate use of `aggregate_cards` or is this a bug?  If the former, it'd be nice if the error message could be more descriptive.

sgsmob:
I can't replicate this issue because I don't have any object named `baseline_forecaster`.  How are you loading it?

jacobbien:
The function `baseline_forecaster` is exported by evalcast, so this should be replicable once you run `library(evalcast)`.

PR #326, opened 2020-11-24
Update R plotting metadata check

### LINKED ISSUE #270: Signals without metadata fail to plot

chinandrew:
Fixes #270 

Summary of changes
- Update the metadata check in the plotting function to look specifically at mean and stdev values
- Add test that the warning is raised if one of the values is missing and not raised if both are present.

### DIFF
--- a/R-packages/covidcast/tests/testthat/test-plot.R
+++ b/R-packages/covidcast/tests/testthat/test-plot.R
@@ -133,9 +133,7 @@ test_that("warn on incomplete metadata", {
   
   expect_warning(plot(fake_data), NA) 
   attributes(fake_data)$metadata = list(geo_type = "state", mean_value = 0)
-  
+
   expect_warning(plot(fake_data),
-                 "Metadata for signal mean and standard deviation not",
-                 "available; defaulting to observed mean and standard",
-                 "deviation to set plot range.")
+                 class="covidcast_plot_meta_not_found")
 })
--- a/R-packages/covidcast/tests/testthat/test-plot.R
+++ b/R-packages/covidcast/tests/testthat/test-plot.R
@@ -116,3 +116,23 @@ test_that("simple county bubble plot", {
                       suppressWarnings(
                         plot(fb_county, plot_type = "bubble")))
 })
+
+
+test_that("test incomplete metadata", {
+  fake_data <- structure(data.frame(
+    data_source = "foo",
+    signal = "bar",
+    value = c(1, 2, 0, 3),
+    geo_value = c("pa", "in", "tx", "wy"),
+    time_value = as.Date("2020-01-01"),
+    issue = as.Date("2020-02-01"),
+    stderr = 0.5),
+    class = c("covidcast_signal", "data.frame"),
+    metadata = list(geo_type = "state")
+  )
+  
+  expect_warning(plot(fake_data),
+                 "Metadata for signal mean and standard deviation not",
+                 "available; defaulting to observed mean and standard",
+                 "deviation to set plot range.")
+})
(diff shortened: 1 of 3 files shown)


sgsmob (approved):

R-packages/covidcast/tests/testthat/test-plot.R
```diff
@@ -116,3 +116,26 @@ test_that("simple county bubble plot", {
...
+    metadata = list(geo_type = "state", mean_value = 0, stdev_value = 1)
+  )
+  
+  expect_warning(plot(fake_data), NA) 
+  attributes(fake_data)$metadata = list(geo_type = "state", mean_value = 0)
+  
+  expect_warning(plot(fake_data),
+                 "Metadata for signal mean and standard deviation not",
+                 "available; defaulting to observed mean and standard",
+                 "deviation to set plot range.")
```
capnrefsmmat, line 140: Having a warning class means we don't have to match the message:

```suggestion
                 class = "covidcast_plot_meta_not_found")
```

R-packages/covidcast/tests/testthat/test-plot.R
```diff
@@ -116,3 +116,26 @@ test_that("simple county bubble plot", {
                       suppressWarnings(
                         plot(fb_county, plot_type = "bubble")))
 })
+
+
+test_that("test incomplete metadata", {
```
capnrefsmmat, line 121:
```suggestion
test_that("warn on incomplete metadata", {
```

capnrefsmmat (commented):
Thanks; a couple suggested improvements

R-packages/covidcast/tests/testthat/test-plot.R
chinandrew: So I tried this, but couldn't get it to work since this passes (as expected)
```
  expect_warning(plot(fake_data), class="covidcast_plot_meta_not_found")
```
but this also passes
```
  expect_warning(plot(fake_data), class="this_should_fail")
```

capnrefsmmat: wat

sigh:

```
> expect_warning(warn("ducks", class="duck_warning"), class="boing")
> expect_error(abort("ducks", class="duck_error"), class="boing")
Error: `abort("ducks", class = "duck_error")` threw an error with unexpected class.
Expected class: boing
Actual class:   duck_error/rlang_error/error/condition
Message:        ducks
```

So I guess the documentation is a lie for warnings. We might be able to fix this by opting in to [testthat's third edition](https://www.tidyverse.org/blog/2020/10/testthat-3-0-0/#3rd-edition), since that unifies the handling for all three. Worth trying?

chinandrew: Seems to work
```
> library(testthat)
> local_edition(3)
> expect_warning(warn("ducks", class="duck_warning"), class="boing")
Error: `warn("ducks", class = "duck_warning")` did not throw the expected warning.
In addition: Warning message:
ducks 
```
Updated code to use 3rd ed

PR #329, opened 2020-11-25
Add additional aggregation tests to R covidcast package

chinandrew:
Partially addresses #239

Summary of changes:
- add three tests:
    a. test that you can lag a list of signals with a list of lags
    b. test that you can lag a list of signals with a single lag vector applied to all elements
    c. test that a full join occurs when aggregating two signals with different date ranges


capnrefsmmat (approved):
looks good

PR #330, opened 2020-11-25
fix failing tests, should build now

dajmcdon:


jacobbien:
The changes on this PR look fine to me.  However, when I run R CMD check I get a lot of messages.  These don't appear to be associated with this particular PR but rather they all seem to stem from `plot_trajectory`, so I will make note of that somewhere else.

dajmcdon:
Yeah, that function needs some cleanup. The issue here was that the documentation was missing an `itemize` environment, so when `aardvardk` tried to install, it couldn't install the `evalcast` dependency. This was the smallest possible change to make it work.

Issue #100, opened 2020-10-02
not hard-coding covidhub_probs


jacobbien:
Currently, the function `check_valid_forecaster_output` (in [evaluate.R](https://github.com/cmu-delphi/covidcast/blob/evalcast/R-packages/evalcast/R/evaluate.R)) makes sure that the forecaster has outputted all 23 quantile levels, hard-coded as `covidhub_probs`:

`covidhub_probs <- c(0.01, 0.025, seq(0.05, 0.95, by = 0.05), 0.975, 0.99)`

However, this is overly restrictive since, as Vishnu pointed out, [the COVID Hub submission instructions](https://github.com/reichlab/covid19-forecast-hub/blob/master/data-processed/README.md) specify that for the task of "N wk ahead inc case", only 6 quantile levels need to be submitted.

Some options:

* simplest approach: remove this check or make it a warning rather than an error
* enforce (based on the forecasting task) one of these two COVID Hub sets of probs
* don't enforce any particular choice but make sure the set of quantile levels reported is consistent across all locations and aheads for a forecaster?

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970447730609) by [Unito](https://www.unito.io/learn-more)

dajmcdon:
I'm attempting to go with option (c) "don't enforce any particular choice but make sure the set of quantile levels reported is consistent across all locations"

Issue #132, opened 2020-10-21
Allow for a non-quantile forecast in a predictions card


jacobbien:
A predictions card has two columns: `probs` and `quantiles` (see e.g. [this slide](https://cmu-delphi.github.io/covidcast/talks/evalcast/talk.html#(7))).  However, sometimes a forecaster may produce a point forecast.  (For example, on the COVID Hub they allow forecasters to submit point forecasts in addition to quantile forecasts.). This could be useful for hotspot detection where one might simply want to predict a 1 or a 0.

To accommodate this, I would suggest that we allow predictions cards to have a row where `probs` is `NA`.  For point forecasters (including hotspot detection), the predictions card could simply be a 1-by-2 tibble with `NA` in the `probs` column and the value of the point forecast in the `quantiles` column.  More generally, one could allow a quantile forecaster to also have an additional row with a point forecast.

The reason to code this as an `NA` would be so that the `probs` column would remain numeric.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198968030437421/1198970302781344) by [Unito](https://www.unito.io/learn-more)

nickreich:
Wondering if you all have looked in detail at the data structure at zoltardata.com that hosts the Hub forecasts and whether (a) you could download data from the API from there directly via our R or python package or (b) you could design your system to more directly interface with these general forecast data types?

jacobbien:
Thanks Nick.  In issue #99 I had made a note that we should be using zoltr rather than scraping -- definitely the better thing to do!  In terms of accommodating more general forecast data types, this is not something I had considered.  I was able to find [this page](https://docs.zoltardata.com/targets/).  I assume this is what you are referring to?

nickreich:
Yes that is the right reference point. Probably overkill to engineer evalcast to deal with anything other than quantile/point forecasts (although CDC has used bin forecasts in the past as well), but just wanted to make sure that this general conceptual  framework for forecasts was on your radar as you move forward with design.

PR #333, opened 2020-11-27
Correct logic in check_valid_forecaster_output

e3bo:
I was wondering why I was able to evaluate prediction cards for incident case forecasts in spite of them not using the set of probabilities that is hardcoded as a reference as mentioned in #100. Looking into it, I think I may have found a bug in the logic which checks the probabilities and quantiles. I'm opening this pull request just in case I'm correct and you are not aware of this problem. This change does seem to make #100 more urgent though.

### DIFF
--- a/R-packages/evalcast/R/evaluate.R
+++ b/R-packages/evalcast/R/evaluate.R
@@ -173,9 +173,9 @@ check_valid_forecaster_output <- function(pred_card) {
   wrong_format <- pred_card$forecast_distribution %>%
     map_lgl(~ any(names(.x) != c("probs", "quantiles")))
   wrong_probs <- pred_card$forecast_distribution %>%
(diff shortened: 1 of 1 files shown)


dajmcdon:
Hi @e3bo. I'd noticed that problem when I tried to address #132 on [cmu-delphi:evalcast-pointfcasts](https://github.com/cmu-delphi/covidcast/tree/evalcast-pointfcasts) and fixed it there, but haven't merged it yet. That process led to me making wholesale changes, and I may well have broken other things at the same time. It's not clear to me that this check is necessary or desirable. More appropriate perhaps, would be to check that all locations have the same set of quantiles (rather than the covidhub quantiles) which was my attempted fix. But I haven't tested the code my yet.

e3bo:
Sounds good. I can see the package is developing quickly, but I'm already finding it easier to use than creating my own evaluation framework.

PR #334, opened 2020-11-29
Modeltools

### LINKED ISSUE #131: Derivative functionality

ryantibs:
This is new package I'm building called modeltools, as a companion to covidcast and evalcast.  Its purpose is to providing utilities for aiding modeling: mainly forecasting and hotspot prediction.

This draft PR currently has functionality for:

- "sliding" a function on the values over time (making it convenient to do things like a 7-day average or anything else of that nature);
- calculating the percentage change of the values from one week to the next (or over any arbitrary time periods);
- estimating derivatives, currently with three ways: linear regression, smoothing spline, and trend filtering.

The third item here should close #131.  More dev to come on this branch.


R-packages/modeltools/R/modeltools.R
```diff
@@ -0,0 +1,9 @@
+#' modeltools: Tools for Building COVID Forecasters and Related Models 
+#'
+#' The modeltools package provides various tools for building COVID-19 models,
+#' mainly forecasting and hotspot prediction models. It is intended to be used
+#' in conjuction with the covidcast and evalcast packages.
+#'
+#' @docType package
+#' @name modetools
```
huisaddison, line 8:
```suggestion
#' @name modeltools
```

jacobbien:
Nice.  Just a thought that it might be nice to name this [something]cast to go with the other two package names.  For example,  `helpercast` or `utilscast`?

R-packages/modeltools/DESCRIPTION
```diff
@@ -0,0 +1,26 @@
...
+    covidcast and evalcast packages.
+Encoding: UTF-8
+License: MIT
+Imports: 
+    covidcast,
+    dplyr,
+		evalcast,
+    genlasso,
+    slider,
+    tidyr
```
huisaddison, line 25:
```suggestion
    tidyr
Remotes:
    local::../covidcast,
    local::../evalcast
```

Automatic dependency resolution (if installing `modeltools` without, e.g., `evalcast` already installed (cf. https://cran.r-project.org/web/packages/remotes/vignettes/dependencies.html).

R-packages/modeltools/vignettes/estimate-deriv.Rmd
```diff
@@ -0,0 +1,139 @@
...
+## Estimating derivatives via smoothing spline
+
+TODO, blah try smoothing splines two ways, plot them
+
+```{r, fig.width = 8, fig.height = 4}
+case_rates <- estimate_deriv(case_rates, method = "ss", n = 28,
+                             new_col = "deriv_ss1", df = 8)
+
+case_rates <- estimate_deriv(case_rates, method = "ss", n = 28,
+                             new_col = "deriv_ss2", cv = TRUE)
```
huisaddison, line 84:
```suggestion
                             col_name = "deriv_ss2", cv = TRUE)

File: R-packages/modeltools/vignettes/estimate-deriv.Rmd [hunk 4]
```diff
@@ -0,0 +1,139 @@
...
+grid.arrange(p1, p2, nrow = 1)
+```
+
+## Estimating derivatives via smoothing spline
+
+TODO, blah try smoothing splines two ways, plot them
+
+```{r, fig.width = 8, fig.height = 4}
+case_rates <- estimate_deriv(case_rates, method = "ss", n = 28,
+                             new_col = "deriv_ss1", df = 8)
```
huisaddison, line 81:
```suggestion
                             col_name = "deriv_ss1", df = 8)
```

(Same for rest of vignettes).

R-packages/modeltools/vignettes/estimate-deriv.Rmd
```diff
@@ -0,0 +1,139 @@
...
+
+grid.arrange(p1, p2, nrow = 1)
+```
+
+## Estimating derivatives via smoothing spline
+
+TODO, blah try smoothing splines two ways, plot them
+
+```{r, fig.width = 8, fig.height = 4}
+case_rates <- estimate_deriv(case_rates, method = "ss", n = 28,
```
huisaddison, line 80: This line generates many, many
```
spar-finding: non-finite value inf; using BIG value
```
despite the fact that the `smooth.spline()` call is wrapped in `tryCatch(suppressMessages(suppressWarnings(...)))`.  I can see if a tweak is possible to make this more user-friendly (or it could be good that those messages are passing through to the user?)

R-packages/modeltools/DESCRIPTION
ryantibs: Thanks!  Good catch.  I'll stick to the fix used in the evalcast DESCRIPTION to be consistent (which is very similar and likely equivalent).

R-packages/modeltools/vignettes/estimate-deriv.Rmd
ryantibs: Good catch, you spotted this before I pushed my latest commits (which resolve these inconsistencies).

R-packages/modeltools/vignettes/estimate-deriv.Rmd
ryantibs: Same as above.

R-packages/modeltools/vignettes/estimate-deriv.Rmd
ryantibs: Yes this is annoying.  I couldn't, in fact, find a clean way to trap this.  It appears to come from some kind of printf statement from C++, and googling around suggests some hacks but they're very ugly.  If you see a cleanish way, I'd be happy to consider it.

ryantibs:
Thanks @jacobbien.  I too thought it would be nice to have a package name ending in "cast", but I couldn't think of one that I liked.  I would want the package name to readily identify what the package is for.  Hence "modeltools".  I don't really like either suggestion ("utilscast" or "helpercast") because they don't really satisfy this in my opinion.

Also, thanks @huisaddison for the review!  Just a heads up that I added a bunch more functionality just now.  For convenience, let me point you to the compiled vignettes: 
- [getting started](https://htmlpreview.github.io/?https://github.com/cmu-delphi/covidcast/blob/modeltools/R-packages/modeltools/vignettes/modeltools.html)
- [percentage change](https://htmlpreview.github.io/?https://github.com/cmu-delphi/covidcast/blob/modeltools/R-packages/modeltools/vignettes/pct-change.html)
- [derivative estimation](https://htmlpreview.github.io/?https://github.com/cmu-delphi/covidcast/blob/modeltools/R-packages/modeltools/vignettes/estimate-deriv.html)

(The last two have all the code examples in there but the text isn’t fleshed out.). You might find the examples I give near the end of the getting started vignette convenient for implementing and testing out sensorization.  cc @rumackaaron @jingjtang as well.  

If I get around to it, I hope to add some more functionality in the next couple days for forecasting runs.

R-packages/modeltools/R/deriv.R
```diff
@@ -0,0 +1,205 @@
...
+  keep_obj = params$keep_obj
+  deriv = params$deriv
+  params$keep_obj = NULL
+  params$deriv = NULL
+  x = x %>% select(time_value, value) %>% drop_na()
+  params$x = as.numeric(x$time_value)
+  params$y = x$value
+  
+  return(tryCatch(suppressWarnings(suppressMessages({
+    object = do.call(smooth.spline, params)
```
huisaddison, line 152:
```suggestion
    capture.output(object <- do.call(smooth.spline, params))
```
Diverts anything written by `smooth.spline` away from stdout.

R-packages/modeltools/vignettes/estimate-deriv.Rmd
huisaddison: [This](https://github.com/cmu-delphi/covidcast/pull/334/files#r532963129) worked for me (tested locally), and looks pretty clean to me.  (Reference: https://stackoverflow.com/questions/8797314/suppress-messages-displayed-by-print-instead-of-message-or-warning-in-r)

huisaddison:
> (The last two have all the code examples in there but the text isn’t fleshed out.). You might find the examples I give near the end of the getting started vignette convenient for implementing and testing out sensorization. cc @rumackaaron @jingjtang as well.

Thanks!  This provides a clean "canonical" way for end users to perform sensorization and may be a good way for @jingjtang to apply sensorization to Google Symptoms.

My own sensorization code is parallelized (so that we can do multiple sensors and time windows in a reasonable amount of time);   after I clean up my own code, I might offer it (internally in Delphi, if not as a PR against this package) as a "canonical" way to perform large-scale sensorization experiments (if there is demand for that).

R-packages/modeltools/vignettes/estimate-deriv.Rmd
ryantibs: Awesome, thank you!

ryantibs:
Thanks @huisaddison, if you have ways of speeding it up through parallelization then of course that could be useful add to the package.  It all depends on whether a. the current unparallelized version is obstructively slow for real uses, and b. your parallel solution "fits cleanly" into the current structure (so that its oddly be easy to maintain along with the current package).

A bit of googling suggests tools like `multidplyr` or `dtplyr` could be useful for parallelization on top of `dplyr`.  I'm not sure what it is your sensorization code is currently doing, but the `modeltools` version is just based on `dplyr`'s `group_by()` and `group_modify()` verbs, so maybe it could be conveniently parallelized (across groups) with `multidplyr` or `dtplyr`?  OTOH, the `multidplyr` GitHub suggests that you won't see performance gains until you hit millions of rows.

huisaddison:
Yeah, the reason I floated posting the parallel code idiom, rather than incorporating it into the package, is because in my experience parallelism can be fragile, and it may not be worth it if the typical end user does not sensorize multiple indicators at once (as I'm doing in the DAP).

Regarding `multidplyr` - unfortunately, `group_modify()` is not implemented for `multidplyr` (I filed an [issue](https://github.com/tidyverse/multidplyr/issues/102) three weeks ago), not sure about `dtplyr`.  My parallel implementation uses `group_split()` to obtain a list of dataframes, and then uses `parallel::mclapply()`.

krivard:
(just as a drive-by -- the fb-surveys code uses `parallel::mclapply()`, and we've been happy with it)

ryantibs:
Hi all, I just added a bunch more to this PR.  The big thing I added was function to fit quantile autoregressive models, called `quantgen_forecaster()`.  It's set up in such a way that it can be run through evalcast's `get_predictions()`, and makes it easy to run experiments like then one I did in this [old blog post](https://delphi.cmu.edu/blog/2020/09/21/can-symptoms-surveys-improve-covid-19-forecasts/).  This [new vignette](https://htmlpreview.github.io/?https://github.com/cmu-delphi/covidcast/blob/modeltools/docs/modeltoolsR/articles/quantgen-forecast.html) demonstrates how to use it. 

I'm going to go ahead and merge this PR myself (admin privileges) because the modeltools package is pretty stable at this point, and no commits on this PR touched anything outside of the new modeltools package anyway, so even in the worst case nothing else should be screwed up.

Want to leave several comments here: 
- @dajmcdon @jacobbien @huisaddison @ajgreen93 You guys in particular please take a careful look at the `quantgen_forecaster()` function and [associated vignette](https://htmlpreview.github.io/?https://github.com/cmu-delphi/covidcast/blob/modeltools/docs/modeltoolsR/articles/quantgen-forecast.html).  One of the main use cases for this functionality was to facilitate experiments like those we'd want to run in COVIDcast paper number 2.
-  Apologies, I didn't have time to flesh out that vignette text, but the code should be pretty well-commented.  If anybody wants to flesh out the text and update the vignette as evalcast moves forward with fixes, then that'd be great.
- This vignette took a little while to run (several hours), unfortunately it's not all that zippy at the moment.  You'll want to move down to the HRR level for the paper, which will multiply the cost by about 6 (I did 50 states, and there's about 300 HRRs).  
- Note that this vignette sets `lambda = 0` and hence forms quantile regression.  I'm actually running a version now that uses cross-validation (actually, a forward-in-time version of cross-validation) to select the tuning parameter and it's considerably slow.  It's been running for over 24 hours, I think it'll take at least 10 more.  I'll push the rda file containing the results to the modeltools/vignettes subfolder when it finishes, if you guys want to take a look.
- I unfortunately didn't get around to implementing the hotspot predictor function.  It should follow in a pretty similar way to what I did with the quantgen forecaster, just using (say) glmnet to fit the sparse logistic regression instead of quantgen.  There will be a bit more to do because you won't be operating on the raw response signal, but a binarized version (transformed into a 0-1 hotspot label).  I should create an issue for this but just running out of time now, please somebody do that so we can keep track.  Hopefully then somebody can pick this up and implement, say, `glmnet_hotspot_predictor()` in the modeltools package, which we can then use for the hotspot experiments in our paper.
- This last note is actually for @JedGrabman @kateharwood, but also Dan and Jacob.  In the vignette, I wrote a function called `canonical_plot()` to be a kind of "one-function-to-produce-all" in terms of canonical plots for forecast evaluation.  I was saying to Jed/Kate on slack a few days ago that I think one nice way to alleviate a lot of the "plot-overload" feel in our evaluation notebook would be to have a shiny/plotly app that allows you to toggle between different options.  Mean or median.  Unscaled or scaled.  Scale before aggregating or after, etc.  This function allows you to do all that (and much more).  You can see all the plots I produce in the vignette from it.  So while this function could serve as the backend, you'd still need to investigate building a shiny/plotly app around it.  And such a tool might even be the way to go to power the forecast dashboard, so we don't have to rewrite code in D3/JavaScript.

ryantibs:
Oh and one last thing.  I would love it if somebody wanted to be the maintainer for this package.  I only put myself down because R requires you to have a maintainer in the DESCRIPTION.

ryantibs:
Heads up that you don't have to read these comments anymore, I consolidated everything in the new PR #398.

Issue #332, opened 2020-11-26
Fixes to plot_trajectory()


jacobbien:
1. What is the variable `ocation` in the following line of code?
```
      dplyr::select(ocation, time_value, value) %>%
```

2. Running `devtools::check()` reveals a few things:

- a warning related to the package `ggforce` being used without the package being declared as an import.
- Undocumented arguments in documentation object `plot_trajectory`
`nrow` `ncol` `page_facet` `cutoff`
- A large number of complaints about `no visible binding for global variable` ... which we have addressed elsewhere in the package by using the prefix `.data$`.

PR #336, opened 2020-11-29
Evalcast hotfix

dajmcdon:
fix some of the `plot_trajectory()` problems in #332 .

### DIFF
--- a/R-packages/evalcast/DESCRIPTION
+++ b/R-packages/evalcast/DESCRIPTION
@@ -61,7 +61,8 @@ Imports:
     rvest,
     xml2,
     ggplot2,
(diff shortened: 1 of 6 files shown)


dajmcdon:
I'm ignoring the pile of notes due to the missing `.data` prefixes for the moment. It at least is no longer broken. I'm in the process of fixing up a pile of things, and this function is due for an overhaul/simplification.

jacobbien:
My previous commit addresses two more warnings.

jacobbien (approved):

dajmcdon:
Oh, thanks Jacob. I forgot that crud.

PR #349, opened 2020-12-02
Factor out common dashboard code and add tabs

benjaminysmith:
This addresses a few requests in #32, specifically:
- It adds tabs for counties/metros/samples (or states instead of counties where we don't have counties available)
- It adds hour/minute/seconds to generation time
- It extracts more common functions to the utility file instead of duplicating
- It tweaks the graph titles and depends less on (duplicated) text in the markdown 
- It moves the source and signal to params

This does not (yet) update the facebook dashboard since that is using a slightly different structure and contains a correlation graph. I'll do that in a separate PR.


R-notebooks/safegraph_dashboard.Rmd
```diff
@@ -1,67 +1,50 @@
...
 title: "Safegraph dashboard"
 author: "Delphi Lab"
-date: "`r format(Sys.time(), '%B %d, %Y')`"
+date: "`r format(Sys.time(), '%B %d, %Y - %H:%M:%S')`"
 output:
   html_document:
     code_folding: hide
+params:
+  source: "safegraph"
+  signal: "bars_visit_num"
```
krivard, line 10: `bars_visit_num` is likely to have different behavior from `full_time_work_prop`, since they come from different data pipelines. Is there a way we can make these dashboards support multiple signals? Otherwise we'll want to split up sources that combine multiple pipelines so that we have one dash per pipeline.

benjaminysmith: Agreed. Reverted back to median_home_dwell_time for now. I'll add the new signals in a separate PR.

krivard (approved):
xlnt

PR #353, opened 2020-12-03
Add tabs to FB dashboard

benjaminysmith:
Adds tabs to FB dashboard (per #32) by:

- Removing correlations from FB dash (this is in the correlations dash already)
- Splitting into county/msa/sample sections
- Adding MSA code

Note that this does not yet use the common dashboard functions. We will do this in a followup since this dashboard uses > 28 day history.


krivard (approved):
lgtm

PR #354, opened 2020-12-03
interval_coverage update

JedGrabman:
Fixes a bug in `interval_coverage` (`which.min` was applied to a logical vector)
Change parameter to be coverage proportion instead of alpha = (1 - coverage)
Adds tests
Minor readability changes


dajmcdon (approved):
good catch on the error, and thanks for the tests.

PR #337, opened 2020-11-29
Add PR to county and state view.

sarah-colq:
Add Puerto Rico to state and county choropleth maps by using shapefiles.  Remove centroids and change order of transformations to properly reflect what happens in the Python version.  Use megacounties properly for counties by making them the entire state as a transparent background.  Update ubuntu version to 20.04 because it has a more updated version of gdal and projection.

PR #356, opened 2020-12-04
[Draft] Add condition to skip problematic tests on version 3.5

### LINKED ISSUE #350: Add R 3.5 check for covidcast that skips problematic tests
Some of the vdiffr plotting tests don't work the same on 3.5 due to how the SVG is encoded, but the output image is visually the same. We're changing the hCI to run on 3.6, but since the package is 3.5+, this task is to add a 3.5 CI workflow and a condition to the tests so they only run on 3.6+

cc @sarah-colq @capnrefsmmat

chinandrew:
R 3.5 has a slightly different SVG encoding that yields visually identical plots but they fail vdiffr (see [this slack thread](https://delphi-org.slack.com/archives/C01E5M9KVN1/p1606775799011500) for more info). In #337 @sarah-colq is updating the CI to run on R 3.6 so tests pass, and this closes #350 which was opened as a follow up ticket.

### DIFF
--- a/.github/workflows/r_ci.yml
+++ b/.github/workflows/r_ci.yml
@@ -22,7 +22,7 @@ jobs:
         working-directory: R-packages/covidcast/
     strategy:
       matrix:
-        r-version: [3.5]
+        r-version: [3.5, 3.6]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


dshemetov:
Not testing 3.5 in CI anymore

PR #361, opened 2020-12-04
Add axes argument and return axes in plots

### LINKED ISSUE #359: Add ax argument to plotting function
This would make it simple to create multiple subplots of maps

chinandrew:
Closes #359 

Summary of changes:
- Add an optional axes argument so now you can use subplots more easily, e.g.
```
    fig, ax = plt.subplots(1,2)
    covidcast.plot(df2, ax = ax[0])
    covidcast.plot(df1, ax = ax[1])
```
- Also have the functions return the axis instead of the figure since the axis may not be tied to one.

### DIFF
--- a/Python-packages/covidcast-py/covidcast/plotting.py
+++ b/Python-packages/covidcast-py/covidcast/plotting.py
@@ -43,7 +43,8 @@ def plot(data: pd.DataFrame,
          time_value: date = None,
          plot_type: str = "choropleth",
          combine_megacounties: bool = True,
-         **kwargs: Any) -> figure.Figure:
+         ax = None,
+         **kwargs: Any) -> axes.Axes:
     """Given the output data frame of :py:func:`covidcast.signal`, plot a choropleth or bubble map.
 
     Projections used for plotting:
@@ -79,7 +80,8 @@ def plot(data: pd.DataFrame,
       Defaults to `True`.
     :param kwargs: Optional keyword arguments passed to ``GeoDataFrame.plot()``.
     :param plot_type: Type of plot to create. Either choropleth (default) or bubble map.
-    :return: Matplotlib figure object.
+    :param ax: Optional matplotlib axis to plot on.
+    :return: Matplotlib axes object.
 
     """
     if plot_type not in {"choropleth", "bubble"}:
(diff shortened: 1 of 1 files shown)


chinandrew:
oops need to fix some things for tests

chinandrew:
> oops need to fix some things for tests

Fixed.

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -286,21 +289,22 @@ def _plot_bubble(ax: axes.Axes, data: gpd.GeoDataFrame, geo_type: str, **kwargs:
     ax.legend(frameon=False, ncol=8, loc="lower center", bbox_to_anchor=(0.5, -0.1))
 
 
-def _plot_background_states(figsize: tuple) -> tuple:
+def _plot_background_states(figsize: tuple, ax: axes.Axes = None) -> axes.Axes:
     """Plot US states in light grey as the background for other plots.
 
     :param figsize: Dimensions of plot.
```
capnrefsmmat, line 295: So figsize is ignored if axes are provided? Should it also be optional?

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -79,7 +80,8 @@ def plot(data: pd.DataFrame,
       Defaults to `True`.
     :param kwargs: Optional keyword arguments passed to ``GeoDataFrame.plot()``.
     :param plot_type: Type of plot to create. Either choropleth (default) or bubble map.
-    :return: Matplotlib figure object.
+    :param ax: Optional matplotlib axis to plot on.
```
capnrefsmmat, line 83: I think the docstring should say what happens if `ax` is not provided (axes are made, states are plotted). Maybe that goes in the text above, such as in a paragraph after the one-line summary explaining that the maps are plotted on US states by default.

Python-packages/covidcast-py/covidcast/plotting.py
```diff
@@ -92,26 +94,27 @@ def plot(data: pd.DataFrame,
...
-    return fig
+    return ax
 
 
 def plot_choropleth(data: pd.DataFrame,
                     time_value: date = None,
                     combine_megacounties: bool = True,
-                    **kwargs: Any) -> figure.Figure:
+                    **kwargs: Any) -> axes.Axes:
     """Plot choropleths for a signal. This method is deprecated and has been generalized to plot().
```
capnrefsmmat, line 111: Want to use the [deprecated directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated) here?

capnrefsmmat (commented):
Overall looks good; I think just needs some doc tweaks

Python-packages/covidcast-py/covidcast/plotting.py
chinandrew: added

Python-packages/covidcast-py/covidcast/plotting.py
chinandrew: I think I had a bug here, the background states are always plotted and I had left that statement out. I've also added a sentence describing the return behavior.

Python-packages/covidcast-py/covidcast/plotting.py
chinandrew: added. This also made me realize our changelog is a bit incorrect, will make another PR to fix that.

chinandrew:
running into some weird bugs with subplots, working through it now

chinandrew:
:/
Screenshot from 2020-12-07 09-44-11

chinandrew:
Fixed, was an axes limit issue

dshemetov:
This looks like a backwards-compatible change. Should we fix and merge or close? @capnrefsmmat

capnrefsmmat:
Hmm, if this code still works, we might as well merge it. I don't think it's worth developing new features for the package in the long run, but since Andrew already did the hard work here, we might as well use it.

Are you able to test this branch and see if it still works? We'd have to fix the merge conflicts in the tests as well.

PR #363, opened 2020-12-05
add title arg

### LINKED ISSUE #360: Add title argument to plotting functions
Currently the title is auto set as `f"{data_source}: {signal}, {day_to_plot.strftime('%Y-%m-%d')}"`, this issue is to make an optional argument to override that

chinandrew:
Based off #361. Fixes #360.

Summary of changes:
- Add an optional title argument to the plotting function.
- Add test.
- Add some docstrings in the test mocks that I thought might be useful.

### DIFF
--- a/Python-packages/covidcast-py/covidcast/plotting.py
+++ b/Python-packages/covidcast-py/covidcast/plotting.py
@@ -44,6 +44,7 @@ def plot(data: pd.DataFrame,
          plot_type: str = "choropleth",
          combine_megacounties: bool = True,
          ax: axes.Axes = None,
+         title: str = None,
          **kwargs: Any) -> axes.Axes:
     """Given the output data frame of :py:func:`covidcast.signal`, plot a choropleth or bubble map.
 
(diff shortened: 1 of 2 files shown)


capnrefsmmat:
Hmm. Is there a way to set a subtitle? Because it seems like a good strategy would be to take a custom title and let the subtitle be date. Otherwise, if you set a title, the date doesn't get displayed at all, right?

chinandrew:
> Hmm. Is there a way to set a subtitle? Because it seems like a good strategy would be to take a custom title and let the subtitle be date. Otherwise, if you set a title, the date doesn't get displayed at all, right?

matplotlib has a suptitle() (yes with a `p` and not `b`) that places a title centered on the figure, but it doesn't look great

test

Alternatively you can add a \n and that will lead to a second line. 

Is it a big problem if there's no date? A user always has the option to add the date, and if we force it in the subtitle and they don't want it, that could be annoying. The one case where I could see it making sense is when you pass in a df with multiple dates and it uses the most recent time value, having it tell you that date could be useful. However, we could also achieve that with a message/warning of some sort.

capnrefsmmat:
Ah, sorry, I was thinking of the animation method for some reason, maybe because it was being discussed on Slack. There it's important to have a date appear somewhere, because otherwise you have no idea which frame corresponds to what date.

chinandrew:
> Ah, sorry, I was thinking of the animation method for some reason, maybe because it was being discussed on Slack. There it's important to have a date appear somewhere, because otherwise you have no idea which frame corresponds to what date.

Good point. Maybe we add an optional `add_date` argument that adds `f"\n{date}"` to the end of a provided title?

dshemetov:
This looks like a backwards-compatible change. Do we want to fix and merge or close? @capnrefsmmat

Issue #362, opened 2020-12-04
Switch the API clients to use POST requests


capnrefsmmat:
If you request a *lot* of geo_values, the GET query can be too long, and you get Request-URI Too Long (HTTP 414) from the API server.

Fortunately the server seems to support POST with no modification. We should switch the clients to POST and test thoroughly. (Not sure what this will do to the R httptest tests...)

chinandrew:
The Python package uses `delphi-epidata` to make requests, so this should belong there. I'll open a PR for it.

capnrefsmmat:
Released in 0.4.0.

PR #366, opened 2020-12-07
Use POST instead of GET within R package

JedGrabman:
Description: 
Use POST instead of GET in the covidcast R package. This avoids a bug where the URI length limit of 8192 characters can be exceeded when there many `geo_values`, resulting in an error.

Change log:
- Change GET to POST in .request()
- Specify appropriate `max_geos` when not requesting all locations

- Update test files for POST.


capnrefsmmat:
See #362 -- I think you can avoid the `MAX_GEOS` part of this by switching `httr::GET` to `httr::POST`. But making `covidcast_days` aware of `geo_values` for batching is definitely worth keeping≥

JedGrabman:
Looks like the URI issue is being fixed [here](https://github.com/cmu-delphi/delphi-epidata/pull/309), so you're right that the MAX_GEOS part can be removed. I'll narrow it to just the `geo_values` efficiency update.

capnrefsmmat:
That fixes the Python client; the R package uses its own copy of the delphi-epidata code, so it'll need to be changed at the bottom of `covidcast.R`.

capnrefsmmat:
Specifically, the `.request()` function does the GET request and can be changed to use POST.

chinandrew:
Is there any reason why the covidcast R package doesnt use the epidata R client like the python one does? Also FYI there's some discussion [here](https://github.com/cmu-delphi/delphi-epidata/pull/309) regarding get vs post, and for the python client we're having it try a get and then fall back to post if it 414s

capnrefsmmat:
The delphi-epidata R client isn't distributed in any public package repo, only on GitHub, so we'd have to vendor it directly in the code. That's basically what we did, except with some additional modifications. (I actually don't remember how much I changed, since it's been so long since I used the original client...)

JedGrabman:
@capnrefsmmat @sgsmob This should be good for another review. Unfortunately, the test for this breaks our CI because of a bug in `mockery`. I was able to run the test locally though. I commented it out because there is already a [fix](https://github.com/r-lib/mockery/pull/39) in, so we should be able to uncomment it once `mockery` has their next release (unclear when that will be).

R-packages/covidcast/tests/testthat/test-covidcast.R
```diff
@@ -212,7 +214,7 @@ test_that("covidcast_days batches calls to covidcast", {
         stderr = NA,
         sample_size = NA
       )),
-    10)
+    2)
```
sgsmob, line 217: Why was this previously 10?

sgsmob (commented):
Generally looks good but have a quick question about the test.

R-packages/covidcast/tests/testthat/test-covidcast.R
JedGrabman: Mistake when making the test. I was originally was going to simulate 10 days of data, but it seemed like overkill and I changed it to 2 and failed to update the number. It had no impact on the test because the mock only references the first two items of the list, but decided to clean it up since I noticed it now.

sgsmob (approved):

R-packages/covidcast/R/covidcast.R
```diff
@@ -204,7 +211,7 @@ MAX_RESULTS <- 3649
 #'   [`msa_census`], [`state_census`]
 #' @export
 #' @importFrom rlang abort
-#' @importFrom dplyr %>%
+#' @importFrom dplyr %>% filter
```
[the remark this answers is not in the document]
JedGrabman: That was from the first draft, but was removed before I resubmitted.
More generally though, it ensures that the function would use `dplyr::filter` instead of another package's version

R-packages/covidcast/tests/testthat/test-covidcast.R
```diff
@@ -234,3 +236,39 @@ test_that("covidcast_days batches calls to covidcast", {
...
   expect_called(m, 2)
 })
+
+# This test requires the use of stub's depth parameter, but there is a bug in
+# mockery's current release (0.4.2) which doesn't handle locked bindings
+# properly, so it errors when running devtools::check(). The test can be run
+# using testthat::test_package("covidcast"), but overwrites some functions, so
+# devtools:load_all() should be run after. This should be fixed with the next
+# release of mockery, at which point we can uncomment the test.
+# See: https://github.com/r-lib/mockery/pull/39
```
krivard, line 246: Would you create an issue for this please so we don't forget?

R-packages/covidcast/R/covidcast.R
```diff
@@ -761,8 +764,8 @@ covidcast <- function(data_source, signal, time_type, geo_type, time_values,
 .request <- function(params) {
   # API call. Allow base API URL to be replaced, e.g. to use a staging/testing
   # server when needed.
-  response <- httr::GET(getOption("covidcast.base_url", default = COVIDCAST_BASE_URL),
-                        httr::user_agent("covidcastR"), query = params)
+  response <- httr::POST(getOption("covidcast.base_url", default = COVIDCAST_BASE_URL),
+                        httr::user_agent("covidcastR"), body = params)
```
krivard, line 768: [Sam raised](https://github.com/cmu-delphi/delphi-epidata/pull/309#issuecomment-740746189) that GET requests can be cached along the way, while POST request generally can't/aren't. This was handled in the python client by using GET by default, and switching to POST on 414. Is that relevant for httr calls?

R-packages/covidcast/tests/testthat/test-covidcast.R
JedGrabman: Done, #372

R-packages/covidcast/R/covidcast.R
JedGrabman: I would assume so, since it has to do with the type of request and not the origin. I'll change it.

krivard (approved):

PR #370, opened 2020-12-10
Add function to convert epiweeks to timestamps

chinandrew:
See [this slack thread](https://delphi-org.slack.com/archives/C01E5M9KVN1/p1607616621001000).

Add a function that converts dates to timestamps and epiweek dates (CDC format, defined as length 6) to timestamps. Epiweeks are converted to the start of the epiweek using the `epiweeks` package.

Updated d ocs and changelog too, so I think we're good to rebuild docs and then release 0.1.2 after this is merged

### DIFF
--- a/Python-packages/covidcast-py/covidcast/covidcast.py
+++ b/Python-packages/covidcast-py/covidcast/covidcast.py
@@ -226,10 +226,12 @@ def metadata() -> pd.DataFrame:
         metadata.
 
       ``min_time``
-        First day for which this signal is available.
+        First day for which this signal is available. For weekly signals, will be
+        the first day of the epiweek.
 
       ``max_time``
-        Most recent day for which this signal is available.
+        Most recent day for which this signal is available. For weekly signals, will be
+        the first day of the epiweek.
 
       ``num_locations``
         Number of distinct geographic locations available for this signal. For
--- a/Python-packages/covidcast-py/covidcast/covidcast.py
+++ b/Python-packages/covidcast-py/covidcast/covidcast.py
@@ -270,8 +270,8 @@ def metadata() -> pd.DataFrame:
                            meta["message"])
 
     meta_df = pd.DataFrame.from_dict(meta["epidata"])
-    meta_df["min_time"] = meta_df["min_time"].astype(str).apply(_parse_datetime_weeks)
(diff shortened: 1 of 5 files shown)


Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -324,6 +328,28 @@ def aggregate_signals(signals: list, dt: list = None, join_type: str = "outer")
...
+    Datetimes (length 8) are converted to their corresponding date, while epiweeks (length 6)
+    are converted to the date of the start of the week. Returns nan otherwise
+
+    Epiweeks use the CDC format.
+
+    :param date_str: String of date or epiweek
+    :param format: String of the date format to parse.
+    :returns: Timestamp
+    """
+    if len(date_str) == 8:
```
capnrefsmmat, line 344: Do you want to do this by using the `time_type` column instead of length? I don't know if we'll ever have two time_types with identical-length formatted dates, but it seems cleaner to use `time_type` anyway

Python-packages/covidcast-py/docs/changelog.rst
```diff
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+
+v0.1.2, December 10, 2020
+-------------------------
+
+- Bug fix: Fix issue where introduction of weekly signals broke :py:func:`covidcast.metadata` function.
```
capnrefsmmat, line 8:
```suggestion
- Bug fix: Fix issue where introduction of weekly signals broke the :py:func:`covidcast.metadata` function.
```

capnrefsmmat (commented):
Looks good, just one suggestion and a question

Python-packages/covidcast-py/covidcast/covidcast.py
krivard: I'm going to go ahead and insist on switching to `time_type`

chinandrew: ah, didn't realize that was a thing, good idea.

chinandrew: a bit messy of a .apply statement since we need both columns, but implemented now

capnrefsmmat (approved):
looks good to me

PR #379, opened 2020-12-11
housekeeping: rename api param source -> endpoint

sgratzl:
*REQUIRES* https://github.com/cmu-delphi/delphi-epidata/pull/325 to be merged and deployed

see https://github.com/cmu-delphi/delphi-epidata/issues/321

renames the api `source` param to `endpoint`

### DIFF
--- a/R-packages/covidcast/R/covidcast.R
+++ b/R-packages/covidcast/R/covidcast.R
@@ -437,7 +437,7 @@ covidcast_signals <- function(data_source, signal,
 #'
 #' @export
 covidcast_meta <- function() {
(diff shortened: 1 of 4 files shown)


sgsmob (approved):

dshemetov:
This looks like a backwards-incompatible change. Motion to close PR? @capnrefsmmat @krivard

krivard:
We moved the `talks` directory to another repo a while ago, so none of those changes are needed.

The change to covidcastR is no longer relevant since we switched the base url to `/epidata/covidcast/`.

PR #380, opened 2020-12-11
Support weekly time_type in covidcast R package

capnrefsmmat:
This adds support for `time_type = "week"` in `covidcast_signal()` and friends, by using the `MMWRweek` package to convert. Users still specify ordinary dates, and they get results back as ordinary dates -- the beginning of the corresponding epiweek. This is necessary to support the NCHS mortality data.

@JedGrabman Adding you for review because this touches some of your recent work on batching, so it's possible I messed something up. I think this cleans things up a bit, though, by having a sequence of dates generated outside `covidcast_days` so the internal logic doesn't have to know the specific `time_type`.


R-packages/covidcast/R/covidcast.R
```diff
@@ -163,6 +163,9 @@ MAX_RESULTS <- 3649
 #'   with `time_value` of June 3 will only be included in the results if its
 #'   data was issued or updated on June 6. If `NULL`, the default, return the
 #'   most recently issued data regardless of its lag.
+#' @param time_type The temporal resolution to request this data. Most signals
```
sgsmob, line 166: What happens if a user requests data at a different resolution?  For daily signals is it aggregated?  For weekly signals is it split?

sgsmob: Whatever the answer is, would be good to add tests to cover those cases.

sgsmob (changes requested):

R-packages/covidcast/R/covidcast.R
capnrefsmmat: It depends on if that time_type is in the API. The API has to explicitly offer every time_type; if it doesn't, the client just reports the error for a missing or non-existent signal. There's a separate issue here, which is that the API docs will need to expand to mention the time_type available for every signal.

sgsmob (approved):

R-packages/covidcast/R/covidcast.R
capnrefsmmat: I opened https://github.com/cmu-delphi/delphi-epidata/pull/335 to make sure this is clearly documented for each source

PR #382, opened 2020-12-11
Evalcast killcards

dajmcdon:
This is a rather massive PR. It handles a number of outstanding issues. Among them:

1. Allowing point forecasts.
2. Generalizing plotting functionality.
3. Merges in the pkgdown site.
4. The updated DESCRIPTION.
5. Allowing alternative `actual` signals for evaluation.
6. Warning rather than stopping.

But with many additions comes some other things. Nested tibbles are gone, some of the `as_of` functionality is no longer enforced, and the user is allowed to do less-advisable evaluations if they try.

The vignette is updated to reflect all the changes, so this was the main method for testing functionality. I added some tests as I went, but I also had to disable some of the existing tests.

Open problems to be addressed:
1. Better test coverage. Some the were removed need to be reworked (others are now moot).
2. Incorporating static versions of API calls (as suggested by @capnrefsmmat): see [here](https://enpiar.com/r/httptest/index.html).
3. Correctly using [test fixtures](https://testthat.r-lib.org/articles/test-fixtures.html) to avoid repeatedly running `get_predictions()` in all the tests.
4. `plot_trajectory()`: I intend to make a vignette showing how this should work. I don't think it does everything I want yet.
5. Full pass over the documentation.

I would characterize the overall status as: when I try to do the things I want to do, they *work*, but if you try something I didn't think of, it may break.


dajmcdon:
Just for reference, or to avoid breaking issues, the previous version of evalcast was [tagged and released](https://github.com/cmu-delphi/covidcast/releases/tag/evalcast-0.1). I guess for population monitoring?

dajmcdon:
CI checks are failing because we need to install the Ubuntu backend to `stringi`. See PR #383.

chinandrew:
> CI checks are failing because we need to install the Ubuntu backend to `stringi`. See PR #383.

If this PR introduces that dependency, I'd recommend putting the #383 change in this PR. It's just one line and then your PR accounts for all new dependencies in a self contained way and checks can run. Hopefully makes your life easier by not blocking checks with another PR.

If stringi was already imported and they recently changed some backend dependency so all checks are failing on all branches, ignore me.

dajmcdon:
@chinandrew good point. I didn't think of that...

sgsmob (changes requested):
I haven't looked at the tests yet, but I figured this review is already long enough and I may as well give you some things to correct.  Mostly looks good; mainly formatting/style questions and requests for more documentation.

In particular, I don't like that there are a lot of uses of `x %>% f(.data$y)` rather than `x %>% f(y)` since (as far as I understand) they are equivalent but the latter is much easier to understand and when I see the former I am primed to expect something unusual is happening such as `x %>% f(other$y)`

R-packages/evalcast/.gitignore
```diff
@@ -1 +0,0 @@
-docs
```
krivard, line 1: was this on purpose or did it get nuked in a rebuild?

krivard (commented):
ci change looks fine, but is `docs` supposed to be here?

R-packages/evalcast/man/evaluate_predictions.Rd
```diff
@@ -2,31 +2,50 @@
...
+or \code{\link[=get_predictions]{get_predictions()}} (class is \code{predictions_cards})}
 }
 \value{
-List of "score cards", with one element per ahead value.
+tibble of "score cards". Contains the same information as the
+\code{predictions_cards()} with additional columns for each err_measure and
+for the truth (named \code{actual}). All the columns may be necessary for
+some types of plotting, but it may be useful to shrink this down
 }
 \description{
```
sarah-colq, line 50: The description needs to be updated to reflect the fact that we don't use prediction cards in the previous way.  It's confusing even if it still works, since get_predictions can generate things for multiple dates and here it says it needs to be a tibble with each predictions card being the same date.  Clarification is needed.  (If you think this should be pushed to not this review, it can wait, but some sort of update isn't clear here).

R-packages/evalcast/man/compute_actual_vs_nominal_prob.Rd
```diff
@@ -0,0 +1,39 @@
...
+A tibble containing grp_vars, nominal_prob (the claimed interval
+coverage), prop_below (the proportion of actual values falling below
+the forecast quantile) and prop_above (the proportion of
+actual values falling above the forecast quantile). All
+proportions are calculated for each available probability at each
+combination of grouping variables by averaging over any variables listed
+in avg_vars.
+}
+\description{
+Compute calibration of a quantile forecaster
```
sarah-colq, line 33: How is this different from compute_calibration?

sarah-colq (commented):
I think the idea of cards is confusing, particularly for the predictions size.  Maybe we should get rid of the nomenclature entirely.

R-packages/evalcast/R/calibration.R
```diff
@@ -1,70 +1,158 @@
...
+  score_card <- score_card %>% 
+    filter(!is.na(.data$quantile)) %>%
+    select(all_of(grp_vars), all_of(avg_vars),
+           .data$quantile, .data$value, .data$actual) 
+  
+  score_card <- averaging_checks(score_card, grp_vars, avg_vars)
+  
+  score_card %>% 
+    group_by(across(all_of(c(grp_vars,avg_vars)))) %>%
+    mutate(rev_value = rev(.data$value)) %>%
```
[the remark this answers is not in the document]
dajmcdon: The "averaging_checks" on 84 should be enforcing it.

R-packages/evalcast/R/calibration.R
```diff
@@ -1,70 +1,158 @@
...
+    group_by(across(all_of(grp_vars)), .data$quantile) %>%
+    summarise(prop_below = mean(.data$below, na.rm = TRUE),
+              prop_above = mean(.data$above, na.rm = TRUE)) %>%
+    rename(nominal_prob = .data$quantile)
+  score_card
 }
+
+#' @inherit compute_calibration
+#' @export
+compute_actual_vs_nominal_prob <- compute_calibration
```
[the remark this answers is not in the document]
dajmcdon: It's for backwards compatibility. `compute_actual_vs_nominal_prob()` used to exist, but had basically the same functionality.

R-packages/evalcast/R/err_measures.R
```diff
@@ -24,75 +22,100 @@ weighted_interval_score <- function(quantile_forecasts, actual_value) {
...
+  quantile <- quantile[!point_fcast]
+  value <- value[!point_fcast]
+  if (all(is.na(actual_value))) return(NA)
+  actual_value <- actual_value[1]
+  num_prob <- length(quantile) # 23
+  if (num_prob %% 2 != 1 || num_prob < 3){
+    warning(paste("WIS calculation:",
+                  "Forecaster must return at least 3 symmetric probabilities",
+                  "for this calculation. Returning NA."))
+    return(NA)
```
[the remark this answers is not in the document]
dajmcdon: Since this applies to all locations all time periods etc, when we actually do `evaluate_predictions()`, this function may run thousands of times. So if there's one missing value, rather than gumming up everything, better to warn and keep going than stop everything.

R-packages/evalcast/R/err_measures.R
```diff
@@ -7,14 +7,12 @@
 #' al. (2020)](https://arxiv.org/abs/2005.12881) for discussion in the context
 #' of COVID-19 forecasting.
 #' 
-#' @param quantile_forecasts Tibble of quantile forecasts.
+#' @param quantile vector of forecasted quantiles
```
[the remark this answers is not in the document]
dajmcdon: It's easier to apply to the big dataframe at evaluation time.

R-packages/evalcast/R/evaluate.R
```diff
@@ -1,4 +1,4 @@
-#' Create a list of score cards
+#' Create a data frame of score cards
```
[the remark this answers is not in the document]
dajmcdon: @sarah-colq can you fix this?

R-packages/evalcast/R/evaluate.R
```diff
@@ -29,143 +29,175 @@
...
+    message("ahead = ", unique_ahead[iter])
+    scorecards[[iter]] <- evaluate_predictions_single_ahead(
+      filter(predictions_cards, .data$ahead == unique_ahead[iter]),
       err_measures = err_measures,
       backfill_buffer = backfill_buffer)
   }
-  return(scorecards)
+  scorecards <- bind_rows(scorecards)
+  class(scorecards) <- c("score_cards", class(scorecards))
+  scorecards
```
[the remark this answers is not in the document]
dajmcdon: The tidyverse guide says not to, though covidcast seems to use explicit. I've tried to catch most of them.

R-packages/evalcast/R/evaluate.R
```diff
@@ -29,143 +29,175 @@
...
+                                         forecast_dates,
+                                         incidence_period,
+                                         ahead,
+                                         geo_type,
+                                         geo_values)
+  if (nrow(target_response) == 0) {
+    return(empty_score_card(predictions_cards, err_measures))
+  }
   as_of <- attr(target_response, "as_of")
   . <- "got this idea from https://github.com/tidyverse/magrittr/issues/29"
```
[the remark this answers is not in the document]
dajmcdon: The . gets used in the warning message below, but has to be assigned something before being used or else there's an error.

R-packages/evalcast/R/get_predictions.R
```diff
@@ -80,28 +69,17 @@ get_predictions <- function(forecaster,
...
-#'   ahead.
-#' @param name_of_forecaster String indicating name of the forecaster.
-#' @template signals-template
-#' @template forecast_date-template
-#' @template incidence_period-template
-#' @template ahead-template
-#' @template geo_type-template
-#' @template geo_values-template
-#' @template apply_corrections-template
-#' @param ... Additional arguments to be passed to `forecaster()`.
```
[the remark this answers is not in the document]
dajmcdon: It's not an exported function but the roxygen documentation gets rendered. I'll add a line pointing to the previous function.

R-packages/evalcast/R/get_zoltar_predictions.R
```diff
@@ -0,0 +1,183 @@
...
+#' @param forecaster_names Vector of strings indicating of the forecaster(s)
+#'   (matching what it is called on the COVID Hub).
+#' @param forecast_dates Vector of Date objects (or strings of the form
+#'   "YYYY-MM-DD") indicating dates on which forecasts will be made. If `NULL`,
+#'   the default, then all currently available forecast dates from the given
+#'   forecaster in the COVID Hub will be used.
+#' @param geo_values vector of character strings containing FIPS codes of
+#'   counties, or lower case state abbreviations (or "us" for national). The
+#'   default "*" fetches all available locations
+#' @param forecast_type "quantile", "point" or both (the default)
```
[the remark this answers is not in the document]
dajmcdon: The point forecast has an `NA` in the quantile column. We use 0.5 internally, and duplicate it when uploading. But other teams have a mean forecast or something.

R-packages/evalcast/R/plot_measure.R
```diff
@@ -1,17 +1,13 @@
-#' Plot error measure
+#' Plot error measure (deprecated)
```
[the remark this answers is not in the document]
dajmcdon: Planning to replace it with something from the notebook. So I'm just keeping it around for now.

R-packages/evalcast/R/target.R
```diff
@@ -23,85 +22,100 @@ get_target_period <- function(forecast_date, incidence_period, ahead) {
...
+    MMWRweek = ew_frcst_date$MMWRweek,
+    MMWRday = 1) # 1 is Sunday
   # From https://github.com/reichlab/covid19-forecast-hub/blob/master/data-processed/README.md:
   # "For week-ahead forecasts with forecast_date of Sunday or Monday of EW12, a
   # 1 week ahead forecast corresponds to EW12 and should have target_end_date of
   # the Saturday of EW12. For week-ahead forecasts with forecast_date of Tuesday
   # through Saturday of EW12, a 1 week ahead forecast corresponds to EW13 and
   # should have target_end_date of the Saturday of EW13."
-  week_ahead <- ifelse(wday(forecast_date) <= 2, # forecasting on a Sun/Monday
+  week_ahead <- ifelse(lubridate::wday(forecast_date) <= 2, # forecasting on a Sun/Monday
```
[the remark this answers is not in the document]
dajmcdon: lines 7,8 technically make it unnecessary. My preference is: (1) use the namespace qualification if it's a handful of functions from some package or (2) import the whole package and drop the namespaces. I like 1 in case I don't know where the function came from later. I would generally avoid (2) but dplyr/ggplot2 are so ubiquitous it's really annoying. The @importFrom tags let you drop the namespace for a single imported function, but I don't like doing that.

So in this case I removed 7,8.

R-packages/evalcast/R/target.R
```diff
@@ -23,85 +22,100 @@ get_target_period <- function(forecast_date, incidence_period, ahead) {
...
-#'
-#' @importFrom rlang .data
-#' @importFrom tibble enframe
-#' @importFrom dplyr mutate left_join rename bind_rows group_by summarize
-#' @importFrom purrr pmap pmap_dfr pmap_lgl
-#' @importFrom assertthat assert_that
+
+
+
 get_target_response <- function(signals,
```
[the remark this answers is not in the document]
dajmcdon: Not exported, so I don't want any of that.

R-packages/evalcast/R/target.R
```diff
@@ -23,85 +22,100 @@ get_target_period <- function(forecast_date, incidence_period, ahead) {
...
                                 ahead,
-                                geo_type) {
+                                geo_type,
+                                geo_values) {
   response <- signals[1, ]
   target_periods <- forecast_dates %>%
     enframe(name = NULL, value = "forecast_date") %>%
-      mutate(incidence_period = incidence_period,
-           ahead = ahead) %>%
+    mutate(incidence_period = incidence_period, ahead = ahead) %>%
```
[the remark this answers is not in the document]
dajmcdon: the function arguments are scalars. That creates a new column in the data frame with the scalars repeated.

R-packages/evalcast/R/target.R
```diff
@@ -23,85 +22,100 @@ get_target_period <- function(forecast_date, incidence_period, ahead) {
...
-  problems <- target_periods %>%
-    mutate(not_available = .data$end > Sys.Date())
-  assert_that(!any(problems$not_available),
-              msg=paste0("When `ahead` is ", ahead, ", it is too soon to evaluate ",
-                         "forecasts on these forecast dates: ",
-                         paste(forecast_dates[problems$not_available], collapse=", "),
-                         "."))
+  target_periods <- target_periods %>%
+    mutate(available = .data$end <= Sys.Date()) 
+  bad_dates <- forecast_dates[!target_periods$available]
```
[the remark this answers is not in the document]
dajmcdon: I just want a vector, so this is faster than 
```r
target_periods %>% filter(!.data$available) %>% pull("forecast_dates")
```

R-packages/evalcast/.gitignore
dajmcdon: I'm not sure where that is. I think an early version had the docs folder buried in the package instead of alongside the covidcastR directory higher up.

R-packages/evalcast/man/evaluate_predictions.Rd
dajmcdon: You're right. Maybe check my (minor) change? If you don't like it, do you mind doing a PR?

R-packages/evalcast/man/compute_actual_vs_nominal_prob.Rd
dajmcdon: It's the same. That function was in the old version but does the same thing.

R-packages/evalcast/.gitignore
krivard: I think I mean that the former presence of `docs` in .gitignore implies that we didn't want these files in the repository. Has that changed? If you're not sure, then it would help to know the following:
* Are the files in the `docs` directory generated from source? 
* Are there any hand-edits we might want to maintain in the repository?
* Is there any other reason to retain the files in `docs` in the repository?

dajmcdon: Sorry, I should have been more clear. There is no `R-packages/evalcast/docs/` directory (nor should there be).

R-packages/evalcast/R/target.R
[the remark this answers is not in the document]
sgsmob: Can you add back in the first 2 lines of comments so this function isn't bare?

R-packages/evalcast/R/get_predictions.R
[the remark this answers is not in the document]
sgsmob: Can you keep the first line of comment please?

krivard (approved):
CI and .gitignore changes lgtm

R-packages/evalcast/vignettes/intro-evalcast.Rmd
```diff
@@ -44,87 +48,142 @@ library(evalcast)
...
-pc <- predictions_cards_cmu[[1]]
-ca_fips = covidcast::name_to_fips("California")
-ca_code = substr(ca_fips, 0, 2) # States are identified by a 2 character code
-pc[pc$location == ca_code, ]$forecast_distribution[[1]]
+predictions_cards_cmu %>% 
+  filter(geo_value == "mi", forecast_date == forecast_dates_july[1]) %>%
+  select(geo_value, quantile, value, forecast_date) %>% str()
 ```
 
+Forecasts are generally probabilistic, meaning that rather than predicting a single value, a distribution of values at different quantiles. For example, the Covidhub ensemble forecaster would have predicted the following distribution in California for `r predictions_cards_cmu$ahead[1]` week(s) after `r predictions_cards_cmu$forecast_date[1]`
```
JedGrabman, line 89: I find the phrasing in the first sentence unclear. How about, "Forecasts are generally probabilistic, meaning that they predict a likely distribution of values instead of a single value."? Possibly add another sentence about what quantiles are if that needs to be defined.

JedGrabman, line 89: The last part is rendered as "the Covidhub ensemble forecaster would have predicted the following distribution in California for 3 week(s) after 2020-07-20"

"For example, on 2020-07-20 the Covidhub ensemble forecaster made the following prediction for how many cases there would be in California 3 weeks later."

R-packages/evalcast/vignettes/intro-evalcast.Rmd
```diff
@@ -44,87 +48,142 @@ library(evalcast)
...
+```{r}
+predictions_cards_ens %>%
+  filter(geo_value == "ca", forecast_date == forecast_dates_july[1]) %>%
+  select(quantile, value) %>% str()
+```
+
+An `NA` in the `quantile` column indicates a point forecast (possibly different from the median).
```
JedGrabman, line 96: How about "(typically the median or mean)".

R-packages/evalcast/vignettes/intro-evalcast.Rmd
```diff
@@ -44,87 +48,142 @@ library(evalcast)
...
-scorecards[[1]]
+collapse_cards(scorecards) %>% str()
 ```
 
 **Technical note:** What does `backfill_buffer = 10` do?  When we evaluate a forecaster in backtesting we are assuming that we know what actually occurred.  However, in light of backfill, we may not trust the data for some period of time.  The argument `backfill_buffer` allows us to specify how many days until we believe that the data has "settled down" and is unlikely to be updated further.  The choice of this argument will depend on the particular signal you are forecasting.
 
 ### Plots of performance measures
 
-`evalcast` provides a number of tools to visually assess a forecaster (including comparisons with other forecasters):
+While `evalcast` provides a number of tools to visually assess a forecaster (including comparisons with other forecasters), because `score_cards` are just long data frames, many plots can be achieved using [{ggplot2}](http://ggplot2.tidyverse.org).
```
JedGrabman, line 134: Suggest moving the ggplot2 graph / comment below the evalcast ones. Basically, show off what evalcast can do, but then point out that you can customize as you like if evalcast isn't sufficient.

R-packages/evalcast/vignettes/intro-evalcast.Rmd
```diff
@@ -44,87 +48,142 @@ library(evalcast)
...
-predictions_cards_cmu <- get_covidhub_predictions("CMU-TimeSeries", forecast_dates_july)
-
+predictions_cards_ens <- get_covidhub_predictions(
+  "COVIDhub-ensemble", forecast_dates_july, ahead = 3, 
+  signal = "deaths_incidence_num") %>%
+  filter(nchar(geo_value)==2) # remove counties
+predictions_cards_cmu <- get_covidhub_predictions(
+  "CMU-TimeSeries", forecast_dates_july,  ahead = 3, 
+  signal = "deaths_incidence_num")
 # Make new predictions using baseline_forecaster
```
JedGrabman, line 63: This isn't a change you made, but it might help to split the `get_covidhub_predictions` calls from the `get_predictions` call, it's a bit confusing until you realize that they don't affect each other.

I think even adding ellipsis to the comments could help a lot:
```
# Retrieve past predictions from Covidhub...

# ... or make new predictions using baseline_forecaster.
```

R-packages/evalcast/vignettes/intro-evalcast.Rmd
```diff
@@ -44,87 +48,142 @@ library(evalcast)
...
-                                   err_measures, 
-                                   backfill_buffer = 10)
-```
-
-Let's look at the `baseline_forecaster`'s scorecard for 3-epiweek ahead forecasts.
+scorecards <- evaluate_predictions(
+  bind_rows(predictions_cards, 
+            predictions_cards_cmu, 
+            predictions_cards_ens) %>%
+    # not always data for these locations
```
JedGrabman, line 116: Does this cause an issue somewhere if we don't remove them? If so, is there a way to have the functions automatically identify / handle that case?

R-packages/evalcast/vignettes/intro-evalcast.Rmd
```diff
@@ -44,87 +48,142 @@ library(evalcast)
...
+
+Let's look one location:
+
 ```{r}
-pc <- predictions_cards_cmu[[1]]
-ca_fips = covidcast::name_to_fips("California")
-ca_code = substr(ca_fips, 0, 2) # States are identified by a 2 character code
-pc[pc$location == ca_code, ]$forecast_distribution[[1]]
+predictions_cards_cmu %>% 
+  filter(geo_value == "mi", forecast_date == forecast_dates_july[1]) %>%
```
JedGrabman, line 85: Can we still use `name_to_fips`? I imagine some users (particularly international ones) won't immediately recognize US postal abbreviations.

R-packages/evalcast/vignettes/intro-evalcast.Rmd
```diff
@@ -44,87 +48,142 @@ library(evalcast)
...
+            predictions_cards_ens) %>%
+    # not always data for these locations
+    filter(! geo_value %in% c("us", "vi", "gu", "mp")),
+  err_measures = err_measures,
+  backfill_buffer = 10
+  )
+```
+
+
+`evaluate_predictions()` returns a long data frame with one column for each error measure added to the `predictions_cards`. Keeping all quantile predictions may be useful for plotting or other downstream analysis. For a more compact version, use `collapse_cards()` to return one row for each geo_value/ahead/forecast_date/forecaster combination.
```
JedGrabman, line 124: Very minor, but above you refer to it as a long `data.frame` and here it's a "data frame". Either one is fine, but I'd prefer to use the same style throughout.

JedGrabman (commented):
Only reviewed `intro-evalcast.Rmd`. I got pretty specific with how things are phrased, but it was good to see the vignette because it reminded me of / taught me some new things.

R-packages/evalcast/R/download_signals.R
```diff
@@ -33,9 +22,66 @@ download_signal <- function(...) {
...
-  if (args$geo_type == "state") {
-    out$geo_value = substr(covidcast::abbr_to_fips(out$geo_value), 1, 2)
+  # if (args$geo_type == "state") {
+  #   out$geo_value = substr(abbr_2_fips(out$geo_value), 1, 2)
+  # }
+  out # %>% rename(location = .data$geo_value)
+}
+
+
+download_signals <- function(...,
```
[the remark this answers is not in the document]
sarah-colq: I will do this in a further PR.  Making a note.

R-packages/evalcast/R/evaluate.R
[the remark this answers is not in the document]
sarah-colq: Noted for further PR!

R-packages/evalcast/man/compute_actual_vs_nominal_prob.Rd
sarah-colq: Should we get rid of it then?

R-packages/evalcast/man/evaluate_predictions.Rd
sarah-colq: Yeah this is helpful.  I'll further review and possibly make a PR but I think it's good enough for now.

R-packages/evalcast/README.md
```diff
@@ -31,6 +31,6 @@ online](https://cmu-delphi.github.io/covidcast/evalcast/index.html).
 ## Getting started
 
 To get started using this package, view the Getting Started guide at
-`vignette("covidcast")`.
```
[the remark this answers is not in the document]
sarah-colq: Realized this is necessary for Github.  Still should be evalcastR @dajmcdon .

R-packages/evalcast/R/calibration.R
```diff
@@ -1,70 +1,194 @@
...
+averaging_checks <- function(
+  cards, 
+  grp_vars = c("forecaster", "forecast_date", "ahead"),
+  avg_vars = c("geo_value"),
+  return_checks = FALSE) {
+    
+  gr <- cards %>% 
+    group_by(across(all_of(grp_vars)))
+  checks <- gr %>% group_keys()
+  checks$valid <- sapply(gr %>% group_split(),
```
JedGrabman, line 49: Ran into an error while working on the vignette:
```
Error: Assigned data `sapply(...)` must be compatible with existing data.
x Existing data has 3 rows.
x Assigned data has 52 rows.
i Only vectors of size 1 are recycled.
```

I think you want `lapply` instead of `sapply`, but .data$valid is still a list.
```
    checks <- filter(checks, all(unlist(.data$valid))) %>% 
      select(-.data$valid)
```

Seems to work, but I wanted to confirm that that's the intended behavior.

dajmcdon: I think `sapply()` is ok here. It just reduces the list that `lapply()` would return to a vector. The issue is that `checks` has 3 rows, but `sapply(gr %>% group_split(),...)` is a length 52 vector. Can you show me how you're using the function?

R-packages/evalcast/vignettes/intro-evalcast.Rmd
dajmcdon: I may have called a `dataframe` too elsewhere...

R-packages/evalcast/vignettes/intro-evalcast.Rmd
dajmcdon: I removed the codes for states completely. The reason is that covidcast uses lowercase postal codes. So the goal is to be consistent with that. If you really want to convert, you can, but all the state-level data throughout the package uses letters. (This also avoids a bug in `fips_to_name(c("01", "10"))` which returns Alabama twice instead of Alabama and Delaware.

R-packages/evalcast/vignettes/intro-evalcast.Rmd
dajmcdon: `intersect_averagers()` should take care of it. Though in keeping with our discussion yesterday, I think it's good to emphasize that one should be aware of the issue when performing these analyses.

R-packages/evalcast/R/calibration.R
JedGrabman: I'm just calling into it from the vignette:
```
plot_calibration(scorecards, 
                 grp_vars = "forecaster", 
                 avg_vars = c("geo_value", "forecast_date"))
```
Looking into it more, seems like `sapply` can work, I think the matrix is just transposed.
```
> foo = sapply(gr %>% group_split(), 
+        function(x) evalcast:::check_valid_coverage_probs(x, avg_vars))
> dim(foo)
[1] 52  3
> foo
      [,1] [,2] [,3]
 [1,] TRUE TRUE TRUE
 [2,] TRUE TRUE TRUE
 [3,] TRUE TRUE TRUE
 [4,] TRUE TRUE TRUE
 [5,] TRUE TRUE TRUE
...
```
I can add the transpose call.

R-packages/evalcast/vignettes/intro-evalcast.Rmd
JedGrabman: Ah sorry, I meant `name_to_abbr` instead of `name_to_fips`. Made the change, I'll leave the comment open for you to object or resolve.

R-packages/evalcast/R/calibration.R
dajmcdon: Hmmm... something seems strange about that. `check_valid_coverage_probs()` should (I think) not result in a matrix like that. So it must be a bug.

JedGrabman: Just wrapped the `sapply`, in a `t()` call. Leaving comment open for comment or to resolve.

R-packages/evalcast/vignettes/intro-evalcast.Rmd
JedGrabman: I removed the line for now, since I thought it was more confusing than illuminating. It didn't describe why it's important or how to determine which locations to remove. It is probably a good idea to build in `intersect_averagers` into the plotting functions to prevent users from analyzing non-comparable data.

R-packages/evalcast/R/calibration.R
dajmcdon: I found the bug. See line 22 in `calibration.R`

jacobbien:
It'd be nice if `get_zoltar_forecast_dates` could return dates in order.

```
> get_zoltar_forecast_dates("COVIDhub-ensemble")
 [1] "2020-08-17" "2020-07-06" "2020-09-14"
 [4] "2020-07-27" "2020-08-24" "2020-09-07"
 [7] "2020-08-10" "2020-09-21" "2020-06-29"
[10] "2020-07-13" "2020-08-03" "2020-08-31"
[13] "2020-05-04" "2020-05-11" "2020-04-13"
[16] "2020-04-27" "2020-05-25" "2020-05-18"
[19] "2020-06-01" "2020-07-20" "2020-06-22"
[22] "2020-06-08" "2020-06-15" "2020-09-28"
[25] "2020-10-05" "2020-10-12" "2020-11-02"
[28] "2020-10-19" "2020-10-26" "2020-11-09"
[31] "2020-11-16" "2020-11-23" "2020-11-30"
[34] "2020-12-07" "2020-12-14"
```

jacobbien:
First off, I love that `get_covidhub_predictions` is much faster than before. Thanks @JedGrabman (I believe?).  Second, I get an error when I try to call `get_zoltar_predictions` to do the same thing.  Am I calling this function incorrectly?

```
pred <- get_covidhub_predictions("COVIDhub-ensemble",
                                 forecast_dates = "2020-12-07",
                                 signal = "deaths_incidence_num")
pred2 <- get_zoltar_predictions("COVIDhub-ensemble",
                                 forecast_dates = "2020-12-07",
                                 signal = "deaths_incidence_num")
Error in format_error_bullets(x[-1]) : 
  nms %in% c("i", "x", "") are not all TRUE
```

dajmcdon:
@jacobbien This should work now. There were some rogue `.data$location` mentions that needed to be removed.

PR #384, opened 2020-12-13
Add files via upload

mgithub40:
Added google, apple and waze mobility data


capnrefsmmat:
Could you say more about this data and how you imagine it being used in the covidcast package? The package is designed for fetching regularly updated data from the [COVIDcast Epidata API](https://cmu-delphi.github.io/delphi-epidata/api/covidcast.html); would these files need to be updated regularly? How were they generated? Is there documentation on what they contain? We'd probably prefer to distribute data through our API (like we do with our [SafeGraph mobility data](https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/safegraph.html)), so it is regularly updated and aggregated in a consistent way to all our other data sources, rather than adding new ad-hoc data sources.

mgithub40:
Dr. Alex Reinhart, 

Thank you for reviewing. Google and Apple mobility data can be integrated into covidcast to get access to data which displays information about the change in visits to grocery stores, pharmacy, parks, transit stations during the epidemic. It can be fine grained based on city and country level.  
As specified will update COVIDcast Epidata API to include the mobility data so that it can be fetched and updated regularly.
Also be made available to other data sources.

PR #387, opened 2020-12-17
Fix HRR maps

### LINKED ISSUE #381: HRR and MSA data not showing up properly in choropleth plots for covidcast
The data showing up in the R version looks different than the Python version.  We presume the R version is wrong because it was behaving before using the shapefiles instead of usmap.

capnrefsmmat:
Fixes #381 and adds tests to ensure these plots will continue to work. Test cases verified against the COVIDcast web site's rendering of  the same day of data.


R-packages/covidcast/R/plot.R
```diff
@@ -390,7 +390,10 @@ plot_choro = function(x, time_value = NULL, include = c(), range,
       is_alaska = substr(hrr_name, 1, 2) == 'AK',
       is_hawaii = substr(hrr_name, 1, 2) == 'HI',
       is_pr = substr(hrr_name, 1, 2) == 'PR',
-      color = ifelse(hrr_num %in% geo, col_fun(val[hrr_num]), missing_col))
+      # use the HRR numbers to index the named val vector -- but convert to
+      # character, otherwise the indices will bee positional, not using the
```
chinandrew, line 394:
```suggestion
      # character, otherwise the indices will be positional, not using the
```

sgsmob (approved):

PR #388, opened 2020-12-17
New COPSS-NISS talk

ryantibs:


capnrefsmmat:
One note on this line:

> The COVIDcast API is based on HTTP GET queries and returns data in JSON form.

It now supports returning CSVs directly if you add `&format=csv` to the URL, which makes things easier. Although using the packages is even easier still.

PR #389, opened 2020-12-20
changes to weekly report for counties specifically

dajmcdon:
Some minor corrections and a few additional plots for county level examination.

### DIFF
--- a/R-notebooks/weekly_report/covidhub_evaluation_djm.Rmd
+++ b/R-notebooks/weekly_report/covidhub_evaluation_djm.Rmd
@@ -1,0 +1,466 @@
+---
+title: "Evaluating County-Level Case Forecasts"
+author: "Delphi Group"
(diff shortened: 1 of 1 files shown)


JedGrabman:
A lot of good changes here, but I'd rather not add a new file, since it will quickly get out of sync. I will refactor the report, so that the same code will generate a separate county and state level report. Then, let's integrate your changes there so it can update both reports. 

We also may want to do something separate for the trajectory graphs. They are good, but they are more specific than the rest of the report, so I'll need to think about how to organize them and the exact scope we want for the weekly reports.

dajmcdon:
Absolutely. Feel free to just close this PR when you're ready and delete the branch. Better to incorporate the useful parts.

JedGrabman:
Closing this pull request without merging due to substantial unrelated refactors of the touched files.

PR #398, opened 2021-01-04
Modeltools update

ryantibs:
Hi all Happy New Year! This PR updates the modeltools vignettes to contain a new one that uses a quantgen forecaster tuned by cross-validation (actually, it uses forward-validation, but I'll still refer to it as CV for simplicity). This may be perceived cheating because I'm supposed to be on paternity leave since Jan 1, but to be honest this all represents work done before Jan 1, it just took a very long time for the CV simulation to finish 😉.

So you don't have to look in two places, I'm going to move all the comments from #334 (first modeltools PR) over to this one, and consolidate it all below. Also, I'm going to self-merge again, because this PR only touches the modeltools package and nothing else.

Warning: very long message ahead! But I feel like putting details down in written form is most useful now due to the fact I'll be mostly offline for a while.

## Modeltools maintainer

- I would love it if somebody else wanted to be the maintainer for the modeltools package. I only put myself down because R requires you to have a maintainer in the DESCRIPTION to build the package.

- At least, there's some light work that needs to be done on the quantgen forecast vignette itself (details below) so I would like it for somebody to do that. To me, makes sense for folks who are driving COVIDcast paper number 2 to do this, since it's so related to the analysis for that paper.

## Quantgen forecaster vignette

- I recommend keeping the vignette that appears in the pkgdown site to be about quantile autoregression, as [it is currently](https://htmlpreview.github.io/?https://github.com/cmu-delphi/covidcast/blob/modeltools/docs/modeltoolsR/articles/quantgen-forecast.html), and not about CV-tuned quantile lasso, as in [this new one](https://htmlpreview.github.io/?https://github.com/cmu-delphi/covidcast/blob/modeltools/docs/modeltoolsR/articles/quantgen-forecast-cv.html).

- The latter vignette isn't present in the index for the pkgdown site, so you can't find it in the "Articles" dropdown menu, but it's still there in the modeltools/vignettes/ (so it when you build the pkgdown site it gets put into docs/modeltoolsR/articles/ and you can still link to it externally). I'm only leaving it in here for convenience for the folks working on COVIDcast paper number 2; once they've seen it, and probably copied over the files to the repo for that paper, it can be deleted from modeltools/vignettes/.

- The exposition in the quantgen forecast vignette needs to be fleshed out (as with other vignettes). In doing so some of the longer comments I left in the code can be moved into the text.

- Some of the code also should be updated as evalcast gets fixed up, e.g., shouldn't need my custom definitions for `absolute_error()` or `weighted_interval_score()`; and shouldn't need to pass a temporary evaluation metric to `evalcast::get_predictions()` so that it just populates the `actual` column; etc.

-  I think we should integrate the functions `scale_by_forecaster()` and `canonical_plot()` (defined in the vignette) into the evalcast package, and then just call `evalcast::canonical_plot()` from this vignette. That would certainly shorten the vignette, which is currently too long. cc @dajmcdon @JedGrabman 

-  I also think we should shorten the vignette by reducing the number of plots we produce using `canonical_plot()`; basically, we only need show the third figure in the vignette, which gives the punchline and looks like this: 

Screen Shot 2021-01-04 at 5 08 01 PM

- The rest of the plots which demo the functionality available through `canonical_plot()` can be moved into a separate vignette, and it would make the most sense for this vignette to appear in the evalcast package. cc @dajmcdon @JedGrabman 

## Functionality/analysis for paper number 2

- @dajmcdon @jacobbien @huisaddison @ajgreen93 (folks driving COVIDcast paper number 2): please take a close look at the `quantgen_forecaster()` function and the [associated vignette](https://htmlpreview.github.io/?https://github.com/cmu-delphi/covidcast/blob/modeltools/docs/modeltoolsR/articles/quantgen-forecast.html). It should be useful for the analysis we undertake in paper number 2. cc @capnrefsmmat @brookslogan for coordination with the survey paper or COVIDcast paper number 1, either of which might benefit from a plot or two reflecting the results of such experiments (and referring to paper number 2 for the details).

- The [quantgen forecaster vignette](https://htmlpreview.github.io/?https://github.com/cmu-delphi/covidcast/blob/modeltools/docs/modeltoolsR/articles/quantgen-forecast.html) looks at predicting state-level incident COVID-19 cases. It compares a baseline model, and two quantile regression models: QAR3 (quantile autoregression with 3 lags) and QAR3 + CLI3 (quantile autoregression with 3 lags and 3 lags of the CLI-in-community signal as an exogenous feature). You can see QAR3 + CLI3 consistently outperform QAR3. In an aggregate sense (aggregated over all time) both outperform the baseline, but only when the forecast horizon is short enough, about 15 days or less---see the first four plots.  Broken down by time, you can see clear time periods where the baseline wins---see the rest of the plots.

- Comparing relative mean WIS to median relative WIS (note the difference in order of operations). Initially, in the county-level analysis I produced in [my blog post](https://delphi.cmu.edu/blog/2020/09/21/can-symptoms-surveys-improve-covid-19-forecasts/), I found that relative mean WIS was too noisy and only median relative WIS showed something interesting. Now, in the current state-level analysis, I'm finding that both are pretty interesting, and in fact the relative mean WIS plot shows a more consistent benefit to the QAR3 + CLI3 forecaster across all ahead values. Might be that county-level is just a whole totally different problem in terms of its inherent SNR, but truthfully I'm not really sure what would be the reason for this distinction. For the paper, I think we decided on looking at HRRs, which are somewhere in the middle in size/SNR, so I would just try both (by design, `canonical_plot()` makes it easy to toggle between different types of plots). 

- The simulation part of the vignette (saved out to an RData file) runs in a few hours. For HRRs it'll be about 6 times as expensive (I did 50 states, and there's about 300 HRRs), so barring any speedups on the evalcast side or otherwise, we may have to use some kind of parallelization in order to run lots of experiments efficiently.

- There is an additional [quantgen forecaster with CV vignette](https://htmlpreview.github.io/?https://github.com/cmu-delphi/covidcast/blob/modeltools/docs/modeltoolsR/articles/quantgen-forecast.html) that I'm adding as part of this PR. The computational cost is 100 times that of the non-CV version (20 values of lambda x 5 folds). The simulation part of the vignette (also saved out to an RData file) took several days to run. 

- The results are overall very similar to the non-CV version, which I interpret as a good thing: it means even when we take the problem "more seriously" and allow each of QAR3 and QAR3 + CLI3 to self-tune, we still see a benefit to the CLI-in-community signal. Importantly, here I left the first variable in each model---which is the latest value of the response available (first lag in the autoregressive features)---unpenalized. Thus when each considers a higher-level of penalization, it doesn't move them farther away from the baseline model (they're not forced to shrink the contribution of the highly important first lag in the autoregressive features).

- Also encouraging is that CV tuning seems to make QAR3 + CLI3 even more favorable compared to QAR3. Compare the last plot in the two vignettes, as an example: at the highest lag, you see less variability in the QAR3 + CLI3 scores.

- A very important note (I spoke with Dan, Jacob, Alden about this, but not Addison, I believe): in times of a real strong upswing in the response, I've found that the story becomes really muddled. This was based on rerunning my old blog post code for forecast dates in between mid October and mid December. I found that QAR3 + CLI3 (in the nomenclature of the current vignette) was actually worse than QAR3. This was in terms of median relative AE. Though both were much, much better than the baseline (median relative AE well below 1), because during a strong upswing, a flat line is a terrible forecaster. Reconciling/understanding this is going to be critical for COVIDcast paper number 2. We don't want to exclude the third wave just because it doesn't look flattering to us. My hope (call me an optimist) is that enabling penalization with CV tuning is going to be helpful, because then the QAR3 + CLI3 can "choose" to reduce to QAR3 when useful. (Perhaps even, we'll want to leave all autoregressive features unpenalized, in the QAR3 + CLI3 model.)  The only bad news here is that I think it's worth running the CV-tuned models for the paper, at least over the course of the third wave, which means a heck of a lot of computation. 

- (I can think of a way to make the CV setup way more efficient, basically by appropriately recycling work from CV at the previous forecast date---with forward-validation this should be possible---but this just requires more complicated dev work on the modeltools side, and I'm not sure it'd be worth it.)

- Finally, I unfortunately didn't get around to implementing the hotspot predictor function. It should follow in a pretty similar way to what I did with the quantgen forecaster, just using (say) glmnet to fit the sparse logistic regression instead of quantgen. There will be a bit more to do because you won't be operating on the raw response signal, but a binarized version (transformed into a 0-1 hotspot label). Please somebody create an issue for this so we can keep track. Hopefully then somebody can pick this up and implement, say, `glmnet_hotspot_predictor()` in the modeltools package, which we can then use for the hotspot experiments in our paper.

## Forecast evaluation notebook/dashboard

- @JedGrabman @kateharwood @brookslogan @dajmcdon (folks who are working on the forecast evaluation notebook and dashboard): please take a careful look at the function called `canonical_plot()`. I wrote it to be a kind of "one-function-to-produce-all" in terms of canonical plots for forecast evaluation. Unfortunately it grew a bit hairy (some of the function's body may be a bit hard to parse) in order to get all the corner cases correct.

- I was saying to Jed/Kate on slack a few days ago that I think one nice way to alleviate a lot of the "plot-overload" feel in our evaluation notebook would be to have a shiny/plotly app that allows you to toggle between different options. Mean or median.  Unscaled or scaled. Scale before aggregating or after, etc. This function allows you to do all that (and much more). You can see all the plots I produce in the vignette from it. So while this function could serve as the backend, we'd still need to investigate building a shiny/plotly app around it. 

- As I commented on slack such a tool might be the way to go to power the forecast dashboard, so we don't have to rewrite code in D3/JavaScript. If we do this, then we might want to refactor some of the code here, because currently nothing intermediate gets stored and each time you call it, it operates on the raw evaluations data frame, calls `scale_by_forecaster()` then aggregates to compute metrics, etc. The call to `scale_by_forecaster()` can in particular be a bit expensive (probably just a handful of seconds, but may be annoying lag for a dashboard) when `scale_before_aggr = TRUE`. 

- @brookslogan: might be worth drawing attention to folks from Reich Lab to this function, if they're producing similar plots for their forecast eval notebook. I've heard Nick before say that he doesn't think plotting code needs to be functionalized, but they still might appreciate it.

## Zyzzyva integration with evalcast

- @capolitsch @dajmcdon (and anybody else working on zyzzyva, I can't seem to find Mike O'Brien's GitHub handle): after I wrote it, I realized that `quanten_forecaster()`'s functionality may subsume that of zyzzyva. You should be able to compute the necessary variable transformations via the `transform` and `inv_trans` arguments, and compute slopes as features through the `featurize` (though I'm not sure you would need it, lags might be just as useful, worth trying). Beyond this I can't remember what makes zyzzyva different/special, so it's worth somebody taking a close look before attempting to port zyzzyva over to be evalcast-compatible, since `quanten_forecaster()` is already evalcast-compatible.

- Depending on where Collin's headed with the aardvark development (meaning, depending on what's working and what isn't), the template offered in `quanten_forecaster()` might be useful for aardvark too.


dajmcdon:
@sgsmob for Zyzzyva

PR #406, opened 2021-01-07
Remove 28-day specific code from dashboard functions

benjaminysmith:
This refactors dashboard code to not have hardcoded 28-day dependence, allowing us to change the interval.

Continues work on #32

### DIFF
--- a/R-notebooks/dashboard_functions.R
+++ b/R-notebooks/dashboard_functions.R
@@ -20,10 +20,14 @@ make_covidcast_signal <- function(destination, source, geo_type) {
   return(destination)
 }
 
-plot_28_day_unique <- function(df_to_plot, plural_geo_title) {
+plot_unique_geo_types_present <- function(df_to_plot, plural_geo_title) {
   samples_per_day = df_to_plot %>%
     group_by(time_value) %>%
     summarize(n = n())
+  
+  day_diff =
+    samples_per_day %>% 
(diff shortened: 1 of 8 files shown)


R-notebooks/dashboard_functions.R
```diff
@@ -20,26 +20,31 @@ make_covidcast_signal <- function(destination, source, geo_type) {
...
 
-plot_28_day_unique <- function(df_to_plot, plural_geo_title) {
+plot_unique_geo_types_present <- function(df_to_plot, plural_geo_title) {
   samples_per_day = df_to_plot %>%
     group_by(time_value) %>%
     summarize(n = n())
   
+  day_diff =
+    samples_per_day %>% 
+    summarise(difftime(max(time_value), min(time_value), units = "days"))
```
krivard, line 30: ...is this not just the full time span this data is available?

benjaminysmith: It's the range of time for the signal that was retrieved (e.g, if we requested 28 days worth of data and get back 25 due to 3 day lag, this is = 25), i.e., this makes the plotting functions extract the range from whatever signal data is passed in, rather than assuming any specific value.

krivard (approved):
xlnt

benjaminysmith:
@krivard could you plaase merge this?

PR #414, opened 2021-01-12
Add tests of basic functionality to evalcast

### LINKED ISSUE #402: Update tests for evalcast
I've been holding off on this because it will take some work. We need tests for the basic package functionality. Some of these existed before, but take too long to run. Currently, those are turned off. We need to turn them back on.

There are two issues. 

1. We don't want to call the API to run the tests. Alex suggested using a combination of the `mockery` and `httptest` packages. Useful also to consult the `covidcast` [tests](https://github.com/cmu-delphi/covidcast/blob/evalcast/R-packages/covidcast/tests/testthat/test-covidcast.R).
2. Some of the basic functionality involves slow repeated calls (running baseline forecaster multiple times with slightly different arguments, say) and then running those results through lots of tests. This is not really ideal for package development since the checks will take forever. We should likely use "test fixtures". See [this vignette]()https://testthat.r-lib.org/articles/test-fixtures.html#test-fixtures).

sgsmob:
Fixes #402


R-packages/evalcast/R/get_predictions.R
```diff
@@ -54,7 +54,7 @@ get_predictions <- function(forecaster,
                             geo_type,
                             geo_values = "*",
                             apply_corrections = NULL,
-                            signal_aggregation = c("list", "wide", "long"),
+                            signal_aggregation = "list",
```
dajmcdon, line 57: I think you should revert this change. My understanding is that the old version was the right way to do it. The `match.arg()` call below pulls out the first matching argument. The advantage of the old way is that the default shows the available options (rather than just the first).

R-packages/evalcast/tests/testthat/test-evaluate.R
```diff
@@ -0,0 +1,216 @@
...
+      data_source = "source",
+      signal = "signal",
+      target_end_date = as.Date("2020-01-05"),
+      incidence_period = "epiweek"
+    ))
+    expect_warning(evaluate_predictions(pcard, backfill_buffer = 22),
+                   "backfill_buffer")
+    # waited long enough should have no warning:
+    evaluate_predictions(pcard, backfill_buffer = 7)
+  })
```
dajmcdon, line 215: I don't really understand how any of this works, but I assume it is actually testing what we want!

dajmcdon (changes requested):

R-packages/evalcast/R/get_predictions.R
sgsmob: For whatever reason I thought this was causing a bunch of issues but apparently it is ok.  Reverted.

R-packages/evalcast/tests/testthat/test-evaluate.R
sgsmob: This is copied (more or less) from the existing skipped tests.  AFAICT we are testing whether the `backfill_buffer` argument does anything, but right now the only thing it does is raise a warning.  So we shouldn't get any warnings when all the dates are in the buffer.  I have explicitly added a check that creates an error when there are warnings on this run.  Ultimately we don't care what the output is here, but rather just that moving the buffer forwards or backwards has some impact.

PR #417, opened 2021-01-14
Incorporate `scale_by_forecaster()` into evalcast

### LINKED ISSUE #411: Move `scale_by_forecaster()` from modeltools into evalcast
See Ryan's PR #398. 

Adjust it's documentation/functionality as necessary.

sgsmob:
Adds the `scale_by_forecaster()` function from `modeltools` into this package.

Also fixes some tests broken by earlier changes.

Fixes #411


R-packages/evalcast/R/manage_cards.R
```diff
@@ -91,3 +91,34 @@ intersect_averagers <- function(cards, grp_vars, avg_vars) {
...
+#'   values the remaining forecasters' errors will be scaled
+#' @param err_cols vector of column names in `score_card` that contain error measures.
+#'   Elements of `err_cols` that are not in `vars` will be dropped from the final output to
+#'   avoid a mix of scaled and unscaled measures.
+#'
+#' @return
+#' @export
+scale_by_forecaster <- function(score_card, vars, denom_forecaster, 
+                                err_cols = c("ae", "wis")) {
+  df_list <- map(vars, function(var) {
```
dajmcdon, line 109: Can you add a few checks here to ensure that the necessary stuff is in the `score_card` and produce appropriate warnings/errors? For example, make sure `denom_forecaster` is available, and that there are others to scale. Check that the card has a "forecaster" column and contains (I guess at least 1) of the `err_cols`?

R-packages/evalcast/R/manage_cards.R
```diff
@@ -91,3 +91,34 @@ intersect_averagers <- function(cards, grp_vars, avg_vars) {
   }
   cards %>% right_join(averagers_intersected)
 }
+
```
dajmcdon, line 94: Let's move this function to it's own `.R` script.

R-packages/evalcast/tests/testthat/test-manage_cards.R
```diff
@@ -0,0 +1,18 @@
+test_that("scale_by_forecaster works", {
+    score_card <- tibble(
+        forecaster = c("f1", "f2", "f1", "f2"),
+        ahead = c(1, 1, 2, 2),
+        coverage_80 = c(1, 1, 0, 0),
+        ae = c(1, 2, 4, 16),
+        wis = c(3, 5, 7, 9)
+    )
+    scaled_card <- scale_by_forecaster(score_card, c("ae", "wis"), "f2")
```
dajmcdon, line 9: Maybe it's already tested (I can't quite tell), but if not, can you (1) test the checks I suggested above and (2) test that "Elements of `err_cols` that are not in `vars` will be dropped"?

dajmcdon (changes requested):

R-packages/evalcast/R/manage_cards.R
```diff
@@ -91,3 +91,34 @@ intersect_averagers <- function(cards, grp_vars, avg_vars) {
...
+      pivot_wider(names_from = "forecaster", 
+                  names_prefix = var, 
+                  values_from = var) %>% 
+      mutate(across(starts_with(var), ~ .x /
+                      !!sym(paste0(var, denom_forecaster)))) %>%
+      pivot_longer(cols = starts_with(var), 
+                   names_to = "forecaster",
+                   values_to = var) %>%
+      mutate(forecaster = substring(forecaster, nchar(var) + 1)) %>%
+      filter(forecaster != denom_forecaster)
```
JedGrabman, line 121: I wonder if somebody might want to keep the denom_forecaster for plotting purposes. Maybe have an optional argument to keep it.

R-packages/evalcast/R/manage_cards.R
```diff
@@ -91,3 +91,34 @@ intersect_averagers <- function(cards, grp_vars, avg_vars) {
...
+#' @param denom_forecaster name of forecaster in `score_card$forecaster` column by whose error
+#'   values the remaining forecasters' errors will be scaled
+#' @param err_cols vector of column names in `score_card` that contain error measures.
+#'   Elements of `err_cols` that are not in `vars` will be dropped from the final output to
+#'   avoid a mix of scaled and unscaled measures.
+#'
+#' @return
+#' @export
+scale_by_forecaster <- function(score_card, vars, denom_forecaster, 
+                                err_cols = c("ae", "wis")) {
```
JedGrabman, line 108: Recommend some fast fail errors, if `err_cols` not in the columns of `score_card` or `vars` not in `err_cols`.

R-packages/evalcast/R/manage_cards.R
```diff
@@ -91,3 +91,34 @@ intersect_averagers <- function(cards, grp_vars, avg_vars) {
   }
   cards %>% right_join(averagers_intersected)
 }
+
+#' Scale error measures based on those of a particular forecaster.
+#'
+#' @param score_card score_card like that returned by `evaluate_predictions()`
+#' @param vars vector of column names in `score_card` to normalize
+#' @param denom_forecaster name of forecaster in `score_card$forecaster` column by whose error
```
JedGrabman, line 99: Suggest `base_forecaster` or `baseline_forecaster` instead of `denom_forecaster`

R-packages/evalcast/R/manage_cards.R
```diff
@@ -91,3 +91,34 @@ intersect_averagers <- function(cards, grp_vars, avg_vars) {
...
+#'   values the remaining forecasters' errors will be scaled
+#' @param err_cols vector of column names in `score_card` that contain error measures.
```
JedGrabman, line 101: Recommend `score_cols`

R-packages/evalcast/R/manage_cards.R
```diff
@@ -91,3 +91,34 @@ intersect_averagers <- function(cards, grp_vars, avg_vars) {
...
+      mutate(across(starts_with(var), ~ .x /
+                      !!sym(paste0(var, denom_forecaster)))) %>%
+      pivot_longer(cols = starts_with(var), 
+                   names_to = "forecaster",
+                   values_to = var) %>%
+      mutate(forecaster = substring(forecaster, nchar(var) + 1)) %>%
+      filter(forecaster != denom_forecaster)
+  })
+  return(reduce(df_list, left_join))
+}
```
JedGrabman, line 124: This will return NaN's (if the baseline score is 0) or NA's (if the baseline doesn't exist). I think that behavior is correct, but might warrant a warning.

JedGrabman (commented):
No major changes, just suggested name changes and adding some warnings / early errors.

R-packages/evalcast/R/manage_cards.R
sgsmob: Wouldn't its values all be 1 though?

R-packages/evalcast/R/manage_cards.R
dajmcdon: Good point. Yes, we should do something about this. To avoid the NaN, we could add a tolerance parameter so that we divide by `max(baseline, tol)`. I think the NA is probably the right thing if the baseline doesn't have a forecast, but should probably document this behavior, and throw a warning before the return.

R-packages/evalcast/R/manage_cards.R
sgsmob: Implemented failure if `score_cols` (formerly, `vars`) is not a subset of `numeric_err_cols` (formerly `err_cols`).  I don't think it makes sense to validate that `numeric_err_cols` is a subset of `score_card`'s columns since if some are missing, who cares.

R-packages/evalcast/R/manage_cards.R
sgsmob: Done

R-packages/evalcast/R/manage_cards.R
sgsmob: Done

R-packages/evalcast/R/manage_cards.R
sgsmob: After writing more tests, the `across()` statement fails when all forecasts are from `baseline`.  It now throws an error if the baseline is not in the `forecaster` column or is the only entry.

R-packages/evalcast/tests/testthat/test-manage_cards.R
sgsmob: Done.

R-packages/evalcast/R/manage_cards.R
sgsmob: Done

R-packages/evalcast/R/manage_cards.R
sgsmob: Done.

R-packages/evalcast/R/scale_by_forecaster.R
```diff
@@ -0,0 +1,61 @@
...
+                                score_cols,
+                                base_forecaster_name, 
+                                numeric_err_cols = c("ae", "wis")) {
+    # Validate columns arguments
+    columns <- colnames(score_card)
+    assert_that("forecaster" %in% columns,
+                msg = 'score_card must have a column named "forecaster"')
+    
+    extraneous_score_cols <- setdiff(score_cols, columns)
+    assert_that(length(extraneous_score_cols) == 0,
```
dajmcdon, line 23: A bit more clear would be
`assert_that(all(score_cals %in% columns))`

R-packages/evalcast/R/scale_by_forecaster.R
```diff
@@ -0,0 +1,61 @@
...
+#' @param numeric_err_cols vector of column names in `score_card` that contain numeric error
+#'   measures. Elements of `numeric_err_cols` that are not in `score_cols` will be dropped from the
+#'   final output to avoid a mix of scaled and unscaled measures.
+#'
+#' @return
+#' @export
+scale_by_forecaster <- function(score_card,
+                                score_cols,
+                                base_forecaster_name, 
+                                numeric_err_cols = c("ae", "wis")) {
```
dajmcdon, line 16: I don't think I like this function definition. The user needs to name the columns to normalize and then also name all the numeric scores (or hope those in the score_card match the default). I understand the goal of dropping unscaled columns, but I think convenience should supercede.

If `score_card` has the right class, then we should know which columns it has, and the rest will be `numeric_err_cols`. We can then drop any that aren't in `score_cols`. I think the 4th argument should be something like `id_vars = NULL`. If score card has the right class, we ignore it, otherwise, if someone passes in a random tibble with some identifying vars and some `score_cols` they want scaled, then we require them to list those things.

R-packages/evalcast/R/scale_by_forecaster.R
```diff
@@ -0,0 +1,61 @@
...
+                msg = paste("score_cols contains columns",
+                            paste(extraneous_score_cols, collapse=", "),
+                            "not present in the columns of score_card"))
+
+    nonnumeric_score_cols <- setdiff(score_cols, numeric_err_cols)
+    assert_that(length(nonnumeric_score_cols) == 0,
+                msg = paste("score_cols contains columns",
+                            paste(nonnumeric_score_cols, collapse=", "),
+                            "that are not numeric errors"))
+
```
dajmcdon, line 33: Some of this will become unnecessary given my proposed changes to the arguments.

dajmcdon (changes requested):

R-packages/evalcast/R/manage_cards.R
JedGrabman: It would be. I can think of scenarios where a user would want that though. It makes it clear just by looking at the dataframe or a resulting plot that the data is scaled and by which forecaster.

R-packages/evalcast/R/manage_cards.R
JedGrabman: I would expect that the user who believes that the `numeric_err_cols` is part of the score card they provided would want to know that they are either specifying the wrong columns or using the wrong score card.

R-packages/evalcast/R/scale_by_forecaster.R
JedGrabman: > If score_card has the right class, then we should know which columns it has

`eval_predictions` can take a `side_truth` and arbitrary scoring functions, so I don't agree that we can know which columns will be present in a score card. The longer term solution is probably to have all objects of class `score_card` track which columns are identifiers vs. score columns.

dajmcdon: If the df comes out of `evaluate_predictions()`, we expect it to have some identifying columns and the `actual`. Anything else is either scores or extraneous. I'd rather the default behavior return the necessary id columns and the scaled scores (maybe the actuals) and dump the rest. If the user wants more columns in the return object, then they can `left_join()` the input and the results. If they provide something that's different than the `score_card` class, then let them specify. That way it "works" out of the box without requiring too much input.

R-packages/evalcast/R/scale_by_forecaster.R
sgsmob: Done

R-packages/evalcast/R/manage_cards.R
sgsmob: Obsolete given the removal of that argument

R-packages/evalcast/R/manage_cards.R
sgsmob: Added

R-packages/evalcast/R/scale_by_forecaster.R
sgsmob: changed to make the columns to keep an argument with a default value.

dajmcdon (approved):
Looks good to me. I'll wait on @JedGrabman before merging.

R-packages/evalcast/R/manage_cards.R
JedGrabman: I think your changes address if there is no baseline forecaster, but not if the value is 0.

R-packages/evalcast/R/scale_by_forecaster.R
```diff
@@ -0,0 +1,65 @@
+#' Scale error measures based on those of a particular forecaster.
+#'
+#' @param score_card score_card like that returned by `evaluate_predictions()`
+#' @param score_cols vector of column names in `score_card` to normalize
+#' @param base_forecaster_name name of forecaster in `score_card$forecaster` column by whose error
+#'   values the remaining forecasters' errors will be scaled
+#' @param id_cols vector of column names in `score_card` that identify distinct forecasts (i.e. the
+#'   independent variables of `score_card`).
+#' @param drop_base_entries whether to include the entries in `score_card` from
```
JedGrabman, line 9: Suggest changing either `drop` or `include` to match the other.

R-packages/evalcast/R/manage_cards.R
sgsmob: Done

R-packages/evalcast/R/scale_by_forecaster.R
sgsmob: Done

R-packages/evalcast/R/manage_cards.R
JedGrabman: The error text doesn't seem accurate, since division by 0 doesn't result in an error in R, it results in the value NaN. I think we should use a warning instead of an error, since a user may or may not want to keep the NaNs. Also, see dajmcdon's comment in this thread about using a tolerance parameter.

sgsmob: Ok I have changed it to just be a warning but have let the NaN remain

JedGrabman (approved):

dajmcdon (approved):

Issue #418, opened 2021-01-15
fips_to_name can be quite slow for long vectors


jacobbien:
Here's a relatively small example (just one month) demonstrating the slowness: 
```
dat <- covidcast_signal(data_source = "safegraph", 
                         signal = "restaurants_visit_prop",
                         start_day = "2020-12-13", 
                         end_day = "2021-01-13",
                         geo_type = "county")

system.time({a <- fips_to_name(dat$geo_value)})
# user  system elapsed 
# 34.926   0.113  35.112 
```
The function `fips_to_name` calls `grep_lookup`, which has a for loop over `dat$geo_value`.  This is wasteful because `dat$geo_value` will often have a large number of repeated elements.

capnrefsmmat:
Fixed in #419.

PR #419, opened 2021-01-15
Make fips_to_name faster.

jacobbien:
This is to address #418.  The idea is to pass `grep_lookup()` a vector of the unique elements of the fips vector.

In the issue referenced above, I gave a small example that took over 30 seconds.  The same example is now about 20 times faster:

```
dat <- covidcast_signal(data_source = "safegraph", 
                         signal = "restaurants_visit_prop",
                         start_day = "2020-12-13", 
                         end_day = "2021-01-13",
                         geo_type = "county")

system.time({a <- fips_to_name(dat$geo_value)})
# user  system elapsed 
# 1.685   0.007   1.697 
```

### DIFF
--- a/R-packages/covidcast/NEWS.md
+++ b/R-packages/covidcast/NEWS.md
@@ -1,3 +1,12 @@
+# covidcast 0.4.1
+
+Released TODO.
+
+## Minor changes
+
+- Geographic lookup functions, such as `fips_to_name()` or `name_to_cbsa()`, are
+  now much faster when given input vectors with many duplicate values.
+
 # covidcast 0.4.0
 
 Released January 13, 2021.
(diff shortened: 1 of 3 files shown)


sgsmob (approved):

capnrefsmmat:
Thanks, this is a good idea. Since there are a bunch of functions using `geo_lookup`, and they're all likely to have the same slow performance when given a vector with duplicates (like if you have ten months of state-level data), I've pushed a commit that moves your deduplication into `grep_lookup`. Now all of the functions should benefit.

I also pushed a few extra tests, since it turns out this behavior wasn't tested at all. I think this should be ready to go.

jacobbien:
Great!

sgsmob (approved):

dajmcdon:
@capnrefsmmat Can you let me know when this gets merged to `main/covidcast`? There's a very similar fix in `evalcast`, so I'll swap it out for the covidcast version.

capnrefsmmat:
It should be out in covidcast 0.4.1.

Issue #399, opened 2021-01-05
`download_signals()` failures


dajmcdon:
From Ryan:

I’ve encountered weird bugs running `get_predictions()`  in some cases where I think what’s happening is that training data for the forecaster “as of” the given forecast date wasn’t available at all.  I think this is a failure in `download_signals()`, where it probably tries to manipulate an empty object or something like that, instead of catching that it got back an empty data frame `covidcast_signals()`.  This is just my guess; aside from just looking at download_signals() to see if you can anticipate the problem you could try to call `get_predictions()` with the baseline forecaster for cases data from USAFacts at the start of June.  This is what I was doing (if my memory serves me right) before I ended up switching to JHU.

dajmcdon:
The problem seems to be that there is no data in the API for with a pre-June `as_of`. That's because we didn't have data at that point.

I think we should (potentially) do two things:

1. Stop completely with an interpretable error if download_signals() returns no data. Should be implemented [around here](https://github.com/cmu-delphi/covidcast/blob/e0f4457619c7f98b96ca16a3208f41473a74f74f/R-packages/evalcast/R/get_predictions.R#L127). This should explain that, given the desire to do proper backtesting, you wouldn't actually be able to make predictions at that time. We can’t anticipate what the forecaster would return, so it would be difficult to, for example, return empty predictions for that forecast date and try others.
1. Add an argument to `get_predictions()` that overrides the `as_of` and strongly discourage using it in the documentation. To a large degree, doing this eliminates the purpose of `evalcast`. Now we're just downloading current data and running a forecaster on it, so you should just use `covidcast` plus your forecaster.

I’m tagging @jacobbien in case there’s something I’m not thinking of. We should definitely do 1. I'm less convinced that 2. is a good idea.

dajmcdon:
Rereading Ryan's initial comment, it sounds at least possible that the error happens inside `download_signals()` before the [linked line](https://github.com/cmu-delphi/covidcast/blob/e0f4457619c7f98b96ca16a3208f41473a74f74f/R-packages/evalcast/R/get_predictions.R#L127). Probably still not interpretable. @sarah-colq can you verify where it happens?

sarah-colq:
[...]

Issue #404, opened 2021-01-05
Wedge plots


dajmcdon:
Remove the top part of the "X".

PR #426, opened 2021-01-20
Wedge plots and download signals fix 

sarah-colq:
Fix download_signals failures. Update wedge plots.

Fixes issues #399 , #404 .

### DIFF
--- a/R-packages/evalcast/R/download_signals.R
+++ b/R-packages/evalcast/R/download_signals.R
@@ -65,16 +65,18 @@ download_signals <- function(...,
   
   out <- base::suppressMessages({covidcast_signals(...)})
   empty_signals <- sapply(out, function(x) (nrow(x) == 0))
-  if(any(empty_signals)) {
-    missing <- signals[empty_signals, ]
-    missing_text <- paste("\t-", missing$signal, "from data_source",
+
+  missing <- signals[empty_signals, ]
+  assert_that(!any(empty_signals),
+              paste("For as_of date", args$as_of,
+              "there is no data in the covidcast API for signals:\n",
+              paste(paste("\t-", missing$signal, "from data_source",
                           missing$data_source, "between", missing$start_day,
-                          "and", args$end_day)
-    stop(paste("For as_of date", args$as_of,
-               "there is no data in the covidcast API for signals:\n",
-               paste(missing_text, collapse="\n"))
-        )
-  } 
+                          "and", args$end_day),
+                    collapse="\n")
+                   )
+              )
+  
   if (signal_aggregation != "list") {
     out <- covidcast::aggregate_signals(out, dt = signal_aggregation_dt,
                                         format = signal_aggregation)
--- a/R-packages/evalcast/R/download_signals.R
+++ b/R-packages/evalcast/R/download_signals.R
@@ -64,6 +64,17 @@ download_signals <- function(...,
   for (i in seq_along(msg)) message(msg[i])
   
   out <- base::suppressMessages({covidcast_signals(...)})
+  empty_signals <- sapply(out, function(x) (nrow(x) == 0))
+  if(any(empty_signals)) {
+    missing <- signals[empty_signals, ]
+    missing_text <- paste("\t-", missing$signal, "from data_source",
+                          missing$data_source, "between", missing$start_day,
+                          "and", args$end_day)
+    stop(paste("For as_of date", args$as_of,
+               "there is no data in the covidcast API for signals:\n",
+               paste(missing_text, collapse="\n"))
+        )
+  } 
   if (signal_aggregation != "list") {
     out <- covidcast::aggregate_signals(out, dt = signal_aggregation_dt,
                                         format = signal_aggregation)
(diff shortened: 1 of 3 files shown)


sarah-colq:
Fixed PR from previous version (I had a bunch of changes to main for other files in my evalcast-killcards branch)

R-packages/evalcast/R/get_predictions.R
```diff
@@ -74,11 +75,10 @@ get_predictions <- function(forecaster,
                  apply_corrections = apply_corrections,
                  signal_aggregation = signal_aggregation,
                  signal_aggregation_dt = signal_aggregation_dt),
-                 params))) %>%
-    bind_rows()
-
-  # for some reason, `value` gets named and ends up in attr
```
sgsmob, line 80: I think this comment should be left in

R-packages/evalcast/R/get_predictions.R
```diff
@@ -29,6 +29,8 @@
 #' @param signal_aggregation_dt for any data format, 
 #'   [covidcast::aggregate_signals()] can perform leading and lagging for you.
 #'   See that documentation for more details.
+#' @param use_covidcast this argument indicates whether your forecaster uses
+#'   covidcast data.  True by default.
```
dajmcdon, line 33: We should carefully explain  what will happen if you don't. Specifically, it looks like the forecaster needs to have its first argument be a NULL data frame.

dajmcdon: I'm actually not sure what we should expect to happen if we don't use covidcast data. Then we're just running their forecaster on their data and formatting it to work with this package's other tools? Maybe we should rethink: if they don't want any data, call a different function called `make_compatible_predictions()` or something.

dajmcdon (changes requested):

R-packages/evalcast/R/get_predictions.R
sgsmob: +1.  Is there a specific use case in mind for this argument we have here?

sarah-colq: I mean the idea was to use other forecasters.  Do we want to completely seperate that out from get_predictions?

dajmcdon: Maybe I didn't fully think this through before. It seems like there are two cases:

1. Someone has a forecaster that they want to evaluate with evalcast but using no covidcast data.
2. Someone wants some data from covidcast, but also some other data.

I think for the first case, we should write a separate function that somehow makes a prediction card that conforms. It's not really clear how this would work. Maybe something that allows one to "fill up" a prediction card with whatever they want?

For the second case, the additional data can just be interior to the forecaster, so this is maybe not an issue. But I think it *is* worth providing a demonstration. Suppose I want to do an AR(1) for weekly cases and also include unemployment claims. How would you do that?

dajmcdon (changes requested):

sarah-colq:
@dajmcdon To be clear, we want for the first case to pass their forecaster and the data to forecaster?  The other option is just to pass in outputted data to format it.   The first case makes more sense to me, and is easier to do, because you just need to define your forecaster with certain output columns (or define a wrapper that does it for you) and then you make a card.

dajmcdon:
@sarah-colq I think the rest of this PR is fine. For now, can you revert the change to `get_predictions()` and we'll figure out how to address the issue in our meeting tomorrow and eventually as a separate PR?

sarah-colq:
@dajmcdon @sgsmob Is this a decent make_compatible_predictions function?  Before I fix everything else?  The parameters that are still there are those required to make a predictions card, and can be used by a forecaster.

sarah-colq:
@dajmcdon Welp we commented at the same time.

R-packages/evalcast/R/make_compatible_predictions.R
```diff
@@ -0,0 +1,101 @@
...
+  out %>% 
+    mutate(
+      forecaster = name_of_forecaster,
+      forecast_date = forecast_date,
+      data_source = signals$data_source[1],
+      signal = signals$signal[1],
+      target_end_date = get_target_period(forecast_date, 
+                                          incidence_period, ahead)$end,
+      incidence_period = incidence_period) 
+}
```
sgsmob, line 101: Trailing newline

sgsmob (approved):

R-packages/evalcast/R/get_predictions.R
```diff
@@ -74,11 +77,11 @@ get_predictions <- function(forecaster,
                  apply_corrections = apply_corrections,
                  signal_aggregation = signal_aggregation,
                  signal_aggregation_dt = signal_aggregation_dt),
-                 params))) %>%
-    bind_rows()
-
+                 use_covidcast = use_covidcast,
+                 params)))
+  out <- bind_rows(out)
```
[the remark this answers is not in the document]
sarah-colq: This was a merge error on my part.  It's been fixed back now.

sarah-colq:
@dajmcdon Now the PR is just wedge plots and download_signals, as discussed.   Just need your approval on the changes.

dajmcdon (approved):

PR #427, opened 2021-01-21
Score function reliability update

JedGrabman:
* Allows `interval_coverage` to handle point estimates
* Adds a function for checking for common parameter issues in score functions (`score_func_param_checker`)
* Adds various tests for score functions
* Changes duplicate quantiles to be an error instead of a warning


R-packages/evalcast/R/err_measures.R
```diff
@@ -104,6 +99,35 @@ interval_coverage <- function(coverage) {
...
+#'
+#' @param quantiles vector of forecasted quantiles
+#' @param values vector of forecasted values
+#' @param actual_value actual_value, either as a scalar or a vector of the same 
+#'   length as `quantiles` and `values`
+#' @param id string to identify the caller of the function and displayed in
+#'   error messages (recommended to be the parent function's name)
+score_func_param_checker <- function(quantiles, values, actual_value, id = ""){
+  id_str = paste0(id, ": ")
+  if(length(actual_value) > 1){
```
sarah-colq, line 115: Are there cases where having a vector of actual values the same length as quantiles would be encountered?

sarah-colq (commented):
Just wanted to confirm the one case is necessary.  (see above comment).  Otherwise LGTM

R-packages/evalcast/R/err_measures.R
dajmcdon: yeah, depending on the score card, it may be replicated.

dajmcdon (approved):

PR #429, opened 2021-01-29
Add protected flag for asynchronous calls in covidcast

chinandrew:
Adding a constant to covidcast that, if set to True, will call an asynchronous version of `_fetch_single_geo()` that sends 100 requests at a time. I've set it to False so there should be no behavior changes, but if an indicator wants to use it they can. This was done with the combo indicator and nowcast indicator in mind I guess there's an argument to having the indicators make the async call directly through the Epidata package and not even surfacing this in covidcast, so open to discussion there. @capnrefsmmat @krivard thoughts?

Note this currently will not work until the latest Epidata client is published to pypi, so I've been running it with a local Epidata file. There's also a change in here to no longer loop through all geos after https://github.com/cmu-delphi/delphi-epidata/pull/237

usage example:
```
In [1]: from datetime import date                                                                                                                                    

In [2]: from covidcast import covidcast                                                                                                                              

In [3]: %time sync_df = covidcast.signal("fb-survey", "smoothed_cli", date(2020,4,6), date(2020,12,31), "state")                                                     
CPU times: user 6.62 s, sys: 345 ms, total: 6.97 s
****Wall time: 3min 33s****

In [4]: covidcast._ASYNC_CALL = True                                                                                                                                 

In [5]: %time async_df = covidcast.signal("fb-survey", "smoothed_cli", date(2020,4,6), date(2020,12,31), "state")                                                    
CPU times: user 1.39 s, sys: 62.5 ms, total: 1.45 s
****Wall time: 3.25 s****

In [6]: sync_df.equals(async_df)                                                                                                                                     
Out[6]: True
```

Have not written tests yet.


krivard:
I'm in favor

chinandrew:
With the release of Epidata 0.0.11 (thanks Katie), this is ready for review

Python-packages/covidcast-py/Makefile
```diff
@@ -22,7 +22,6 @@ build:
 
 lint:
 	. env/bin/activate; pylint covidcast/ --rcfile .pylintrc
-	. env/bin/activate; mypy covidcast --config-file mypy.ini
```
chinandrew, line 25: to standardize with indicator repo

Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -415,6 +425,56 @@ def _fetch_epidata(data_source: str,
...
+                         start_day: date,
+                         end_day: date,
+                         geo_type: str,
+                         geo_value: Union[str, Iterable[str]],
+                         as_of: date,
+                         issues: Union[date, tuple, list],
+                         lag: int) -> Union[pd.DataFrame, None]:
+    """Fetch data for a single geo asynchronously.
+
+    signal() wraps this to support fetching data over an iterable of
```
krivard, line 439:
```suggestion
    signal() wraps this to support fetching data over an iterable set of
```
?

krivard (approved):
Looks reasonable to me; lmk if you still want to wait for Alex

Python-packages/covidcast-py/covidcast/covidcast.py
chinandrew: This reminded me the docstring is a bit out of date, updated it.

krivard:
(also why is this failing evalcast build?)

chinandrew:
> (also why is this failing evalcast build?)

Appears to be unrelated to this PR, see [this thread](https://delphi-org.slack.com/archives/C013AH5N01E/p1613174696151100?thread_ts=1613173622.151000&cid=C013AH5N01E)

chinandrew:
@capnrefsmmat poke for review

capnrefsmmat (commented):
Does this require editing setup.py to specify a minimum version of delphi-epidata that supports async calls?

chinandrew:
> Does this require editing setup.py to specify a minimum version of delphi-epidata that supports async calls?

Good idea, updated

capnrefsmmat (approved):
lgtm, although I've never used the async epidata features

PR #430, opened 2021-01-29
Remove individual geo calls

chinandrew:
Redundant after https://github.com/cmu-delphi/delphi-epidata/pull/237. Also removed a test that expected 2 calls with 2 geos.

### DIFF
--- a/Python-packages/covidcast-py/covidcast/covidcast.py
+++ b/Python-packages/covidcast-py/covidcast/covidcast.py
@@ -175,26 +175,8 @@ def signal(data_source: str,
                          "start_day = '{start}', end_day = '{end}'".format(
                              start=start_day, end=end_day))
 
-    if isinstance(geo_values, str):
-        # User only provided one, not a list
-        geo_values = [geo_values]
-
-    dfs = [
-        _fetch_single_geo(
-            data_source, signal, start_day, end_day, geo_type, geo_value,
-            as_of, issues, lag)
-        for geo_value in set(geo_values)
-    ]
-
-    try:
-        # pd.concat automatically filters out None
-        out = pd.concat(dfs)
-    except ValueError:
-        # pd.concat raises ValueError if all of the dfs are None, meaning we
-        # found no data
-        return None
-
-    return out
+    return _fetch_single_geo(
+        data_source, signal, start_day, end_day, geo_type, geo_values, as_of, issues, lag)
 
 
 def metadata() -> pd.DataFrame:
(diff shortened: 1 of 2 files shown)


Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -175,26 +175,8 @@ def signal(data_source: str,
...
-        # pd.concat automatically filters out None
-        out = pd.concat(dfs)
-    except ValueError:
-        # pd.concat raises ValueError if all of the dfs are None, meaning we
-        # found no data
-        return None
-
-    return out
+    return _fetch_single_geo(
+        data_source, signal, start_day, end_day, geo_type, geo_values, as_of, issues, lag)
```
capnrefsmmat, line 179: seems like we should merge this function and `_fetch_single_geo`, since `_fetch_single_geo` no longer fetches a single geo?

capnrefsmmat (changes requested):
seems like a good opportunity to simplify, or at least rename `_fetch_single_geo` so its name is no longer false...

Python-packages/covidcast-py/covidcast/covidcast.py
chinandrew: Renamed to _fetch_epidata and refactored a little bit. Didn't merge  it with signal() yet in case we go ahead with #429

Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -379,15 +368,15 @@ def _detect_metadata(data: pd.DataFrame,
...
-                      as_of: date,
-                      issues: Union[date, tuple, list],
-                      lag: int) -> Union[pd.DataFrame, None]:
+def _fetch_epidata(data_source: str,
+                   signal: str,  # pylint: disable=W0621
+                   start_day: date,
+                   end_day: date,
+                   geo_type: str,
+                   geo_value: str,
+                   as_of: date,
```
capnrefsmmat, line 377: type annotation false: can't this now be a str or list of strs?

capnrefsmmat (commented):
safe to merge with the type annotation fixed, I think

Python-packages/covidcast-py/covidcast/covidcast.py
chinandrew: ah yes, added

capnrefsmmat (approved):
Looks good

PR #434, opened 2021-01-31
Add geo_type column to predictions_cards so evaluate_predictions()  doesn't need to guess

### LINKED ISSUE #433: [evalcast] extend `evaluate_predictions()` support beyond geo_type in ('state', 'county')
Logic here assumes predictions cards will only have geo type in state or county: https://github.com/cmu-delphi/covidcast/blob/evalcast-killcards/R-packages/evalcast/R/evaluate.R#L138-L143  Specifically, it tests the lengths of strings and if there are strings of varying lengths, it assumes that there are a mix of geo types

Issue: some geo types have strings of varying lengths, e.g., hrr, which are numbers (cast to string by covidcast) of length 1-3.

Proposal: Require predictions cards to also have a geo_type column and check that there is only one unique value there instead.

huisaddison:
Resolves #433

May break compatibility with predictions_cards created before this commit (because they don't have a geo_type() column); new predictions_cards (generated by get_predictions()) will have the necessary column.

### DIFF
--- a/R-packages/evalcast/R/evaluate.R
+++ b/R-packages/evalcast/R/evaluate.R
@@ -135,7 +135,7 @@ get_covidcast_data <- function(predictions_cards,
   assert_that(length(incidence_period) == 1,
               msg = "All predictions cards should have the same incidence
                      period.")
-  geo_type <- unique(predictions_card$geo_type)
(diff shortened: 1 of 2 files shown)


dajmcdon:
See #435

PR #435, opened 2021-02-01
Evalcast hotfix geotype

dajmcdon:
This PR supersedes @huisaddison #434 and addresses #433. 

*Goal*: better determination of the geo_type in `evaluate_predictions()` where possible (specifically, to allow `hrr` or other choices). This functionality is fully tested.

* Issues: there is a failing test related to `download_signals()`. I believe this is the cause of the `zyzzyva` failure in production. On first glance, it appeared to be due to a mixup between `signals` and `signal` which don't both necessarily exist. I tried to fix this, but there seems to be more trouble. Please @sgsmob and @sarah-colq check this out. `test-predictions` fails.
* I cleaned up a bunch of R CMD notes that the CI hasn't been catching. These are often of the form "unknown global variable" due to naming a column in a data frame after piping. These must be prefixed with `.data$` to avoid. As in `df %>% select(var)` should be `df %>% select(.data$var)`.


R-packages/evalcast/R/geo_type_selector.R
```diff
@@ -0,0 +1,50 @@
...
+  # tries to determine which geo_type we want and/or validate
+  # our selection based on the content of predictions_cards
+  # 
+  # Wraps the function below for ease of testing (we test the validator,
+  # rather than this function)
+  pred_card_types <- predictions_cards %>%
+    transmute(type_len = nchar(.data$geo_value),
+              has_chars = str_detect(.data$geo_value, "[aA-zZ]")) %>%
+    distinct()
+  geo_type <- match.arg(geo_type, c("county", "hrr", "msa", "dma", "state"))
```
sgsmob, line 12: This should be the first line so we quit without doing anything if the wrong type is specified.

R-packages/evalcast/R/geo_type_selector.R
```diff
@@ -0,0 +1,50 @@
...
+validate_geo_type <- function(geo_type, pred_card_types) {
+  if (nrow(pred_card_types) > 1L) {
+    assert_that(geo_type %in% c("hrr", "msa", "dma"),
+                msg = paste("predictions_cards",
+                            "contain multiple geo_types."))
+  }
+  if (any(pred_card_types$has_chars)) {
+    assert_that(geo_type %in% c("state","county"),
+                msg = paste("predictions_cards contain non-numeric characters",
+                            "but geo_type was not `state`"))
```
sgsmob, line 25: or `county`

R-packages/evalcast/R/geo_type_selector.R
```diff
@@ -0,0 +1,50 @@
...
+    } else {
+      assert_that(!pred_card_types$has_chars, # actually is state
+                  pred_card_types$type_len == 5L,
+                  msg = paste("input was geo_type = ",
+                              "county, but predictions_cards do not entirely",
+                              "conform to this selection."))
+    }
+  }
+  return(geo_type)
+}
```
sgsmob, line 50: trailing newline

R-packages/evalcast/R/evaluate.R
```diff
@@ -111,51 +114,48 @@ evaluate_predictions <- function(
...
                      period.")
-  geo_type_len <- predictions_cards %>%
-    mutate(type_len = nchar(geo_value)) %>%
-    distinct(type_len)
-  assert_that(nrow(geo_type_len) == 1,
-              msg = "All predictions cards should have the same geo_type.")
-  geo_type <- ifelse(geo_type_len$type_len == 2L, "state", "county")
-
+  
+  geo_type <- geo_type_selector(geo_type, predictions_cards)
```
sgsmob, line 143: This return value is never used so it should not be named.

R-packages/evalcast/R/geo_type_selector.R
```diff
@@ -0,0 +1,50 @@
...
+    assert_that(geo_type %in% c("hrr", "msa", "dma"),
+                msg = paste("predictions_cards",
+                            "contain multiple geo_types."))
+  }
+  if (any(pred_card_types$has_chars)) {
+    assert_that(geo_type %in% c("state","county"),
+                msg = paste("predictions_cards contain non-numeric characters",
+                            "but geo_type was not `state`"))
+  }
+  if (geo_type == "state") {
```
sgsmob, line 27: I don't understand why we want to try to automatically correct the geo type.

R-packages/evalcast/R/plot_trajectory.R
```diff
@@ -213,42 +214,4 @@ setup_plot_trajectory <- function(predictions_cards,
...
-    MMWRweek::MMWRweek2Date(MMWRyear = ew_day$MMWRyear,
-                            MMWRweek = ew_day$MMWRweek,
-                            MMWRday = xxx) + 7
-  } else {
-    MMWRweek::MMWRweek2Date(MMWRyear = ew_day$MMWRyear, 
-                            MMWRweek = ew_day$MMWRweek, 
-                            MMWRday = xxx)
-  }
-}
+}
```
sgsmob, line 217: Trailing newline

sgsmob (changes requested):
Haven't taken a look at the failing test but I have some small changes I would like to see.

sgsmob:
I can't replicate any test failure by downloading this branch.

dajmcdon:
Here's what I get when I run R CMD Check locally:

```
  Running the tests in ‘tests/testthat.R’ failed.
   Last 13 lines of output:
     Error: the condition has length > 1
     Backtrace:
         █
      1. ├─evalcast::get_predictions(...) test-predictions.R:104:4
      2. │ └─`%>%`(...)
      3. ├─dplyr::bind_rows(.)
      4. │ └─rlang::list2(...)
      5. └─purrr::map(...)
      6.   └─evalcast:::.f(.x[[i]], ...)
      7.     ├─base::do.call(...)
      8.     └─(function (forecaster, name_of_forecaster, signals, forecast_date, ...
     
     [ FAIL 1 | WARN 0 | SKIP 10 | PASS 189 ]
     Error: Test failures
     Execution halted
✓  checking for non-standard things in the check directory ...
✓  checking for detritus in the temp directory
```

Collin is getting it on the `evalcast-killcards` branch as well. See [here](https://delphi-org.slack.com/archives/C01D9R5QDT7/p1612187981006400).
   See
     ‘/Users/dajmcdon/Documents/COVID-Delphi/covidcast/R-packages/evalcast.Rcheck/00check.log’
   for details.
   
```

sgsmob:
I see.  This only shows up to me as a warning.  I think it is because we have multiple types for `geo_value`.  It is checking `if (geo_value == "*")` which evaluates to a vector of bools when `geo_value` contains a list of string values (that should all be false).

R-packages/evalcast/R/evaluate.R
dajmcdon: It gets used on line 165.

R-packages/evalcast/R/geo_type_selector.R
dajmcdon: I think we probably shouldn't, but we did previously. The new version *should* detect state vs. county automatically (the old version did this), but also allow geo_type as an argument. The old version didn't have the argument, so we would break a number of things if we don't keep the auto-detection.

R-packages/evalcast/R/evaluate.R
sgsmob: Ah, my ctrl+f skipped that for whatever reason

sgsmob (approved):

dajmcdon:
I fixed the `geo_value == "*"` thing which cleared up the test failure. I'm still unable to reproduce the errors that @capolitsch and @bnaras reported. So I hope that my change to `download_signals()` handles it.

Issue #436, opened 2021-02-01
Use API instead of HTML parsing when retrieving forecast dates


JedGrabman:
`evalcast::get_covidhub_forecast_dates` does html parsing to find what dates a forecaster has submitted for. This breaks if github changes the html of pages they return (which they appear to have just done). I'm making a quick fix to look for the new html, but the longer term solution is to use the github API and retrieve the file names instead.

jacobbien:
Just want to connect this issue to the previous issue #99.

JedGrabman:
Using the Zoltar API may be preferable to the Github API since it's already used in the package and we could potentially transition to using `get_zoltar_predictions` as well.

PR #437, opened 2021-02-01
Fix date retrieval

JedGrabman:
Forecast date retrieval in evalcast broke, apparently because Github added a div to their HTML. This is a short-term fix to get the functionality working again, but issue #436 would be the long term solution.


sarah-colq (approved):
LGTM

dajmcdon (approved):

Issue #432, opened 2021-01-29
`evalcast::download_signals()` issue (on `evalcast-killcards` branch)


bnaras:
@dajmcdon The `assert_that()` call [here](https://github.com/cmu-delphi/covidcast/blob/0a84602a4dbe7c9b28f1d461ec89a7cb3fedef69/R-packages/evalcast/R/download_signals.R#L70) assumes that `empty_signals` is a logical vector. But it is not as the attached log shows where I have used an instrumented version of the `evalcast` package with print statements to see what is going on. It turns out that `empty_signals` is a list of two elements, one of which is `NULL`, because `out` itself is also list of tibbles and a trailing `NULL` and `sapply` is unable to simplify (the `s` in `sapply`) and silently returns back a list.  Lines 2750+ of the attached print values for inspection. 
[run_zyzzyva.txt](https://github.com/cmu-delphi/covidcast/files/5896043/run_zyzzyva.txt)

dajmcdon:
@sarah-colq This has to do with the fix you made the other day. Can you try to debug?

sgsmob:
I can't replicate this.  Below is the output of `print(sapply(out, head))` and `print(empty_signals)` right after they are initialized.
```
            [,1]        [,2]        [,3]        [,4]       
data_source Character,6 Character,6 Character,6 Character,6
signal      Character,6 Character,6 Character,6 Character,6
geo_value   Character,6 Character,6 Character,6 Character,6
time_value  Numeric,6   Numeric,6   Numeric,6   Numeric,6  
issue       Numeric,6   Numeric,6   Numeric,6   Numeric,6  
lag         Integer,6   Integer,6   Integer,6   Integer,6  
value       Numeric,6   Numeric,6   Numeric,6   Numeric,6  
stderr      Logical,6   Logical,6   Numeric,6   Numeric,6  
sample_size Logical,6   Logical,6   Numeric,6   Logical,6  
[1] FALSE FALSE FALSE FALSE
```
What date are you using for the forecast date?

dajmcdon:
Naras and I just chatted on Zoom. It seems to be intermittent and hard to reproduce, possibly in `covidcast` rather than `evalcast`.  The result he was getting from `download_signals()` was something like
```r
ttt <- list(a=tibble(z=1), b=tibble(z=2), c=NULL)
sapply(ttt, function(x) nrow(x)==0) # old check
[...]

PR #438, opened 2021-02-01
fix download_signals to handle NULL tibbles

dajmcdon:
Hopefully addresses #432


dajmcdon:
It looks like lots of changes, but not really. Mostly I wanted to clean up the .R files. 
1. I moved functions around to files with matching names. 
2. I also rebuilt the documentation and the pkgdown site (when I uncovered the vignette fail #449 ). 
3. I deleted `plot_measure()`.
4. I stopped exporting `get_target_period()` 
5. I exported `plot_canonical()`
6. I cleaned up the multiple/missing final lines on most files.

PR #441, opened 2021-02-02
properly cast factor to numeric

huisaddison:
Calling as.numeric() on a factor variable of real numbers, as happens here: https://github.com/cmu-delphi/covidcast/blob/main/R-packages/modeltools/R/quantgen.R#L270 has undefined behavior.  In some systems, the coercion goes through as expected.  In others, the level codes are returned instead.  

The fix ensures that we always obtain the factor values, not the factor level codes.

Reference: https://stackoverflow.com/questions/3418128/how-to-convert-a-factor-to-integer-numeric-without-loss-of-information

### DIFF
--- a/R-packages/modeltools/R/quantgen.R
+++ b/R-packages/modeltools/R/quantgen.R
@@ -267,7 +267,8 @@ quantgen_forecaster = function(df, forecast_date, signals, incidence_period,
                    names_to = "quantile",
                    values_to = "value") %>%
       mutate(quantile = as.numeric(quantile),
-             value = as.numeric(value),
+             # https://stackoverflow.com/questions/3418128/how-to-convert-a-factor-to-integer-numeric-without-loss-of-information
+             value = as.numeric(levels(value))[value],
              ahead = a)
     
     # TODO: allow train_obj to be appended to forecaster's output. This would


jacobbien:
I wanted to understand why `value` would be a factor, so I looked into this a bit, and I think I see what's happening.
 
https://github.com/cmu-delphi/covidcast/blob/6bca816b62e700cc90f31b0e3d5a6c0edda31da5/R-packages/modeltools/R/quantgen.R#L264-L271

`predict_mat` is a matrix of numeric, but when it's `cbind`-ed to a character vector, the numeric entries become character entries.  So you get "5.11329053298113" as a value.

Ok, and now here's the interesting thing.  I believe that new versions of R now changed the default of the `stringsAsFactors` argument of `data.frame` from TRUE to FALSE.  So `as.data.frame` now does not convert characters to factors by default (@huisaddison I think this explains why we were getting different results).

So the fix above would take a numeric to a character to a factor to a numeric!  I think the better fix would be to avoid the numeric to character conversion in the first place.  I think the solution is actually quite simple: Replace `cbind` with `bind_cols`.

huisaddison:
Superseded by #442

PR #442, opened 2021-02-02
Modeltools quantgen_forecaster - avoid numeric to character conversion

jacobbien:
This addresses the problem described in #441 identified by @huisaddison.  This PR implements the fix that I described in my comment on that PR.  To make this PR self-contained, here's a copy of that comment:

I wanted to understand why `value` would be a factor, so I looked into this a bit, and I think I see what's happening.
 
https://github.com/cmu-delphi/covidcast/blob/6bca816b62e700cc90f31b0e3d5a6c0edda31da5/R-packages/modeltools/R/quantgen.R#L264-L271

`predict_mat` is a matrix of numeric, but when it's `cbind`-ed to a character vector, the numeric entries become character entries.  So you get "5.11329053298113" as a value.

Ok, and now here's the interesting thing.  I believe that new versions of R now changed the default of the `stringsAsFactors` argument of `data.frame` from TRUE to FALSE.  So `as.data.frame` now does not convert characters to factors by default (@huisaddison I think this explains why we were getting different results).

So the fix above would take a numeric to a character to a factor to a numeric!  I think the better fix would be to avoid the numeric to character conversion in the first place.  I think the solution is actually quite simple: Replace `cbind` with `bind_cols`.


capnrefsmmat:
I'll approve once @huisaddison does, since modeltools isn't in the CODEOWNERS file and GitHub defaults to requiring review from Katie or me. Who should be listed as owners for modeltools?

R-packages/modeltools/R/quantgen.R
```diff
@@ -106,7 +106,7 @@
 #'   `value`. The `quantile` column gives the probabilities associated with
 #'   quantile forecasts for that location and ahead. 
 #' 
-#' @importFrom dplyr filter select pull summarize between
+#' @importFrom dplyr filter select pull summarize between bind_cols
```
krivard, line 109: I'm mildly surprised to find that `%>%` is not needed in this import, since it's not a built-in operator. Maybe `modeltools` is always used in an environment where `tidyr` is already loaded?

capnrefsmmat: Looks like another function must already `@importFrom %>%`, since it's in the `NAMESPACE`: https://github.com/cmu-delphi/covidcast/blob/6bca816b62e700cc90f31b0e3d5a6c0edda31da5/R-packages/modeltools/NAMESPACE#L15

huisaddison:
Ran locally on Ryan's vignette; works as expected.

huisaddison (approved):

huisaddison:
> I'll approve once @huisaddison does, since modeltools isn't in the CODEOWNERS file and GitHub defaults to requiring review from Katie or me. Who should be listed as owners for modeltools?

@capnrefsmmat I have volunteered to be listed as maintainer for modeltools; so I guess that would be me...

capnrefsmmat (approved):
rubber-stamping for Addison

krivard (approved):
👍

dajmcdon:
@capnrefsmmat or @krivard I think one of you has to merge this as CODEOWNERS

PR #445, opened 2021-02-04
Allow for evaluation of "nation" geo_type

JedGrabman:
Leave `geo_type` validation to COVIDcast. This will allow us to use `geo_type` of "nation" or "hhs". Also link to COVIDcast's documentation, which will be the single source of truth for acceptable `geo_type`s.

**Change Log**:
* Remove `geo_type` validation and automated switching between counties / states when mis-specified.
* Require `geo_type` and don't provide defaults
* Update `geo_type` template to link to COVIDcast documentation.

**Additional notes**:
* COVIDcast currently throws an error if  you attempt to specify an unsupported `geo_type`
* If a supported `geo_type` is specified, but the `geo_values` do not match, COVIDcast will return no results. Evalcast will issue a warning per day that there are no errors


R-packages/covidcast/R/covidcast.R
```diff
@@ -204,7 +204,7 @@ COVIDCAST_BASE_URL <- 'https://api.covidcast.cmu.edu/epidata/api.php'
 #' @importFrom dplyr %>%
 covidcast_signal <- function(data_source, signal,
                              start_day = NULL, end_day = NULL,
```
dajmcdon, line 206: this change should be removed. If you reinstall `covidcast` it loads 0.4.1. The new options are
```r
geo_type = c("county", "hrr", "msa", "dma", "state", "hhs", "nation")
```

dajmcdon: Unfortunately, this branch has gotten too far behind `main`.

R-packages/evalcast/R/evaluate.R
```diff
@@ -67,7 +67,7 @@ evaluate_predictions <- function(
   backfill_buffer = 10,
   side_truth = NULL,
   grp_vars = c("forecaster", "forecast_date", "ahead", "geo_value"),
-  geo_type = c("county", "hrr", "msa", "dma", "state")) {
+  geo_type = c("county", "hrr", "msa", "dma", "state", "nation")) {
```
dajmcdon, line 70: see previous comment: should also support `hhs`

dajmcdon (changes requested):
I'm not quite sure I understand what's going on here. The current version is meant to be backwards compatible but allow new geo_types that weren't previously supported. This means ensuring that we can still discriminate between state/county. Other than those, the earlier version of `evalcast-killcards` won't work. So as long as we can still discriminate between those 2 without additional arguments, no existing code breaks. Then we can simply use the argument for any other `geo_type`. If we could just require the a `geo_type` argument (my preference), all the old stuff will break.

So the logic of the old version was: default to `county` as `covidcast` does. If it looks like the `predictions_cards` have states, then switch. Anything else requires a user to request a geo_type.

Now, I missed the fact that `covidcast` 0.4.1 has "nation" and "hhs", but those can be added. Is there a reason to try to parse out all these things? It's possible that national data in the predictions_card could cause an error in the current version, but I'm not sure we need anything more than that.

R-packages/evalcast/R/geo_type_selector.R
```diff
@@ -1,50 +1,93 @@
...
+  if (length(geo_type_lengths) > 1) {
+    assert_that(any(geo_types %in% c("hrr", "msa", "dma")),
+                msg = paste("Provided geo_values vary in length, which is not",
+                            "allowed for provided geo_types:", geo_types))
+  }
+  return_type <- NULL
+  for (geo_type in geo_types) {
+    if (geo_type == "county") {
+      if ((geo_type_lengths == 5) & !any(has_letters)) {
+        return_type <- "county"
```
sgsmob, line 69: I think this idiom here and below can be replaced with a simple `return(X)` and then a `return(NULL)` at the end if we find nothing.

R-packages/evalcast/R/geo_type_selector.R
```diff
@@ -1,50 +1,93 @@
...
+  has_letters <- str_detect(geo_values, "[aA-zZ]")
+
+  if (length(geo_type_lengths) > 1) {
+    assert_that(any(geo_types %in% c("hrr", "msa", "dma")),
+                msg = paste("Provided geo_values vary in length, which is not",
+                            "allowed for provided geo_types:", geo_types))
+  }
+  return_type <- NULL
+  for (geo_type in geo_types) {
+    if (geo_type == "county") {
```
sgsmob, line 67: Since these nested ifs have no corresponding else statements, consider combining them each into a single if.  However, I also understand why it might be more clear as written (branch on the `geo_type` then verify it works).  Your call.

R-packages/evalcast/R/geo_type_selector.R
```diff
@@ -1,50 +1,93 @@
...
-                              "state, but predictions_cards do not entirely",
-                              "conform to this selection."))
+      geo_types <- geo_type
     }
   }
-  if (geo_type == "county") {
-    if (pred_card_types$type_len == 2L && pred_card_types$has_chars) {
-      geo_type <- "state" # switch this automatically
+  found_geo_type <- validate_geo_type(geo_types,
+                                     predictions_cards$geo_value,
```
sgsmob, line 31: align indent

R-packages/evalcast/tests/testthat/test-validate_geo_type.R
```diff
@@ -1,23 +1,30 @@
...
-  expect_equal(validate_geo_type("state", pct), "county")
-  expect_equal(validate_geo_type("hrr", pct), "hrr")
-  pct <- tibble(type_len = 1L, has_chars = TRUE)
-  expect_error(validate_geo_type("state",pct))
-  expect_error(validate_geo_type("county",pct))
-  pct <- tibble(type_len = 1L, has_chars = FALSE)
-  expect_error(validate_geo_type("state",pct))
-  expect_error(validate_geo_type("county",pct))
+  state_geos <- c("ak", "mn")
+  expect_error(validate_geo_type("xyz", state_geos))
```
sgsmob, line 3: `expect_error` can take another argument to match the error message.  Would be stronger here to include the error message since there are multiple errors that could be triggered

sgsmob (changes requested):

JedGrabman:
@dajmcdon These changes maintain the functionality for callers into `geo_type_selector`, for existing geo_types, the functional changes for outside callers are limited to:  
1.) county and state can default to nation
2.) they may now specify that they want to strictly check the geo_type, instead of allowing it to automatically switch between county and state if incorrectly specified.

I agree that requiring a geo_type is ideal, which is why adding the auto_select parameter is worthwhile, so that we can choose when to switch between county and state, vs always allowing them to be specified as each other. 

However, if the only reason to keep the switching code is for backwards compatibility, we should remove it entirely. `evalcast-killcards` is an unreleased and breaking version change, that is explicitly unstable during development.

dajmcdon:
Ok. Let's just add the argument and quit trying to figure it out automatically. `covidcast` has county as the default. I'm tempted to switch the order to default to state since most of the uses internally have that, and there will be less breakage. But maybe better to just keep the same set as `covidcast`.

R-packages/covidcast/R/covidcast.R
JedGrabman: Fixed.

R-packages/evalcast/R/evaluate.R
JedGrabman: Removed all options per discussion.

R-packages/evalcast/R/geo_type_selector.R
```diff
@@ -1,50 +1,93 @@
-
-geo_type_selector <- function(geo_type, predictions_cards) {
+geo_type_selector <- function(geo_type, predictions_cards, auto_select = TRUE) {
```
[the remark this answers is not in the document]
JedGrabman: File removed, since we are no longer doing validation in evalcast.

R-packages/evalcast/R/geo_type_selector.R
JedGrabman: File removed.

R-packages/evalcast/R/geo_type_selector.R
JedGrabman: File removed.

R-packages/evalcast/R/geo_type_selector.R
JedGrabman: File removed.

R-packages/evalcast/tests/testthat/test-validate_geo_type.R
JedGrabman: Tests removed since we are no longer doing the validation in evalcast.

JedGrabman:
Updated to no longer handle `geo_type` validation within `evalcast`.

sgsmob (approved):

R-packages/evalcast/R/evaluate.R
```diff
@@ -67,7 +67,7 @@ evaluate_predictions <- function(
   backfill_buffer = 10,
   side_truth = NULL,
   grp_vars = c("forecaster", "forecast_date", "ahead", "geo_value"),
-  geo_type = c("county", "hrr", "msa", "dma", "state")) {
+  geo_type) {
```
dajmcdon, line 70: Can you add the defaults back in here and below? To match `covidcast::covidcast_signal()` use
```r
geo_type = c("county", "hrr", "msa", "dma", "state", "hhs", "nation")
```

Then in the body of the function use:
```r
geo_type <- match.arg(geo_type, c("county", "hrr", "msa", "dma", "state", "hhs", "nation"))
```
The result is that `geo_type` will be the first match, unless there is no match. Then it throws an error.

dajmcdon: It will also do partial matching, so and argument `geo_type = "coun"` will lead to `geo_type <- "county"` in the function body.

R-packages/evalcast/R/evaluate.R
```diff
@@ -197,8 +195,8 @@ get_covidcast_data <- function(predictions_cards,
 #'   the observed value on the date. The `quantile` and `value` columns are
 #'   dropped, as the actual value does not depend on the quantile predictions.
 #' @export
-get_actuals <- function(predictions_cards, 
-                        geo_type = c("county", "hrr", "msa", "dma", "state")) {
+get_actuals <- function(predictions_cards,
+                        geo_type) {
```
dajmcdon, line 199: Same as above.

R-packages/evalcast/R/plot_trajectory.R
```diff
@@ -38,7 +38,7 @@ plot_trajectory <- function(predictions_cards,
                             side_truth = NULL,
                             show_points = TRUE,
                             show_quantiles = TRUE,
-                            geo_type = c("county", "hrr", "msa", "dma", "state"),
+                            geo_type,
```
dajmcdon, line 41: same as above.

dajmcdon (changes requested):

R-packages/evalcast/R/evaluate.R
capnrefsmmat: You only need to do `geo_type <- match.arg(geo_type)`, no second argument, since it gets its second argument automatically from the function's default arguments

dajmcdon: You learn something new every day. Now I'll be deleting the second argument all over my hard drive. 😄

JedGrabman: Done

R-packages/evalcast/R/evaluate.R
JedGrabman: Done

R-packages/evalcast/R/plot_trajectory.R
JedGrabman: Done

dajmcdon (approved):

Issue #449, opened 2021-02-10
evalcast vignette fails


dajmcdon:
I think it's mainly outdated. First error was on `collapse_cards()`, but this is no longer used. Second error was

```r
plot_calibration(scorecards, type = "wedgeplot")
Error: In compute_actual_vs_nominal_prob: grp_vars must be present in the score_card. See details.
```

dajmcdon:
Check if related to #444

sarah-colq:
These are breaking because plot_calibration no longer works with the scorecards since they don't have quantile and value in them.

dajmcdon:
Is the vignette good to go? Like ready for the "public"?

PR #451, opened 2021-02-10
Evalcast cleanup

dajmcdon:
It looks like lots of changes, but not really. Mostly I wanted to clean up the .R files.

1. I moved functions around to files with matching names.
2. I also rebuilt the documentation and the pkgdown site (when I uncovered the vignette fail #449 ).
3. I deleted `plot_measure()`.
4. I stopped exporting `get_target_period()`
5. I exported `plot_canonical()`
6. I cleaned up the multiple/missing final lines on most files.


R-packages/evalcast/R/download_signals.R
```diff
@@ -64,7 +64,17 @@ download_signals <- function(...,
   for (i in seq_along(msg)) message(msg[i])
   
   out <- base::suppressMessages({covidcast_signals(...)})
+  empty_signals <- purrr::map_lgl(out, ~nrow(.x) == 0 || is.null(.x))
```
sgsmob, line 67: Should the conditions here be reversed for lazy evaluation purposes?  I assume `nrow(NULL)` is well defined, but it seems better to skip that.

R-packages/evalcast/R/empty_.R
```diff
@@ -0,0 +1,20 @@
+
```
sgsmob, line 1: What is the convention for the trailing underscore in the file name here?  If this is supposed to be a file that isn't user facing, the Python convention is to lead with the underscore, which I find much more readable.

R-packages/evalcast/R/get_target_response.R
```diff
@@ -52,7 +11,7 @@ get_target_response <- function(signals,
     enframe(name = NULL, value = "forecast_date") %>%
     mutate(incidence_period = incidence_period, ahead = ahead) %>%
     pmap_dfr(get_target_period)
-
+
```
sgsmob, line 14: Remove trailing spaces here and in 1 other place below.

R-packages/evalcast/R/get_target_response.R
```diff
@@ -103,17 +62,9 @@ get_target_response <- function(signals,
...
-}
-
-empty_actual <- function(){
-  out <- tibble(geo_value = character(0), forecast_date = lubridate::ymd(),
-                actual=double(0), start = lubridate::ymd(), 
-                end = lubridate::ymd()) %>% group_by(.data$geo_value)
-  attr(out, "as_of") <- Sys.Date()
-  out
-}
+}
```
sgsmob, line 70: Trailing newline

sgsmob (approved):

R-packages/evalcast/R/get_target_response.R
dajmcdon: I think I got these, but maybe check to be sure.

sgsmob: LGTM

PR #454, opened 2021-02-16
add as_of_override

dajmcdon:
add a function to override the as_of date. Easily tested locally. The `get_predictions()` example fails (since there aren't historical as_of's in the API due to bigint) unless we override.


jacobbien:
Looks good.  In the documentation I suggest specifying that `as_of_override` should return a date.

sgsmob (approved):

dajmcdon:
@jacobbien yes definitely. I just added a line.

PR #456, opened 2021-02-18
Fix linting error that snuck into main

chinandrew:
Fix
```
covidcast/covidcast.py:20:0: R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
```

### DIFF
--- a/Python-packages/covidcast-py/covidcast/covidcast.py
+++ b/Python-packages/covidcast-py/covidcast/covidcast.py
@@ -185,6 +185,8 @@ def signal(data_source: str,
         out["data_source"] = data_source
         out["signal"] = signal
         return out
+    return None
+
 
 def metadata() -> pd.DataFrame:
     """Fetch COVIDcast surveillance stream metadata.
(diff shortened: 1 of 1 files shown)


capnrefsmmat:
Hm. So in what circumstances is it possible to return `None` here? When no data is returned? (Or does that raise an error?) Because `None` isn't documented in the docstring.

If it's not supposed to ever happen, maybe we could add a comment above indicating it's unpossible

chinandrew:
> Hm. So in what circumstances is it possible to return `None` here? When no data is returned? (Or does that raise an error?) Because `None` isn't documented in the docstring.
> 
> If it's not supposed to ever happen, maybe we could add a comment above indicating it's unpossible

No data returned will reutrn none. We're not actually changing anything, since no return in python means it returns None. We're just making it explicit here for the linter. I'll update the docstring

From main right now:
```
In [13]: df = covidcast.signal("fake", "test", date(2022,1,1), date(2022,1,1))                                                                                          
/home/andrew/Documents/covidcast/Python-packages/covidcast-py/covidcast/covidcast.py:402: NoDataWarning: No fake test data found on 20220101 for geography 'county'
  warnings.warn(f"No {data_source} {signal} data found on {day_str} "

In [14]: df is None                                                                                                                                                     
Out[14]: True
```

capnrefsmmat (approved):
Looks good

PR #457, opened 2021-02-18
fix coverage plot issues

### LINKED ISSUE #444: Coverage plots do not work with default evalcast objects
Coverage plots require quantile predictions and values. Currently, evalcast stores quantile predictions in prediction cards and actual results in score cards, but there is no default object and a join is required.

Suggested Fixes

1.) Have the plots take in a predictions card object and automatically retrieve the appropriate scores
OR
2.) Have the plots take in a predictions card and score card object.

1 is nicer for Covidcast users, but wouldn't work for any users with external data.

sgsmob:
### Description
Fixes #444 by changing `plot_coverage()` to take in predictions cards and joining it with downloaded covidcast data.

### Changelist
- `plot_coverage()` expects predictions cards rather than score cards
- `plot_coverage()` expects user-specified facet row and column variables rather than inferring them
- `plot_canonical()` has a fixed typo for error handling
- Fixes an error where `ggplot` wouldn't render the image because a numeric value hadn't been cast to a factor

### DIFF
--- a/R-packages/evalcast/R/plot_coverage.R
+++ b/R-packages/evalcast/R/plot_coverage.R
@@ -37,8 +37,6 @@ plot_coverage <- function(predictions_cards,
                           coverage = 0.8, 
                           legend_position = "bottom") {
   type <- match.arg(type)
-  print(facet_rows)
-  print(facet_cols)
   if (!is.null(facet_rows)) {
    non_grouped_facet <- setdiff(facet_rows, grp_vars)
    assert_that(length(non_grouped_facet) == 0,
@@ -97,9 +95,9 @@ plot_coverage <- function(predictions_cards,
     theme(legend.position = legend_position))
 }
 
-Interaction <- function(...) {
-  params <- list(...)
-  if (length(params) == 0) return(NULL)
-  else if (length(params) == 1) return(as.factor(params[[1]]))
-  else return(interaction(...))
-}
+# Interaction <- function(...) {
+#   params <- list(...)
+#   if (length(params) == 0) return(NULL)
+#   else if (length(params) == 1) return(as.factor(params[[1]]))
+#   else return(interaction(...))
+# }
--- a/R-packages/evalcast/R/plot_coverage.R
+++ b/R-packages/evalcast/R/plot_coverage.R
@@ -29,9 +29,25 @@ plot_coverage <- function(predictions_cards,
                           type = c("all", "one"), 
                           grp_vars = c("forecaster", "forecast_date", "ahead"),
                           avg_vars = c("geo_value"),
+                          facet_rows = c("forecaster"),
+                          facet_cols = c("forecast_date"),
                           coverage = 0.8, 
                           legend_position = "bottom") {
   type <- match.arg(type)
+  print(facet_rows)
+  print(facet_cols)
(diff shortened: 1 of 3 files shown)


R-packages/evalcast/R/plot_coverage.R
```diff
@@ -40,17 +58,17 @@ plot_coverage <- function(scorecards,
     
     
     if (ngrps == 3L) {
-      g <- g + geom_line(aes(color = !!sym(grps[1]))) +
+      g <- g + geom_line(aes(color = as.factor(!!sym(grps[1])))) +
```
dajmcdon, line 61: Starting here is the cruft that I'd like to get rid of. Can you follow the logic in `plot_canonical` (see [here](https://github.com/cmu-delphi/covidcast/blob/256944ce6bdccf83b0878997bc6b72179f333d89/R-packages/evalcast/R/plot_canonical.R#L74) for example) and get rid of this automatic choice garbage?

@JedGrabman assuming such a change doesn't screw up the eval dashboard too much.

dajmcdon (changes requested):

R-packages/evalcast/R/plot_coverage.R
JedGrabman: The eval dashboard uses data processing from evalcast, but has its own frontend, so updating plot functions shouldn't affect it.

sgsmob: Do you want me to change the signature to use `facet_rows` and `facet_cols` arguments like `plot_canonical()` does?  Or is there something different?

dajmcdon: Yeah, I think change the signature. So faceting, and colors exactly as done in `plot_canonical()`.

sgsmob: Done.

R-packages/evalcast/R/plot_coverage.R
```diff
@@ -1,91 +1,105 @@
...
+    facet_layer +
+    theme_bw() +
+    theme(legend.position = legend_position))
+}
+
+Interaction <- function(...) {
+  params <- list(...)
+  if (length(params) == 0) return(NULL)
+  else if (length(params) == 1) return(as.factor(params[[1]]))
+  else return(interaction(...))
```
dajmcdon, line 104: I don't think we need this since it's in `plot_canonical.R`

dajmcdon (approved):
Thanks!!

If you can just make the 1 minor change, it's approved.

R-packages/evalcast/R/plot_coverage.R
```diff
@@ -1,91 +1,105 @@
...
                           type = c("all", "one"), 
                           grp_vars = c("forecaster", "forecast_date", "ahead"),
                           avg_vars = c("geo_value"),
+                          facet_rows = c("forecaster"),
+                          facet_cols = c("forecast_date"),
                           coverage = 0.8, 
                           legend_position = "bottom") {
   type <- match.arg(type)
-  # make sure scorecards are comparable:
+  print(facet_rows)
```
sarah-colq, line 40: print statement shouldn't be here

Issue #447, opened 2021-02-08
covidcast_signals() returns unrequested NAs


dajmcdon:
In `evalcast` development, we’ve run into a problem with `covidcast_signals()` (or perhaps `aggregate_signals()`) that we’ve been unable to reproduce. However it just happened to me.

```r
others <- suppressMessages(
  covidcast_signals(
    data_source = c("fb-survey", "indicator-combination"),
    signal = c("smoothed_hh_cmnty_cli", "nmf_day_doc_fbc_fbs_ght"),
    geo_type = "county", 
    start_day = "2021-01-15")) %>%
  aggregate_signals(format = "long")
others %>% select(data_source, signal) %>% distinct()
```

The result should be a 2 row tibble basically returning my `data_source` and `signal` arguments. However, I get this:

```
            data_source                  signal
1             fb-survey   smoothed_hh_cmnty_cli
2                  <NA>                    <NA>
3 indicator-combination nmf_day_doc_fbc_fbs_ght
```

Running without the `aggregate_signals()` call seems ok. After `aggregate`, I have 560 rows with `NA` for `data_source`.

dajmcdon:
I just had this happen again during forecaster production run tonight. I ran the following:
```r
states <-  suppressMessages(
+   covidcast_signals(
+       data_source = covidcast_data_source, 
+       signal = covidcast_signal,
+       geo_type = "state", 
+       start_day = start_date,
+       end_day = Sys.Date())
+ )
Warning messages:
1: Data not fetched for the following days: 2021-02-12, 2021-02-13 
2: Data not fetched for the following days: 2021-02-03, 2021-02-12, 2021-02-13 
> 
> # Aggregating signals, long format
> states_all <- aggregate_signals(states, format = "long") 
> any(is.na(states[[1]]$signal))
[1] FALSE
> any(is.na(states[[2]]$signal))
[1] FALSE
> any(is.na(states_all$data_source))
[1] TRUE
> sum(is.na(states_all$data_source))
[1] 203
> states_all[which.max(is.na(states_all$data_source))]
Error in `[.data.frame`(states_all, which.max(is.na(states_all$data_source))) : 
  undefined columns selected
> states_all[which.max(is.na(states_all$data_source)),]
      data_source signal geo_value time_value issue lag stderr sample_size dt value
17630        <NA>   <NA>        ak 2021-02-03  <NA>  NA     NA          NA  0    NA
> sum(is.na(states_all$signal))
[1] 203
```

[...]

PR #459, opened 2021-02-21
fix annoying NAs created by aggregate_signals()

dajmcdon:
Should fix issue #447. See there for details of the problem. 

* The solution is to simply set both `data_source` and `signal` inside the `x_all` object in `apply_shifts_one()`. Then do the `full_join()` on those columns as well. 

* There's also a test to ensure that this doesn't happen again.

* Note: there remains an `<NA>` in the issue column.

Finally, it may be worth considering that if `dt=NULL` and `format="long"`, we should maybe avoid all processing and just do `bind_rows()` on the list of signals.


capnrefsmmat (approved):
Looks good to me.

sgsmob (approved):

PR #460, opened 2021-02-21
scale_by_forecaster behavior

dajmcdon:
In the R check, there are a few warnings that this avoids.

* But more importantly, for some reason, scaling seems to ignore any columns that come after the `base_forecaster` in the `df_list`. So we move it to the end. I don't understand why this is the case.

### DIFF
--- a/R-packages/evalcast/R/scale_by_forecaster.R
+++ b/R-packages/evalcast/R/scale_by_forecaster.R
@@ -47,7 +47,7 @@ scale_by_forecaster <- function(score_card,
     for (var in score_cols){
         base_values <- filter(score_card,
                               .data$forecaster == base_forecaster_name)[[var]]
-        if (any(base_values <= 0)) {
+        if (any(base_values == 0) & !is.na(base_values)) {
             warning("scale_by_forecaster will divide by zero in column ", var)
         }
     }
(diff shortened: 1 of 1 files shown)


sgsmob (approved):
I don't understand why this is necessary but it looks like it should be safe

R-packages/evalcast/R/scale_by_forecaster.R
```diff
@@ -47,17 +47,21 @@ scale_by_forecaster <- function(score_card,
     for (var in score_cols){
         base_values <- filter(score_card,
                               .data$forecaster == base_forecaster_name)[[var]]
-        if (!all(base_values > 0)) {
+        if (any(base_values <= 0)) {
```
JedGrabman, line 50: Shouldn't this just be `==`? I'm not sure what score metric would allow negative values, but that should be a separate warning if we want it.

JedGrabman (approved):
Also not sure why this changes things, but no objection if it fixes the issue.

One minor comment, but don't need to re-review.

R-packages/evalcast/R/scale_by_forecaster.R
dajmcdon: I think either would be fine. I switched `!all` to `any` because the first will fail if there's an `NA`. But it turns out that the second fails as well.

PR #462, opened 2021-02-22
Plot_calibration updates

sarah-colq:
Update plot_calibration and compute_calibration to use predictions_cards instead of scorecards.   Update using facet_layer and color_layer scale instead of case coding for plots (as done in PR #457).  Fixes some issues in intro vignette (see issue #449 ).


R-packages/evalcast/R/plot_calibration.R
```diff
@@ -1,27 +1,53 @@
...
+      facet_cols <- c(grps[1])
+      facet_rows <- NULL
+    } else if (length(grps) == 2) {
+      facet_cols <- c(grps[2]) 
+      facet_rows <- c(grps[1])
+    } else {
+      facet_rows <- NULL
+      facet_cols <- NULL
+    }
+
```
dajmcdon, line 43: I'd like to get rid of as much of this automatic faceting as possible. Instead put it in the function call the same as `plot_canonical()`.

R-packages/evalcast/R/plot_calibration.R
```diff
@@ -40,43 +66,48 @@ plot_calibration <- function(scorecard,
...
+      facet_cols <- c(grps[2])
+      facet_rows <- NULL
+    } else if (length(grps) == 3) {
+      facet_cols <- c(grps[3]) 
+      facet_rows <- c(grps[2])
+    } else {
+      facet_rows <- NULL
+      facet_cols <- NULL
+    }
+
```
dajmcdon, line 86: Same deal here.

dajmcdon (changes requested):

R-packages/evalcast/R/plot_calibration.R
sarah-colq: I tried doing that, but we would want grouping to be done differently for traditional and wedge plots in terms of facets and colors, so I don't know how we could have default arguments that would work for both.

dajmcdon: Three options then: 
1. Make wedge plots their own function that handles single forecasters only.
2. Change the wedge-plot colors to be only a single color and then allow multiple forecasters and faceting as with everything else.
3. Do both or maybe add a case to `plot_calibration()` to do either option 1 or 2 based on how many forecasters there are.

sarah-colq: Another idea, how about we leave it as is but have two wrapper functions for wedge plots and traditional plots?  Then we pass in ... to the plot_calibration, and have default values for the wedge plots and traditional plots for the facet_rows and facet_cols?   Then it's the same behavior in terms of using the function, just some internal refactoring to have default values for each.

sarah-colq: I did the idea I mentioned, lmk what you think.

R-packages/evalcast/R/compute_calibration.R
```diff
@@ -21,10 +22,19 @@
...
 compute_calibration <- function(
-  score_card, 
+  predictions_cards,
+  geo_type,
   grp_vars = c("forecaster", "forecast_date", "ahead"),
   avg_vars = c("geo_value")) {
   
+  score_card <- left_join(predictions_cards,
+                          get_covidcast_data(predictions_cards,
+                                             backfill_buffer,
```
dajmcdon, line 32: I don't think `backfill_buffer` exists. This will throw an error.

R-packages/evalcast/R/plot_calibration.R
```diff
@@ -1,82 +1,39 @@
...
 #'
-#' @param scorecard Single score card.
+#' @param predictions_cards Either one predictions_card or several joined
+#'   together using bind_rows().
+#' @template geo_type-template 
 #' @param type One of "wedgeplot" or "traditional".
 #' @param grp_vars variables over which to compare calibration. These
 #'   determines the color of the lines and faceting depending on `type`
 #' @param avg_vars variables over which we average to determine the calibration.
 #' @param legend_position Legend position, the default being "bottom".
```
dajmcdon, line 10: You'll need to document the `...`.

R-packages/evalcast/R/plot_calibration.R
```diff
@@ -98,3 +55,108 @@ setup_wedgeplot <- function(calib) {
...
+                            non_grouped_facet))
+  }
+  if (!is.null(facet_cols)) {
+    non_grouped_facet <- setdiff(facet_cols, grp_vars)
+    assert_that(length(non_grouped_facet) == 0,
+                msg = paste("Variables must be grouped in order to be faceted in cols:",
+                            non_grouped_facet))
+  }
+  
+  facet_layer <- facet_grid(rows = vars(!!!syms(facet_rows)),
```
dajmcdon, line 89: Do we want this to always be a grid instead of `facet_wrap`?

R-packages/evalcast/R/plot_calibration.R
```diff
@@ -98,3 +55,108 @@ setup_wedgeplot <- function(calib) {
...
+    labs(x = "Quantile level",
+         y = "Proportion",
+         title = sprintf("%s (ahead %s): Calibration", 
+                         .data$name, .data$ahead)) +
+    geom_line(aes(color = .data$color, group = .data$color)) +
+    scale_color_viridis_d() +
+    facet_layer 
+  
+  return(g)
+}
```
dajmcdon, line 162: need an empty line here.

R-packages/evalcast/R/plot_calibration.R
```diff
@@ -98,3 +55,108 @@ setup_wedgeplot <- function(calib) {
...
+                            non_grouped_facet))
+  }
+  if (!is.null(facet_cols)) {
+    non_grouped_facet <- setdiff(facet_cols, grp_vars)
+    assert_that(length(non_grouped_facet) == 0,
+                msg = paste("Variables must be grouped in order to be faceted in cols:",
+                            non_grouped_facet))
+  }
+  
+  facet_layer <- facet_grid(rows = vars(!!!syms(facet_rows)),
```
dajmcdon, line 144: Same as above.

dajmcdon (changes requested):
I think the `...` version is not quite the right thing. But I don't mind splitting it out like this. I think it would be better if we list the arguments to `plot_calibration` specifically and ignore them if they're inappropriate.

R-packages/evalcast/R/plot_calibration.R
sarah-colq: I like facet_grid so that you can sort visually across different categories, is there a reason to use facet_wrap instead?

dajmcdon: If you facet on only one category, then wrap works better.

dajmcdon: I think this is fine. Did the `...` get fixed?

sarah-colq: The documentation or removing it?

dajmcdon: replacing it with `facet_rows` and `facet_cols` explicitly.

sarah-colq: Done.  If they're both null they just use the defaults for each plot type.

dajmcdon (approved):

Issue #466, opened 2021-02-25
wis calculations should not require 0.5 quantile


JedGrabman:
We should be able to calculate wis on arbitrary sets of quantiles, as long as they valid (i.e. the associated values are monotonic).
Currently, we return NA if there is not a 0.5 quantile.

This should be fixed, since 1.) there are some forecasters that don't submit 0.5 quantiles and 2.) We have a long-term goal for this package to be useable by arbitrary forecasters.

PR #467, opened 2021-02-26
Evalcast wis components

dajmcdon:
1. add functions to compute the components of WIS needed by the visualization. 
2. Assure that our calculations match those in `epiforecasts::scoringutils` as used by the hub.
3. Fix a bug in `scale_by_forecaster()` that I introduced in #460 .
4. Remove need for WIS to have a median described in #466 .


R-packages/evalcast/R/error_measures.R
```diff
@@ -99,6 +99,94 @@ interval_coverage <- function(coverage) {
...
+#' @export
+overprediction <- function(quantile, value, actual_value) {
+  score_func_param_checker(quantile, value, actual_value, "overprediction")
+  if (!is_symmetric(quantile)) {
+    warning(paste0("overprediction/underprediction/sharpness require",
+                   "symmetric quantile forecasts. Using NA."))
+    return(NA)
+  }
+  if (all(is.na(actual_value))) return(NA)
+  actual_value <- unique(actual_value)
```
sgsmob, line 120: Why are we going anything if `actual_value` contains multiple values?

R-packages/evalcast/R/error_measures.R
```diff
@@ -99,6 +99,94 @@ interval_coverage <- function(coverage) {
...
+  
+  if (length(med) > 1L) return(NA)
+  m <- ifelse(length(med == 1L),
+              (med - actual_value) * (med > actual_value),
+              NULL)
+  
+    
+  
+  ans <- mean(c(
+    rep((lower - actual_value) * (lower > actual_value), 2), m))
```
sgsmob, line 133: Why are we weighting the predicted values double over the median?

R-packages/evalcast/tests/testthat/test-err-measures.R
```diff
@@ -74,3 +74,115 @@ test_that("Error when actual_value is the wrong length", {
...
+  # test_data <- data.frame(true_value =   rep(c(1, -15, 22), times = 3),
+  #                         quantile = rep(c(0.25, 0.5, 0.75), each = 3),
+  #                         prediction = c(c(0, 1, 0), c(1, 2, 3), c(2, 2, 3)),
+  #                         model = c("model1"),
+  #                         date = rep(1:3, times = 3))
+  # 
+  # eval <- scoringutils::eval_forecasts(
+  #   test_data,
+  #   interval_score_arguments = list(count_median_twice = FALSE))
+  # exp_wis <- eval$interval_score
```
sgsmob, line 131: Is this code supposed to be used?  Would prefer not to check in commented out code

R-packages/evalcast/tests/testthat/test-err-measures.R
```diff
@@ -74,3 +74,115 @@ test_that("Error when actual_value is the wrong length", {
...
+test_that("wis over/under works, 2 intervals and median", {
+  
+  test_data <- data.frame(true_value =   rep(c(1, -15, 22), times = 5),
+                          quantile = rep(c(0.1, 0.25, 0.5, 0.75, 0.9), each = 3),
+                          prediction = c(c(-1, -2, -2), c(0, 1, 0), c(1, 2, 3),
+                                         c(2, 2, 3), c(3, 4, 4)),
+                          model = c("model1"),
+                          date = rep(1:3, times = 5))
+  
+  # eval <- scoringutils::eval_forecasts(
```
sgsmob, line 169: Same

R-packages/evalcast/R/error_measures.R
dajmcdon: We don't. Multiple values get killed by `score_func_param_checker()` on 113.

R-packages/evalcast/R/error_measures.R
dajmcdon: This is just how WIS gets calculated. Essentially counted once for the lower bound and once for the upper bound. But the median gets only once.

R-packages/evalcast/tests/testthat/test-err-measures.R
dajmcdon: Not used, but I took this from the other package that does these calculations for the Hub. This is to ensure that our scores are the same as theirs. I think it's good to have it in case. But I didn't want to import their package just for the tests.

sgsmob: Can you pull this out into a separate test that is skipped or otherwise ignored then?

dajmcdon: done

sgsmob (approved):

PR #469, opened 2021-03-01
Update plot_width

sarah-colq:
Add facet_vars and color_vars instead of the ngrps checking.  Add facet_cols and facet_rows as default.   Add a check for computing the width for entries with NA values for quantiles (not point estimates) and filter those out as well as warning the user.


R-packages/evalcast/R/plot_width.R
```diff
@@ -73,11 +86,18 @@ compute_width <- function(cards,
...
+
   cards <- cards %>% 
     filter(!is.na(.data$quantile)) %>%
     select(all_of(grp_vars), all_of(avg_vars),
            .data$quantile, .data$value)
   
+  if(any(is.na(cards$value))) {
+    message("Some quantiles values have NA values; these will be removed for
+            the interval width plots.")
+    cards <- cards %>% filter(!is.na(.$value))
```
dajmcdon, line 98: I think this needs to be `filter(!is.na(.data$value))`

dajmcdon (approved):

Issue #468, opened 2021-02-28
[evalcast-killcards] `plot_canonical()` not calling `scale_by_forecaster()` with correct `id_cols` after scaling.


huisaddison:
Calling `plot_canonical()` with a `base_forecaster` defaults to `scale_before_aggr=FALSE`.  But within `plot_canonical()` after aggregation, the only columns are `forecaster`, `ahead`, and the error metric.  `scale_by_forecaster()`'s default `id_cols` assume many more than that:

```
> plot_canonical(evals, x = "ahead", y = "wis", aggr = mean,                                   
+                base_forecaster = "Baseline", subtitle = subtitle,                            
+                xlab = "Days ahead", ylab = "Relative mean WIS"                               
+   ) + geom_hline(yintercept=1.0, linetype='dashed')                                          
Error: id_cols contains columns geo_value, forecast_date, data_source, signal, target_end_date, incidence_period not present in the columns of score_card
```

Proposed fix:  add `id_cols=c('forecaster', 'ahead')` to the function call here: https://github.com/cmu-delphi/covidcast/blob/evalcast-killcards/R-packages/evalcast/R/plot_canonical.R#L67-L69

Or better yet, understand what columns are used for aggregation and input them directly, rather than hardcoding.

PR #476, opened 2021-03-04
Update plot_canonical to pass appropriate columns to scale_by_forecaster

JedGrabman:
Fixes issue #468


JedGrabman:
Please double check my logic.

* `drop_na` shouldn't drop all rows with NAs, because it does not cause issues for most columns to have a value of NA. Really just `x` and `y` that cause problems.
* `id_cols` needs to contain `group_vars` (because that is what identifies different groups) and `x` (since it's implicitly grouped because different values need to be differentiated on the x-axis).

dajmcdon (approved):
Looks good to me. @huisaddison look alright?

PR #479, opened 2021-03-06
Simplify use of evaluation functions

JedGrabman:
There are 2 major changes:

1.)  `get_covidcast_data` has been rewritten so that it can take in predictions with different (`data_source`, `signal`, `ahead`, `incidence_period`) combinations. This makes it easier to use (since it reduces the amount of pre-processing necessary to use the function on cards from `get_predictions`).

2.) `evaluate_predictions` has been refactored, leading to the creation of `evaluate_covid_predictions` for covid specific forecasts. This simplifies the function quite a bit for non-covid forecasters (since the documentation is more concise and relevant and has better defaults). Covid-specific forecasts can still be evaluated using `evaluate_covid_predictions`, which is also simpler than the previous function, as it no longer needs the side-truth functionality and can also focus documentation entirely on covid forecasts / predictions_cards.

Documentation has been updated to reflect the changes (and to address some other outdated info).


R-packages/evalcast/R/evaluate_predictions.R
```diff
@@ -1,103 +1,51 @@
...
-#'   `predictions_cards()` with additional columns for each err_measure and
-#'   for the truth (named `actual`). All the columns may be necessary for
-#'   some types of plotting, but it may be useful to shrink this down
+#'   `predictions_cards()` with additional columns for each `err_measure` and
+#'   for the truth (named `actual`).
 #'
 #' @export
 evaluate_predictions <- function(
   predictions_cards,
+  truth,
```
dajmcdon, line 36: Can we call this something like `truth_data` maybe? Something to emphasize that it needs to be a df with some columns.

Maybe also default to the truth being named "actual" with an option to pass in other names?

R-packages/evalcast/R/evaluate_predictions.R
```diff
@@ -1,103 +1,51 @@
-#' Create a data frame of score cards
+#' Create a score card data frame
 #'
-#' Performs backtesting, through the following steps:
+#' Performs testing, through the following steps:
```
dajmcdon, line 3: Maybe this should say "evaluates the quality of a forecaster" rather than "performs testing".

R-packages/evalcast/R/evaluate_predictions.R
```diff
@@ -116,67 +64,130 @@ evaluate_predictions <- function(
...
+#'   (i.e., observed) returns a scalar measure of error. Null or an empty list
+#'   may be provided if scoring is not desired.
+#' @param backfill_buffer How many days until response is deemed trustworthy
+#'   enough to be taken as correct? See details for more.
+#' @template geo_type-template 
+evaluate_covid_predictions <- function(predictions_cards,
+                                        err_measures = list(wis = weighted_interval_score,
+                                                            ae = absolute_error,
+                                                            coverage_80 = interval_coverage(coverage = 0.8)),
+                                        backfill_buffer = 10,
```
dajmcdon, line 113: Let's make `backfill_buffer = 0` the default. That's what we always use.

dajmcdon (changes requested):
I'd like to avoid merging this for the time being. There's a variety of documentation and scripts elsewhere that relies on the `evaluate_predictions()` call. My thinking is that, once @sarah-colq finishes the plotting cleanup, I'll rebase `evalcast` on `evalcast-killcards`. Then we'll do this and one other "breaking" change on `evalcast` before merging to `main`.

R-packages/evalcast/R/evaluate_predictions.R
JedGrabman: Updated the object to `truth_data`. Didn't touch the "actual" column name code. Seems large and distinct enough to warrant a separate PR.

R-packages/evalcast/R/evaluate_predictions.R
JedGrabman: Done

R-packages/evalcast/R/evaluate_predictions.R
JedGrabman: Done

JedGrabman:
Updated branch to evalcast, now that we are no longer actively developing `evalcast-killcards`

dajmcdon (changes requested):
These changes will need to propagate to the vignette too right?

JedGrabman:
Updated the vignette (which also meant pulling in changes from PR #488)

dajmcdon (approved):

PR #484, opened 2021-03-09
Add documentation examples.

kjytay:
Examples added for slide_by_geo(), pct_change() and estimate_deriv(). Users can get short code snippets from the help file instead of going to the vignettes.


huisaddison (approved):
LGTM, thanks for adding this documentation!

PR #485, opened 2021-03-10
Evalcast killcards

dajmcdon:
Step 1 of `evalcast` release. We will use the evalcast branch as the development branch going forward. `evalcast-killcards` will no longer be touched so that breaking changes won't have downstream implications.


dajmcdon:
Merge `main` into `evalcast` before `killcards` onto `evalcast`?

dajmcdon:
This is ugly unfortunately. I'm going to have to do it slightly differently.

PR #490, opened 2021-03-15
Typo / bug fix for get_covidhub_forecaster_names

JedGrabman:
This fixes a typo that was breaking `get_covidhub_forecaster_names`


dajmcdon (approved):
I'm not going to merge this yet. The check is failing for what seems to be an unrelated reason, but I'd like it to pass.

Issue #486, opened 2021-03-10
Calibration plots do not facet on grouping variables


JedGrabman:
In `format_wedgeplot` (in `plot_calibration`), a user may specify `grp_vars`, `facet_rows` and `facet_cols`. If the user does specify  a value for `grp_vars`, but doesn't specify anything for `facet_rows` or `facet_cols`, it results in a rather confused plot, for example:

![Rplot](https://user-images.githubusercontent.com/7600154/110713485-057a1680-81d0-11eb-8559-f2fd3ee6dcbe.png)

I think the problem stems from `plot_calibration`. Previously, if a user specified grouping variables, it would [auto-determine the faceting](https://github.com/cmu-delphi/covidcast/blob/8a78de916bc0f0c83c9a22b710b5245d384a5380/R-packages/evalcast/R/plot_calibration.R). Now it does not.

Expected behavior:
If both faceting variables are NULL, attempt to auto determine faceting values
If a faceting variable is not null, check that the faceting is reasonable.

I think wedge plots can reasonably facet on 2 variables (one per row and column), while traditional plots can facet on 3 (color, row and column).

It might be a good idea to update `test_legal_faceting` (within `plot_coverage`) to take a maximum groups argument as well.

JedGrabman:
@sarah-colq Can you take a look at this?

sarah-colq:
@JedGrabman Do you remember what you specified for `grp_vars` here?

JedGrabman:
I think I was using 
grp_vars = c("ahead")
avg_vars = c("geo_value", "forecast_date")
on a predictions_card.

sarah-colq:
So in changes have been made since then, where there are not default values for facet_rows and facet_cols for each type of plot (which was the behavior we specified before).  Then it fails on test_legal_faceting.   @dajmcdon do we still want default values for the calibration plots?

sarah-colq:
Okay after discussion with @JedGrabman we have decided on hardcoded values for each calibration plot type and improve the test_legal_faceting error message.

PR #491, opened 2021-03-15
Restore default facet_rows and facet_cols for plot_calibration

sarah-colq:
Set default values so that the faceting is properly tested when no `facet_rows` or `facet_cols` values are given by the user.  Improve `test_legal_faceting` error message to give directions on what to do.  Fixes issue #486


R-packages/evalcast/R/plot_coverage.R
```diff
@@ -90,14 +90,18 @@ test_legal_faceting <- function(facet_rows, facet_cols, grp_vars) {
...
+                  non_grouped_facet, ". Group ", non_grouped_facet,
+                  " or change the value of facet_rows to something in grp_vars."))
   }
   if (!is.null(facet_cols)) {
     non_grouped_facet <- setdiff(facet_cols, grp_vars)
     assert_that(length(non_grouped_facet) == 0,
-                msg = paste("Variables must be grouped in order to be faceted in cols:",
-                            non_grouped_facet))
+                msg = paste0(
+                  "Variables must be grouped in order to be faceted in rows: ",
```
JedGrabman, line 102: columns, not rows.

R-packages/evalcast/R/plot_calibration.R
```diff
@@ -127,6 +132,11 @@ format_traditional_calib_plot <- function(calib,
                                           facet_rows,
                                           facet_cols){
   
+  if (is.null(facet_rows) && is.null(facet_cols)){
+    facet_rows <- c("ahead")
```
JedGrabman, line 136: Do we need to facet on both by default? Since the traditional ones can plot multiple lines, I think I've seen them facet on one fewer variable then the wedge plots.

R-packages/evalcast/R/plot_coverage.R
```diff
@@ -90,14 +90,18 @@ test_legal_faceting <- function(facet_rows, facet_cols, grp_vars) {
   if (!is.null(facet_rows)) {
     non_grouped_facet <- setdiff(facet_rows, grp_vars)
     assert_that(length(non_grouped_facet) == 0,
-                msg = paste("Variables must be grouped in order to be faceted in rows:",
-                            non_grouped_facet))
+                msg = paste0(
+                  "Variables must be grouped in order to be faceted in rows: ",
+                  non_grouped_facet, ". Group ", non_grouped_facet,
+                  " or change the value of facet_rows to something in grp_vars."))
```
JedGrabman, line 96: Suggest rephrasing to something like. "The following variables... ." Also change "something" to "a subset" or something similar.

JedGrabman (changes requested):

R-packages/evalcast/R/plot_calibration.R
sarah-colq: The default `grp_vars` have three variables, so the default plots one var for rows, one for cols, and one for colours.  This is the behavior that existed originally.

dajmcdon (changes requested):
see above

sarah-colq:
Feedback addressed in pull #505 .

Issue #470, opened 2021-03-02
Questions and Roadmap for covidcast on CRAN


statsmaths:
Following discussion last week, I took a close look at the current R covidcast package. I tried to review it the way I would a submission for the RJournal, JOSS, ROpenSci or equivalent, with the caveat that I know it's still in the works. Overall it seems to run well, be well documented, and has the core functionality I would expect.

Here are some suggestions for how to streamline the package for getting it onto CRAN. I am sure some of these are bad ideas, but though I would err or the side of too many ideas rather than too few:

1. Standardize all of the `=` assignments to `<-`. 
2. The dplyr/tidyr/purrr functions are used in a non-trival way, so I think they are fine even though they introduce a lot of dependencies. But, I recommend removing pipes. Great for interactive work but hard to debug, particularly deep within a package. This should not be a heavy lift.
3. Clean up the format and number of global variables. Some all ALL_CAPS, others are not. Most seem like they are only used once and could just be defined within the scope of a function.
4. Try to see if some of the Imports that we are only using one or two function from are really needed. A few seemed easily replacable.
5. The metadata files are a bit too big and contain a lot of variables. Can we limit these to just a few for the package? Only a few are documented in the package anyway. I don't think they will pass CRAN with the current sizes.
6. There should be a more complete README.md for the package with a simple worked out example.
7. The vignettes are nice and very informative, but take far too long to build. Should we replicate them with static data, remove the time-consuming bits, or provide pre-built HTML files to CRAN?
8. Is there any possibility of keeping the R package in its own repository? This one is huge (as you know) and one-repo-per-package is defintiely the norm. This isn't needed at all for CRAN, and might have other drawback, but just thought I would ask as it would certainly simplify some things, particularly as the project becomes more maintainence rather than active development.
9. I think the biggest challenge is going to be supporting the plotting, particularly the spatial maps. What's the thought on the extent to which this code and data needs to be in the package? .... My usually preference is to have a package provide well-formated data and produce plots myself (full control and no need to learn a bunch of other plotting APIs), but I know that's not everyones preference. At any rate, the shape files are just too big for CRAN and we need some alternative.

That's all I have for now. Once I have a sense of how to address these, happy to take them on as soon as possible.

capnrefsmmat:
> The metadata files are a bit too big and contain a lot of variables. Can we limit these to just a few for the package? Only a few are documented in the package anyway. I don't think they will pass CRAN with the current sizes.

[...]

PR #493, opened 2021-03-16
Package changes for CRAN

statsmaths:
This commit address the major issues in #470. It has compressed shapefiles that fit under the CRAN limits, has a cached version of the slowest vignette, removes extra metadata columns, and adds `.data$` to silence the warning about tidy evaluation within a package.


R-packages/covidcast/R/cor.R
```diff
@@ -2,8 +2,7 @@
 #'
 #' Computes correlations between two `covidcast_signal` data frames, allowing
 #' for slicing by geo location, or by time. (Only the latest issue from each
-#' data frame is used for correlations.) See the [correlations
-#' vignette](https://cmu-delphi.github.io/covidcast/covidcastR/articles/correlation-utils.html)
```
sgsmob, line 6: why remove this?

R-packages/covidcast/R/cor.R
```diff
@@ -21,46 +20,53 @@
...
   # Perform time shifts, then compute appropriate correlations and return
-  return(z %>% dplyr::group_by(geo_value) %>% # group by geo value
-         dplyr::arrange(time_value) %>%  # sort rows by increasing time
-         dplyr::mutate(value.x = shift(value.x, n = dt_x), # shift values
-                       value.y = shift(value.y, n = dt_y)) %>%
-         dplyr::ungroup() %>% # get rid of grouping by geo value
-         dplyr::group_by(dplyr::across(dplyr::all_of(by))) %>% # group by what's specified
-         dplyr::summarize(value = cor(x = value.x, y = value.y, # compute cor
-                                      use = use, method = method)))
+  z <- dplyr::group_by(z, .data$geo_value)
```
sgsmob, line 55: Does the warning require that we unchain these lines?  I find it to be much more understandable with the chaining.

R-packages/covidcast/vignettes/correlation-utils.Rmd.orig
```diff
@@ -0,0 +1,195 @@
+---
```
sgsmob, line 1: Is this file meant to be included in this PR?

sgsmob (commented):
Mostly looks good but a few questions of intent

capnrefsmmat:
It looks like the plotting tests failed, probably because the compressed GeoJSON has slightly different coordinates and so the plots changed slightly. Can you run `vdiffr::manage_cases()` to visually check that nothing broke, then commit the updated test files by blessing the changes?

capnrefsmmat:
This is looking great, though, so I'll take a look tomorrow for more review.

statsmaths:
Thanks! Will take a look tomorrow. Once I verify that the plots are okay and just a result of the changed shapefiles, will re-commit. I was running with `--as-cran` and did not realize some tests were skipped.

capnrefsmmat:
Yeah, it looks like vdiffr doesn't report failures on CRAN: 

https://github.com/r-lib/vdiffr/blob/be3c1b08aef0480deb3d7f11eaff1ee44dcdc072/R/testthat-ui.R#L47-L51

R-packages/covidcast/R/cor.R
statsmaths: I meant to replace it with a link to \code{vignette("correlation-utils", package = "covidcast")} to link to the actual local vignette, but forgot to do that. Fixed in the updated, good catch!

R-packages/covidcast/R/cor.R
statsmaths: I just generally tried to avoid pipes in package because they can be hard to debug (the traceback gives the entire pipe), particularly when users are submitting bugs that may be hard to replicate without all of their data. I did not have time to do it everyone in the package, but did here when I was fixing the other warning.

R-packages/covidcast/vignettes/correlation-utils.Rmd.orig
statsmaths: Yes, see comment #470. This is the "raw" code that creates a pre-build correlation-utils.Rmd to speed up the installation time. It needs an odd extension so that R does not try to build it on installation.

sgsmob (approved):

statsmaths:
@capnrefsmmat The svg files change slightly with the new shapefiles (obviously, I guess, not sure how they couldn't). I committed those but am still getting an error in the autocheck. Everything is passing locally, and it's only the plots that are throwing an error. Have you had any trouble with this in the past? Any possibility that this could be a slight difference in the plot based on the OS that is being detected?

statsmaths:
So, could be this issue:

   https://github.com/r-lib/vdiffr/issues/86

I am running 4.0.3 but the test is on a version that is over 2 years old (3.6). Is there a reason we are testing on R 3.6 and not the new 4.0 branch?

capnrefsmmat:
Yeah, that would explain why your commit doesn't fix the issue. We have the package listed as being compatible with >= 3.5.0, and in the past we've had some compatibility issues between the two that suggest we should run CI on multiple versions. For example, R 4.0.0 disabled `stringsAsFactors`, and code written assuming it's disabled broke on older R versions.

Supporting at least 3.6 is important because recent Linux distributions still use 3.6 -- e.g. Ubuntu 20.04 distributes 3.6.3.

So I think the resolution here is

1. Decide the minimum R version we want to support. (I'm not sure why we picked 3.5.0 and not 3.6.0.)
2. Set one version (maybe 4.0.0) as the standard for vdiffr and set those tests (but not the other tests) to be skipped on other R versions, as suggested in #350 and implemented in #356.
3. Regenerate plots using that version.

statsmaths:
Perfect, works now. Are you okay with merging now into r-pkg-devel?

capnrefsmmat:
Looks good. Just two questions:

1. I see your `skip_if` also checks the `$minor` part of the version, which e.g. means that the tests won't run on my machine (4.0.4). Should we pull the condition out into a function so all the `skip_if` calls use the same condition and we can easily change it when we figure out a better CI strategy here?
1. Should the process for going from .Rmd.orig -> Rmd be documented in `DEVELOP.md`? Right now I don't know how the pre-compiled vignette works or how I'd make changes to the vignette.

statsmaths:
I suppose just checking 4.x is fine for now since CI is 3.6 and in theory the 4.0 releases should be consistent. Changed those and added a line to the DEVEL.md page (you just run the `precompile.R` script in the vignettes directory. You can let the checks re-run and then hopefully merge.

capnrefsmmat:
I take it `precompile.R` was meant to be committed but got lost in the process?

statsmaths:
Ah, that would explain the confusion! There's a "*.R" entry in the vignettes ".gitignore" file. Just added.

capnrefsmmat (approved):
Looks good to merge now. Let's file a follow-up issue for changing the CI to run on R 4.0.0 (either replacing 3.6 or do both), but that doesn't need to block this

PR #494, opened 2021-03-17
Add a function for creating testing and training matrices

### LINKED ISSUE #478: Create X, y matrix function
Creates the `train_X`, `train_Y`, and `test_X` matrices from the lagged data frame and an associated ahead.

sgsmob:
Add a function to take a lagged data frame and produce the corresponding X and y matrices to be passed into the models.

Fixes #478.


R-packages/modeltools/R/matrix.R
```diff
@@ -0,0 +1,75 @@
...
+#'
+#' @param lagged_df Data frame of lagged data.  It should have the following columns:
+#'     \itemize{
+#'     \item{`geo_value`}{Strings of geographic locations.}
+#'     \item{`time_value`}{Dates of training data.}
+#'     \item{Covariate columns}{Columns with names of the form `value-{days}:{signal}` whose values
+#'         correspond to `{signal}` `{days}` before `time_value`}
+#'     \item{Response columns}{Columns with names of the form `value+{n}:{response}` whose values
+#'         correspond to `{response}` `{n}` incidence period units after `time_value`.  Since
+#'         columns beginning with "value+" are interpretted as the response rather than a
```
huisaddison, line 11:
```suggestion
 #'         columns beginning with "value+" are interpreted as the response rather than a
```
sorry for the super nitpick

huisaddison:
LGTM, but I've been told by @kjytay that there are some variable name / conventions that need to be resolved in #495 that also apply here.  

Also -- I made the tiniest of typo fixes, 'interpretted' to `interpreted' in the documentation.  I committed it directly to the PR because I thought that would be easiest, but I now realize that it also needs to be fixed in the "compiled" Rd documentation file.  We can hand edit that too since it's such a small change, or you could fix the typo and recompile in your fork later on after we sort out the stuff in #495.

sgsmob:
The change in convention from #495 has been ported to this PR.

R-packages/modeltools/R/matrix.R
```diff
@@ -0,0 +1,72 @@
...
+#' @param lagged_df Data frame of lagged data.  It should have the following columns:
+#'     \itemize{
+#'     \item{`geo_value`}{Strings of geographic locations.}
+#'     \item{`time_value`}{Dates of training data.}
+#'     \item{Covariate columns}{Columns with names of the form `value-{days}:{signal}` or
+#'         `value+0:{signal} whose values correspond to `{signal}` `{days}` before `time_value`}
+#'     \item{Response columns}{Columns with names of the form `response+{n}:{response}` whose values
+#'         correspond to `{response}` `{n}` incidence period units after `time_value`.}
+#'     }
+#'     A data frame in this format can be made using `covidcast::aggregate_signals()`.
```
kjytay, line 12: I think it is no longer true that `covidcast::aggregate_signals`` gives this because of the new naming convention.

R-packages/modeltools/R/matrix.R
```diff
@@ -0,0 +1,72 @@
...
+#'         from the `training_window_size` days prior to it.}
+#'     \item{`train_y`}{Vector of response data from the `response+{ahead}:{response}` column of
+#'         `lagged_df` corresponding to the rows of `train_x`.}
+#'     \item{`predict_x`}{Matrix of prediction data in the same format as `train_x`.  The
+#'         prediction data contains the most recent `training_window_size` days.}
+#'     \item{`predict_geo_values`}{Vector of `geo_values` corresponding to the rows of `predict_x`.}
+#'     }
+#'
+#' @importFrom tibble tibble
+#'
```
kjytay, line 32: Could we include one example call? A simplified version of the test will do.

R-packages/modeltools/R/matrix.R
```diff
@@ -0,0 +1,72 @@
...
+#' @export 
+create_train_and_predict_matrices <- function(lagged_df, ahead, training_window_size) {
+    out <- list()
+
+    train_df <- lagged_df %>%
+        select(geo_value, time_value, tidyselect::matches("^value(\\+0|-)"))
```
kjytay, line 38: With the new naming convention, can the regex be shortened to just "value"? (This appears another time below too.)

R-packages/modeltools/R/matrix.R
```diff
@@ -0,0 +1,72 @@
...
+#'     \itemize{
+#'     \item{`geo_value`}{Strings of geographic locations.}
+#'     \item{`time_value`}{Dates of training data.}
+#'     \item{Covariate columns}{Columns with names of the form `value-{days}:{signal}` or
+#'         `value+0:{signal} whose values correspond to `{signal}` `{days}` before `time_value`}
+#'     \item{Response columns}{Columns with names of the form `response+{n}:{response}` whose values
+#'         correspond to `{response}` `{n}` incidence period units after `time_value`.}
+#'     }
+#'     A data frame in this format can be made using `covidcast::aggregate_signals()`.
+#' @param ahead Number of incidence period units (i.e., epiweeks, days, etc.) ahead to forecast
```
kjytay, line 13: I don't think this function works correctly for any incidence period other than days. (The way `quantgen_forecaster` does it only works correctly for incidence_period = "day".)

Part of me feels that it would be much easier for this function (and other related ones) to force the user to do all these operations in days. It might be easier for the user to convert the epiweek aheads into day aheads (perhaps using a separate function), then use the day aheads in these functions. I suspect this is already what @dajmcdon is doing with anteater (forecasting 1-4 epiweek ahead on a Monday corresponds to 5, 12, 19, 26 days ahead).

R-packages/modeltools/tests/testthat/test-matrix.R
```diff
@@ -0,0 +1,44 @@
...
+    expect_equal(out$train_x,
+                 as.matrix(tibble(
+                    `value-2:signal_1` = c(1, 2),
+                    `value-2:signal_2` = c(11, 12),
+                    `value-1:signal_1` = c(3, 4),
+                    `value-1:signal_2` = c(13, 14),
+                    `value+0:signal_1` = c(5, 6),
+                    `value+0:signal_2` = c(15, 16)))
+    )
+    expect_equal(out$train_y, c(19, 20))
```
kjytay, line 33: I think this is not correct. You have two response+2 columns, so based on your code I think both of these columns should appear in `out$train_y`.

R-packages/modeltools/R/matrix.R
dajmcdon: Re epiweek/days: I think this has gotten bungled across different packages. There are various iterations of `sum_to_epiweek` that don't feel quite right to me. If I'm doing everything from scratch, I would do two things:

1. Rather than summing up to epiweek, I would "average up". Best to use `modeltools::Mean(x)*7`. In my opinion, there's nothing special about the day an epiweek begins on from the standpoint of the forecasting. Any weekly seasonality is removed with the "average up" operation. This allows us to then use daily data always. There's of course lots of correlation between the observations, which reduces the effective sample size, but we were often reducing our sample size from 7 to 1 per week explicitly. `Mean(x) = mean(x, na.rm=T)` so as long as a week has one observation, we have a value. This automatically imputes missing data with the trailing mean (there are some potential issues here which should get handled by the corrections).
2. Add a simple function that determines the correct target day based on incidence period and forecast date. I did this manually as @kjytay describes. I think this is the only thing necessary to enable the "epiweek" task:
```r
get_target_ahead <- function(forecast_date, incidence_period, ahead) {
  forecast_date <- lubridate::ymd(forecast_date)
  ed <- evalcast::get_target_period(forecast_date, incidence_period, ahead)$end
  return(as.numeric(ed - forecast_date))
}
```

R-packages/modeltools/R/matrix.R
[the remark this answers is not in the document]
sgsmob: obsolete

R-packages/modeltools/R/matrix.R
```diff
@@ -0,0 +1,75 @@
...
+#'         correspond to `{signal}` `{days}` before `time_value`}
+#'     \item{Response columns}{Columns with names of the form `value+{n}:{response}` whose values
+#'         correspond to `{response}` `{n}` incidence period units after `time_value`.  Since
+#'         columns beginning with "value+" are interpretted as the response rather than a
+#'         covariate, observed data from the same day as `time_value` should be in columns
+#'         `value-0:{signal}` rather than `value+0:{signal}`}
+#'     }
+#'     A data frame in this format can be made using `covidcast::aggregate_signals()`.
+#' @param ahead Number of incidence period units (i.e., epiweeks, days, etc.) ahead to forecast
+#' @param training_window_size Size of the local training window in days to use. For example, if
```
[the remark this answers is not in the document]
sgsmob: The training window size is always in days because the signal variables are reported daily or aggregated over some period but associate with a single day.  This is because we have no way to express time values that are not days (i.e., "epiweek 33").  So `incidence_period` only applies to our interpretation of `ahead`s (future data), not to the downloaded signals (past data).

Reworded slightly to clarify.

R-packages/modeltools/R/matrix.R
```diff
@@ -0,0 +1,75 @@
...
+#'     `training_window_size = 14`, then to make a 1-day-ahead forecast on December 15, we train on
+#'     data from December 1 to December 14.
+#'
+#' @return Named list with entries:
+#'     \itemize{
+#'     \item{`train_x`}{Matrix of training data whose columns correspond to the
+#'         `value-{days}:{signal}` columns in `lagged_df`.  The training data consists of the
+#'         latest date `d` such that there is an observed response at time `d + ahead` and all data
+#'         from the `training_window_size` days prior to it.}
```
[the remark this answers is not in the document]
sgsmob: See above

R-packages/modeltools/R/matrix.R
```diff
@@ -0,0 +1,75 @@
+#' Create a training and prediction matrices for a given ahead.
```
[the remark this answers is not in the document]
sgsmob: Done

R-packages/modeltools/R/matrix.R
sgsmob: Updated to include the new `add_response_columns()` function.

R-packages/modeltools/R/matrix.R
sgsmob: Done

R-packages/modeltools/R/matrix.R
sgsmob: It doesn't matter what the incidence period is for this function because the meaning of the response here is not important.  We only need to match `ahead` in the `"response+{ahead}:{signal}"` column name to the `ahead` value passed in as an argument to this function.  Obviously Garbage In = Garbage Out, but it is not the job of this function to verify that the `reponse*` columns were created in a sensible way.

R-packages/modeltools/tests/testthat/test-matrix.R
sgsmob: Yes something is not right here.  Looking into it

R-packages/modeltools/R/matrix.R
kjytay: Yes, so personally I would prefer that this function talk about everything in terms of days and avoid mention of "incidence period".

sgsmob: All of the computation between time values is done in days here.  The only place that the `incidence_period` only matters in the way that we interpret the `reponse*` columns.  I am against translating things into days since we are currently very strongly using the `ahead = 1:4` epiweek convention all over, such as:

- How we call `get_predictions()`
- How we submit our weekly reports
- How all the researchers (especially senior ones!) talk about it

I've removed some references to `incidence_period` where I can in the documentation.

R-packages/modeltools/R/matrix.R
sgsmob: Yes, done

R-packages/modeltools/tests/testthat/test-matrix.R
sgsmob: It turns out that `pull()` by default selects the last column and silently ignores the others.  I have a check fail in here if there are multiple response columns for the same ahead.  We could also change this to allow users to specify the response variable but I don't know if that's actually something that people would want to do in the same matrix or if they'd rather have one copy of the matrix per response (would be straightforward for users to do this with `add_reponse_columns()`.

R-packages/modeltools/R/matrix.R
dajmcdon: The processing is much easier in days. Maybe (internally):
1. Convert inputs of epiweeks ahead into days ahead.
2. Grab lags.
3. If epiweeks, rewrite the names `response+{ahead in days}` back to `response+{ahead in epiweeks}`.

sgsmob: I don't have too strong of an objection to this but it wouldn't change any code in this PR at all.  In fact, I don't see how the above changes would make a difference anywhere in the code.  Right now we have the following workflow (forgive the loose pseudocode):
```
add_response_columns <- function(df, ahead, incidence_period) {
  period <- get_target_period(incidence_period, ahead)
  sum_columns(df, period$start, period$end)
}

forecaster <- function(df_list) {
  df <- aggregate_signals(df_list, lags)
  for (a in ahead) {
     response_df <- add_response_columns(df, a, incidence_period)
     mat <- create_train_and_predict_matrices(response_df)
     predict(fit(mat))
}
```
If we convert everything to days "internally" I think we just end up with this pseudocode:
```
add_response_columns <- function(df, period_start, period_end) {
  sum_columns(df, period_start, period_end)
}

forecaster <- function(df_list) {
  df <- aggregate_signals(df_list, lags)
  for (a in ahead) {
     a <- a * incidence_period
     period <- get_response_period(a, incidence_period)
     response_df <- add_response_columns(df, period$start, period$end)
     mat <- create_train_and_predict_matrices(response_df)
     mat <- rename_response_columns_to_match_epiweek(mat)
     predict(fit(mat))
}
```
What is gained here/

dajmcdon: I agree. Seemingly nothing.

dajmcdon (approved):

dajmcdon:
I'm ready to merge unless there's @kjytay has anything outstanding.

R-packages/modeltools/R/matrix.R
kjytay: Yes. For me it’s about making sure the documentation is clear.

kjytay (approved):
Looks good to me

PR #499, opened 2021-03-19
R pkg devel ta

statsmaths:
I had local code that created the simplified geojson files for covidcast; this PR pushes them into the repository. I also cleaned up the code that creates the "rda" files. It was adding two files that are not documented or mentioned in the package. If these need to be included, someone can uncomment those lines in "create_rda.R" and propoerly document them in the package.


statsmaths:
Sorry, wanted to merge with r-pkg-devel, not main.

PR #500, opened 2021-03-19
Adding code to create simplified geojson files

statsmaths:
I had local code that created the simplified geojson files for covidcast; this PR pushes them into the repository. I also cleaned up the code that creates the "rda" files. It was adding two files that are not documented or mentioned in the package. If these need to be included, someone can uncomment those lines in "create_rda.R" and propoerly document them in the package.


capnrefsmmat:
I think state_geo and county_geo are indeed safe to remove, judging by NEWS.md:

```
- The `county_geo` and `state_geo` built-in datasets have been removed. These
  were used to support bubble plots, but the changes to
  `plot.covidcast_signal()` rendered them unnecessary.
```

capnrefsmmat (approved):
Looks good, I think.

(also, yay, 500 PRs!)

R-packages/data-raw/create_rda.R
```diff
@@ -57,10 +59,10 @@ county_col_classes = c("numeric", "numeric", "character", "character",
                        "character", "character")
 county_file = system.file("extdata", "us_counties_centroids.csv", package = "usmap")
 county_geo = utils::read.csv(county_file, colClasses = county_col_classes)
-save(county_geo, file = "../covidcast/data/county_geo.rda", compress = "bzip2")
+#save(county_geo, file = "../covidcast/data/county_geo.rda", compress = "bzip2")
```
sgsmob, line 62: Can you delete this line entirely?

R-packages/data-raw/geo/county/cb_2019_us_county_5m.cpg
```diff
@@ -0,0 +1 @@
+UTF-8
```
sgsmob, line 1: What is this file for?

sgsmob (approved):
Generally LGTM just a few minor things

statsmaths:
Great! Glad I could get us over the 500 mark by messing up the first PR.

R-packages/data-raw/geo/county/cb_2019_us_county_5m.cpg
statsmaths: These are the spatial polygon files provided by the US Census Bureau. I think they are in ArcGIS format (could be wrong about that though). I have no idea what each of the individual parts are for, but they are needed to read the data into R. I am putting them here in case at some point we need to re-generate the smaller geojson files that are in the actual package.

R-packages/data-raw/create_rda.R
statsmaths: I wouldn't delete just this line. If we don't want this code saved anywhere, we should delete the whole block. I just did this part so it does not erronously create the output. Note: I did not write this file. Was just fixing an error that I found.

R-packages/data-raw/geo/county/cb_2019_us_county_5m.cpg
capnrefsmmat: Yeah, shapefiles are an archaic format that consists of a bunch of smaller files. Apparently this one is "An optional file that can be used to specify the codepage for identifying the characterset to be used."

https://desktop.arcgis.com/en/arcmap/10.3/manage-data/shapefiles/shapefile-file-extensions.htm

PR #502, opened 2021-03-19
Trim down arguments in `evalcast::get_predictions()`

sgsmob:
Change the signature of `evalcast::get_predictions()` to allow for more flexibility of specifying data retrieval and to minimize the set of arguments passed to the forecaster.   Specifically:

- Remove `ahead` entirely from calls to `get_predictions()`
- Move `geo_type` from an explicit, separate argument to a column in `signals`
- Replace the `...` arguments with an explicit `forecaster_args` list of arguments

Since the calls to the forecasters no longer reflect some of the internal processing, additional tests have been added to verify that the right arguments are being passed to `download_signal()`

### DIFF
--- a/R-packages/evalcast/R/get_predictions.R
+++ b/R-packages/evalcast/R/get_predictions.R
@@ -17,9 +17,6 @@
 #' @template signals-template
 #' @template forecast_dates-template
 #' @template incidence_period-template
-#' @template ahead-template
-#' @template geo_type-template
-#' @template geo_values-template
 #' @template apply_corrections-template
 #' @param signal_aggregation this and the next argument control the type of
 #'   data your forecaster expects to receive from covidcast. By default,
@@ -39,7 +36,7 @@
 #'   You can override this functionality, though we strongly advise you do so
 #'   with care, by passing a function of a single forecast_date here. The 
 #'   function should return a date.
-#' @param ... Additional named arguments to be passed to `forecaster()` 
+#' @param forecaster_args List of forecaster-specific arguments to pass to `forecaster()` 
 #' @template predictions_cards-template
 #' 
 #'
@@ -49,9 +46,9 @@
 #'   tibble::tibble(
 #'     data_source=c("jhu-csse", "usa-facts"),
 #'     signal = c("deaths_incidence_num","confirmed_incidence_num"),
-#'     start_day="2020-08-15"), 
-#'   "2020-10-01","epiweek", 1:4, 
-#'   "state", "mi", signal_aggregation="long")
+#'     start_day="2020-08-15",
+#'     geo_type = "state"), 
+#'   "2020-10-01","epiweek", signal_aggregation="long")
 #' }
 #' 
 #' @export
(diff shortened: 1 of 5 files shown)


dajmcdon:
Crud, I was just working on this...

Of course with different choices.

sgsmob:
What different choices would you be making?

dajmcdon:
1. I retain the `ahead` argument, though, since you removed it, now I'm thinking maybe it is more appropriate as an argument to the forecaster?
2. I'm allowing the `signals` tibble to also contain `as_of` (set to `forecast_date` by default) and `geo_values`.
3. I changed the `get_predictions signature:
```r
get_predictions <- function(forecaster,
                            name_of_forecaster,
                            signals_to_use,
                            forecast_dates,
                            ahead = 1,
                            incidence_period = c("epiweek","day"),
                            apply_corrections = function(signals) signals,
                            response_geo_type = c("county", "hrr", "msa", "dma",
                                                  "state", "hhs", "nation"),
                            response_data_source = signals_to_use$data_source[1],
                            response_data_signal = signals_to_use$signal[1],
                            forecaster_params = list()) {
```
4. I wasn't passing `forecast_date` to the forecaster. 
5. Additionally, I'm working on how to call the forecaster so that it can be a function like `forecaster(signals,...)`. Using `do.call(forecaster, forecaster_params)` it will expand the list to pass them as named arguments. That way you don't need a wrapper. But it's annoying to require one of the arguments to be *named* `signals` or `df_list`, so I was looking for a workaround. Considering making `fcaster <- function(...) forecaster(df,...)`. This would then allow me to do `do.call(fcaster, forecaster_params` and the only requirement on `forecaster` would be that the first argument is the list (but named whatever you want).

**Moving forward**
I'm thinking I merge your PR, then merge your changes into mine and then iterate a bit subject to your thoughts on the above points. Then I'll do a PR and tag you.

dajmcdon (approved):

sgsmob:
Some thoughts on 1-5
1.  `ahead` has no purpose in downloading data so either it is totally unused or passed along to the forecaster.  Either way it should be a forecaster argument.
2.  I like the `as_of` in `signals`.  I thought we decided against `geo_values` entirely in the last PR.  We can add it back in but we would have to think of the best way to do this.  If we have it as a column in `signals` you would have to copy the same value in multiple columns if you are using a mix of `geo_type`s.  Originally I had gotten around this by making a separate frame with columns `geo_type` and `geo_values` such that you are specifying the values once per type.  Of course, this prevents you from having different values for signals of the same types.  Not sure what the typical use cases would be here.
3. What are the `response_*` arguments for?
4. Since we allow multiple forecast dates to be specified for the same call to `get_predictions()` we either need to pass it along to the forecaster or have the forecaster infer it implicitly from the signals.  The latter seems unnecessarily restrictive so I think we should do the former.
5. This is probably a good idea, although we need to think about how this interacts with point 4 above

dajmcdon:
1. I agree. Let's dump this.
2. I think the only real use for `geo_values` is to simplify downloading. I think it would be used infrequently. The tibble construction should allow if to be specified once (and default to "*"). If you don't have something against it, I'll try to make it work, and you can decide in the PR if you hate it.
3. In the current version we assume that the first row of the `signals` tibble tells us what the response is. We need to identify it for evaluation later. These allow us to override that case. A standard case that happens is that `zyzzyva` used `use-facts` but we want to evaluate against `jhu-csse`. My thinking is just to optionally decouple these (of course you could just rename things when the data frame gets returned).
4. Sounds good.
5. I think I can work it so that the forecast_date is passed as a named argument. Then any function with `...` will either use or ignore it.

Issue #421, opened 2021-01-15
Make fips_to_abbr behave in a more predictable way


capnrefsmmat:
Example:

```
> fips_to_abbr("10")
01000 
 "AL" 
Warning message:
Some inputs were not uniquely matched; returning only the first match in each case. 

> fips_to_abbr("10", ties_method="all")
[[1]]
01000 10000 11000 21000 31000 41000 51000 
 "AL"  "DE"  "DC"  "KY"  "NE"  "OR"  "VA" 
```

This isn't great. If I have a bunch of two-digit state FIPS codes and want to match them, I need to do this instead:

```
> fips_to_abbr("^10")
10000 
 "DE" 
```

since it's a regular expression. And so if I have a vector of two-digit FIPS codes from some outside source, I need to do

```
> fips_to_abbr(paste0("^", fipsen))
10000 11000 01000 
 "DE"  "DC"  "AL"
```

Instead, `fips_to_abbr` should understand that two-digit FIPS codes are a prefix match and do this for me.

statsmaths:
The function `fips_to_name` has the exact same problem.

statsmaths:
In other words, is there a reason this doesn't provide what is needed (and if not, what's a counter example that does not work as expected):

```
fips_to_abbr <- function(code)
{
  fips <- sprintf("%02d", covidcast::state_census$STATE)
  index <- match(substr(code, 1, 2), fips)
  output <- covidcast::state_census$ABBR[index]
  names(output) <- code
  output
}
```

capnrefsmmat:
> For me, I don't see a use case where I would want anything other than a single exact match of the a two digit fips code or a single exact match of the first 5 digits of a fips code.

I agree; I don't know of any use cases where I would, say, want to partially match on the inner three digits of a FIPS code. I can't think of any use cases we have for any other type of matching. I don't think there's any meaningful hierarchy to FIPS where one would want, say, all FIPS codes starting with 223 -- 22, yes, but 223 is just an arbitrary set of counties.

As for your suggested code, maybe the only thing I'd change is how the output vector is named:

```
> fips_to_abbr("22")
  22 
"LA" 
> fips_to_abbr("2200")
2200 
"LA" 
> fips_to_abbr("22000")
22000 
 "LA" 
```

For consistency we probably want a common naming used for all states, e.g. every state always gets its two-digit name in the vector.

statsmaths:
[...]

PR #503, opened 2021-03-19
R pkg devel ta

### LINKED ISSUE #273: Add CITATION file to R client
R packages can contain a [CITATION file](https://cran.r-project.org/doc/manuals/R-exts.html#CITATION-files) so that users can type `citation("covidcast")` and get a recommended citation, both in plain text and BibTeX formats.

Once we add this, we should make the home page of the pkgdown site list the citation as well.

This will require deciding on what the citation should be.

statsmaths:
This includes several commits which do the follow:

- Elevates warnings in R CMD CHECK to errors, as discussed in #470 
- Adds a CITATION file, to resolve #273 
- Fixes bug in `fips_to_name` by splitting into two function, as discussed in #421


sgsmob (approved):
In the future would prefer to separate these out into 2-3 separate PRs but what we have here is all ok.

krivard (approved):
lgtm

statsmaths:
I understand the desire to localize changes; that is why I put each resolution into a different commit. However, I don't think there is a way to do a pull request on a specific commit; it is only possible to specify a branch to merge with as far as I am aware. So, I would mean that—unless I stop working each time I resolve an issue while it is reviewed—I need to create an brand new branches for every single issue. Is that really necessary?

R-packages/covidcast/inst/CITATION
```diff
@@ -0,0 +1,24 @@
...
+                    person(given = "Logan", family = "Brooks"),
+                    person(given = "Sarah", family = "Colquhoun"),
+                    person(given = "David", family = "Farrow"),
+                    person(given = "Jed", family = "Grabman"),
+                    person(given = "Pedrito", family = "Maynard-Zhang"),
+                    person(given = "Alex", family = "Reinhart"),
+                    person(given = "Ryan", family = "Tibshirani")),
+         year = year,
+         note = note,
+         url = "https://CRAN.R-project.org/package=nlme")
```
capnrefsmmat, line 19: whoops, wrong URL here -- can you change this to https://cmu-delphi.github.io/covidcast/covidcastR/?

statsmaths: Oh yes, that was certainly an oversight on my part! I think since I already did this, I will need to change it outside the PR.

statsmaths: Okay, fixed now in r-pkg-devel

PR #506, opened 2021-03-19
Evalcast parallelsigs

dajmcdon:
Following on the earlier merge #502. See the discussion there.

@sgsmob There are a few failing tests. All are in `test-predictions.R`. They all have the form of
```r
expect_equal(mock_args(mock_forecaster),...)
```
I couldn't get the mock stuff to run locally, nor could I discern what these are supposed to do. I imagine you can catch them much faster than me. There are 4 failures on lines 89, 141, 206, and 288.

Obviously, I'm open to other comments/suggestions as well.


sgsmob:
Since we don't actually care about or want to depend on what goes on inside the forecaster function, we give a mock function that
- records what arguments your passed into it.
- outputs prespecified output data frames.

But because the mock gives the same output regardless of the input, to make sure we are calling it correctly, we check the arguments it records against what we think should have been passed in.  Thus the `expect_equal(mock_args(mock_forecaster),...)` calls.  The `mock_args()` gives a list with one element per call to the mock forecaster.  The elements of this list are themselves lists with named values for the arguments provided at the call.  So 
```
forecaster(a = 1, b = 2)
forecaster(a = 3)
```
should show up as
```
list(list(a = 1, b = 2), list(a = 3))
``` 

When these checks fail they are a little opaque, but they will often say something about "Component X: Component Y", meaning the Yth entry of the Xth entry of the list doesn't match.  Adding a print statement of the `mock_args` can help illuminate what is going on too.

In your case, because you changed the input format to the forecaster, the mock args don't line up anymore and the tests fails.  You will need to change the values of every one of the nested lists to reflect what arguments you are expecting to get via the `do.call` call.

dajmcdon:
Ok, Everything should work now. Please review as usual.

R-packages/evalcast/R/signal_listcols.R
```diff
@@ -0,0 +1,32 @@
+signal_listcols <- function(signals, forecast_date) {
```
sgsmob, line 1: Please add some documentation as to what this function is supposed to be doing.

R-packages/evalcast/R/signal_listcols.R
```diff
@@ -0,0 +1,32 @@
...
+    }
+    signals$as_of <- as_of
+  }
+  
+  if (! "geo_values" %in% cn) signals$geo_values <- "*"
+  if (! "geo_type" %in% cn) signals$geo_type <- "county"
+  
+  return(signals)
+  
+}
```
sgsmob, line 32: trailing newline

R-packages/evalcast/R/signal_listcols.R
```diff
@@ -0,0 +1,32 @@
...
+    std <- Date(ns)
+    for (i in 1:ns) {
+      if (is.function(signals$start_day[[i]])) {
+        std[i] <- as.Date(signals$start_day[[i]](forecast_date))
+      }
+    }
+    signals$start_day <- std
+  }
+  
+  if (! "as_of" %in% cn) signals$as_of <- forecast_date
```
sgsmob, line 16: would be nice to combine or simplify this in some way:
```
if ("as_of" %in% cn) {
  if (is.list(signals$as_of)) {
    ...
  }
} else {
  signals$as_of <- forecast_date
}
```
or 
```
if (! "as_of" %in% cn) ...
if (is.list(signals)) {
  ...
}
```
or 
```
if (! "as_of" %in% cn) {
  ...
} else if (is.list(signals)) {
  ...
}
```

R-packages/evalcast/tests/testthat/test-predictions.R
```diff
@@ -139,12 +141,17 @@ test_that("get_predictions works when forecaster has additional arguments", {
...
     expect_equal(mock_args(mock_forecaster),
-                 list(list(fake_downloaded_signals,
-                           as.Date("2020-01-01"),
-                           list(a = 1, b = "2")),
-                      list(fake_downloaded_signals,
-                           as.Date("2020-01-02"),
-                           list(a = 1, b = "2"))))
+                 list(
+                   list(
+                     a = 1, b = "2",
```
sgsmob, line 146: It is odd that the extra arguments `a` and `b` get keyword names but the list of downloaded data frames, which is one of the two required arguments to the forecaster, is the entirety of the `...` argument.  I don't think this really makes sense as an API.  Shouldn't it have a common keyword argument (something like `df_list`, `signals`, etc.)?

R-packages/evalcast/tests/testthat/test-data_corrector.R
```diff
@@ -1,68 +0,0 @@
-# signals = tibble::tibble(
```
sgsmob, line 1: Is the purpose of deleting this because we don't have any active tests?

sgsmob (changes requested):

R-packages/evalcast/R/signal_listcols.R
dajmcdon: Done.

R-packages/evalcast/R/signal_listcols.R
dajmcdon: done.

R-packages/evalcast/R/signal_listcols.R
dajmcdon: done.

R-packages/evalcast/tests/testthat/test-predictions.R
dajmcdon: I think this issue is because I don't understand the mocking. What should happen is that we don't care what the first argument is named, only that it expects the data in the first argument. The code on line 130:
```r
df_arg <- methods::formalArgs(forecaster)[1] # grab name of first arg to forecaster
```
actually grabs whatever they call the data, and names the data that before passing it in. So we pass in all arguments as name/value pairs.

However, the mocking seems to not have a first argument, just ...? So it names it `...`. I couldn't figure out how to track the names of the mocked function.

Long story short, I think this is a problem with the test, not the function.

R-packages/evalcast/R/signal_listcols.R
```diff
@@ -0,0 +1,39 @@
...
+    std <- Date(ns)
+    for (i in 1:ns) {
+      if (is.function(signals$start_day[[i]])) {
+        std[i] <- as.Date(signals$start_day[[i]](forecast_date))
+      }
+    }
+    signals$start_day <- std
+  }
+  
+  if (! "as_of" %in% cn) {
```
JedGrabman, line 20: Swap the branches. i.e. `if foo... else` instead of `if !foo... else`

R-packages/evalcast/R/signal_listcols.R
```diff
@@ -0,0 +1,39 @@
...
+# of the forecast_date. Here we evaluate those functions as well as creating
+# default values for `geo_values` and `geo_type` if necessary
+signal_listcols <- function(signals, forecast_date) {
+  
+  cn <- colnames(signals)
+  ns <- nrow(signals)
+  # start_day should be NULL or date. We handle date/chr/fun(forecast_date) input
+  if ("start_day" %in% cn && is.list(signals$start_day)) {
+    std <- Date(ns)
+    for (i in 1:ns) {
```
JedGrabman, line 12: I don't think it's an issue here, but consider using `seq_len` when creating sequences. It avoids issues with objects that can be length 0.
e.g. `for (i in 1:0)` loops through i = 1 and i =0, but `for i in seq_len(0)` doesn't.

R-packages/evalcast/man-roxygen/signals-template.R
```diff
@@ -1,15 +1,43 @@
...
+#'   computed dynamically from the forecast date (e.g., when `forecaster` only
+#'   trains on the most recent 4 weeks of data). 
+#'   
+#'   You may also include a `geo_type` column, a `geo_values` column and/or an
+#'   `as_of` column. 
+#'   
+#'   The first two should contain a string. If unspecified, these will have
+#'   the same defaults as `covidcast::covidcast_signal()`, namely 
+#'   `geo_type = "county"` and `geo_values = "*"`.
+#'   These arguments allow you to override
```
JedGrabman, line 27: typo: "to override to download"

R-packages/evalcast/man-roxygen/signals-template.R
```diff
@@ -1,15 +1,43 @@
-#' @param signals Tibble specifying which variables from the COVIDcast API will be downloaded and passed along to `forecaster`.
-#'   Each row of `signals` represents a separate variable to be downloaded.  Available data
-#'   sources, signals, and geo resolutions are documented in the [COVIDcast signal
+#' @param signals Tibble with mandatory columns `data_source` and `signal` that
```
JedGrabman, line 1: Suggest listing all possible columns at the start, then going into detail about them. Something like "mandatory columns `data_source` and `signal` and optional columns..."

JedGrabman:
@sgsmob "But because the mock gives the same output regardless of the input, to make sure we are calling it correctly, we check the arguments it records against what we think should have been passed in."

Would `with_mock_api` be useful here? It only mocks when specific arguments are provided, so it might be a more direct way to verify. See some documentation we have [here](https://github.com/cmu-delphi/covidcast/blob/main/R-packages/covidcast/tests/testthat/test-covidcast.R). Been a while since I did mocking, so ignore if not relevant.

R-packages/evalcast/R/get_covidhub_predictions.R
```diff
@@ -316,12 +316,34 @@ get_covidhub_forecaster_names <- function(
...
+#' @export
+#'
+#' @examples
+#' covidhub_probs()
+covidhub_probs <- function(type = c("standard", "inc_case")) {
+  type = match.arg(type)
+  switch(type,
+         standard = c(0.01, 0.025, seq(0.05, 0.95, by = 0.05), 0.975, 0.99),
+         inc_case = c(0.025, 0.100, 0.250, 0.500, 0.750, 0.900, 0.975),
+         c(0.01, 0.025, seq(0.05, 0.95, by = 0.05), 0.975, 0.99)
```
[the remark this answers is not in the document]
dajmcdon: Done.

R-packages/evalcast/R/baseline_forecaster.R
```diff
@@ -25,43 +23,36 @@
...
   target_period <- get_target_period(forecast_date, incidence_period, ahead)
   incidence_length <- ifelse(incidence_period == "epiweek", 7, 1)
-  covidhub_probs <- c(0.01, 0.025, seq(0.05, 0.95, by = 0.05), 0.975, 0.99)
   dat <- list()
   s <- ifelse(symmetrize, -1, NA)
   
-  
-  df <- get_response_df(df, signals)
-  
+  if (class(df)[1] == "list") df <- df[[1]]
```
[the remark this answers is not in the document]
dajmcdon: They should expect a list of COVIDcast signals. I think our use of `pmap` means that the result is always a list, even if length one. This also allows someone to pass many signals into baseline, but only the first gets used.

R-packages/evalcast/R/get_predictions.R
```diff
@@ -13,105 +13,104 @@
 #'   `geo_value`, `quantile`, and `value`. The `quantile` column gives the
 #'   probabilities associated with quantile forecasts for that location and
 #'   ahead. If your forecaster produces point forecasts, then set `quantile=NA`.
+#'   
+#'   The first argument must be the list of signals returned by a call to the COVIDcast.
```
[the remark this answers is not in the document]
dajmcdon: That was my intention. I guess "list of signals" isn't clear. Revised.

R-packages/evalcast/R/get_predictions.R
```diff
@@ -13,105 +13,104 @@
...
-#'   forecaster. It's likely, for example, that no data would actually exist
-#'   for the forecast date on the forecast date (there is some latency between
-#'   the time signals are reported and the dates for which they are reported).
-#'   You can override this functionality, though we strongly advise you do so
-#'   with care, by passing a function of a single forecast_date here. The 
-#'   function should return a date.
-#' @param forecaster_args List of forecaster-specific arguments to pass to `forecaster()` 
+#' @param response_data_source String indicating the `data_source` of the response.
+#'   This is used mainly for downstream evaluation. By default, this will be the 
+#'   same as the first row in the `signals` tibble.
```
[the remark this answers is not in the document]
dajmcdon: done.

R-packages/evalcast/R/get_predictions.R
```diff
@@ -13,105 +13,104 @@
...
+#' @param response_data_signal String indicating the `signal` of the response.
+#'   This is used mainly for downstream evaluation. By default, this will be the 
+#'   same as the first row in the `signals` tibble.
```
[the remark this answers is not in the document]
dajmcdon: done.

R-packages/evalcast/R/get_predictions.R
```diff
@@ -13,105 +13,104 @@
...
 get_predictions <- function(forecaster,
                             name_of_forecaster,
                             signals,
                             forecast_dates,
-                            incidence_period,
-                            apply_corrections = NULL,
-                            signal_aggregation = c("list", "wide", "long"),
-                            signal_aggregation_dt = NULL,
-                            as_of_override = function(forecast_date) forecast_date,
+                            incidence_period = c("epiweek","day"),
```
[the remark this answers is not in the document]
dajmcdon: done.

R-packages/evalcast/R/signal_listcols.R
```diff
@@ -0,0 +1,39 @@
+# In the signals tibble, required by `get_predictions()`,
+# start_day` and `as_of` can possibly be functions
+# of the forecast_date. Here we evaluate those functions as well as creating
+# default values for `geo_values` and `geo_type` if necessary
+signal_listcols <- function(signals, forecast_date) {
+  
+  cn <- colnames(signals)
+  ns <- nrow(signals)
```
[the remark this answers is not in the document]
dajmcdon: done.

R-packages/evalcast/R/signal_listcols.R
dajmcdon: I always forget to use that. Done.

R-packages/evalcast/R/signal_listcols.R
dajmcdon: Done.

R-packages/evalcast/man-roxygen/signals-template.R
dajmcdon: done.

R-packages/evalcast/R/get_predictions.R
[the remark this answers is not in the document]
JedGrabman: Clarified in person. It's a list of dataframes, where each dataframe is the result for a single signal.

sgsmob (approved):

PR #509, opened 2021-03-22
Use fread instead of read_csv

JedGrabman:
Update from using `read_csv` to `fread` to read in files.

This has several benefits for users who need to load a large number of files:

- Speed: `fread` is about 3x faster than `read_csv`. The overall speedup for the end user is smaller because of other unchanged steps, but the speed up is still noticeable.
- Memory: `read_csv` has a memory leak, where a small amount of memory becomes inaccessible with each file processed. Because there are thousands of predictions files that can be loaded, this can cause memory issues for the heaviest users
- Accuracy: A more minor point, but `fread` is able to intuit a few malformed dates that `read_csv` does not . e.g. "2020-11-7" is correctly parsed as "2020-11-07" instead of `NA`.

---
Small benchmark test (fread vs read_csv):
```
library(pryr)
gc()
old_mem = mem_used()
start_time = proc.time()
foo = get_covidhub_predictions("CMU-TimeSeries")
end_time = proc.time()
end_time - start_time # 6.21 Seconds vs. 8.3 Seconds
rm(foo)
gc()
new_mem = mem_used()
new_mem - old_mem # 3.39 MB vs. 28.1 MB
```


JedGrabman:
Testing note: 
I ran the eval-dashboard pipeline with these changes (which includes every state level forecast for every forecaster with a designation of "primary") and it found no difference in output other than the extra date parsing ability mentioned above, and floating point level differences in `value` column.

sgsmob (approved):

dajmcdon (changes requested):
Can you run
```r
usethis::use_package("data.table")
```
That should add the dependency to DESCRIPTION and fix the build error.

dajmcdon (approved):

PR #510, opened 2021-03-24
Example forecaster for modeltools

sgsmob:
Adds an example forecaster to `modeltools` that demonstrates how to use the `modeltools` functions to construct a forecaster with `quantgen`.


sgsmob:
@dajmcdon This is my example forecaster

R-packages/modeltools/R/example_forecaster.R
```diff
@@ -0,0 +1,47 @@
...
+#'                                      start_date = as.Date("2021-01-01"),
+#'                                    ),
+#'                                    as.Date("2021-03-01"),
+#'                                    "epiweek")
+#'
+#' @export
+example_forecaster <- function(df_list, forecast_date) {
+    ahead <- 1:4
+    lags <- c(0, -1, -2, -3, -7, -14)
+    quantiles <- modeltools::covidhub_probs()
```
dajmcdon, line 21: Is `covidhub_probs` a function or data? See line 115 in `quantgen.R`.

R-packages/modeltools/R/example_forecaster.R
```diff
@@ -0,0 +1,47 @@
...
+    covariates <- covidcast::aggregate_signals(df_list, dt = lags, format = "wide")
+    covariates_with_response <- add_response_columns(covariates,
+                                                     "jhu-csse_confirmed_incidence_num",
+                                                     forecast_date,
+                                                     "epiweek",
+                                                     ahead)
+
+    results <- list()
+    for (a in ahead) {
+        mats <- create_train_and_predict_matrices(covariates_with_response, a, 14)
```
dajmcdon, line 32: the `14` here is the training window size? I'm wondering if this, `ahead` and `lags` shouldn't be arguments to the function?

R-packages/modeltools/R/quantgen.R
```diff
@@ -109,7 +109,7 @@
 #' @importFrom dplyr filter select pull summarize between bind_cols
 #' @importFrom tidyr pivot_longer
 #' @export
-quantgen_forecaster = function(df, forecast_date, signals, incidence_period,
+quantgen_forecaster = function(df_list, forecast_date, signals, incidence_period,
```
dajmcdon, line 112: This signature won't work with the new `evalcast`. Unless you pass some of these through the `forecaster_args()` list.

dajmcdon: Most of it is actually unnecessary though.

dajmcdon (changes requested):

R-packages/modeltools/R/quantgen.R
sgsmob: The plan is to make it all compliant with the new evalcast although I realize that they are not visible from one another yet.

R-packages/modeltools/R/example_forecaster.R
sgsmob: Fixed

R-packages/modeltools/R/example_forecaster.R
sgsmob: Done.

dajmcdon (approved):

PR #514, opened 2021-03-29
Fix verbose name printing

JedGrabman:
Fix incorrect variable name in verbose printing (`forecasters` -> `covidhub_forecaster_name`)


R-packages/evalcast/R/get_covidhub_predictions.R
```diff
@@ -92,7 +92,7 @@ get_covidhub_predictions <- function(
   }
   for (i in seq_len(length(covidhub_forecaster_name))) {
```
dajmcdon, line 93: Better is 
```r
for (i in seq_along(covidhub_forecaster_name)) {
```

dajmcdon (approved):

R-packages/evalcast/R/get_covidhub_predictions.R
JedGrabman: Fixed

PR #520, opened 2021-04-02
Extend create_train_and_predict_matrices() to work with multiple aheads

kjytay:
Notably, there is a new argument `aheads_separate` that allows the user to decide whether separate X and y matrices should be returned for each ahead value, or just one X and y for all aheads. The latter is useful for the multi-response modeling work Rob and Elena are doing.


R-packages/modeltools/R/matrix.R
```diff
@@ -50,50 +65,106 @@
...
+    predict_geo_values <- lagged_df %>%
         filter(time_value == max(time_value)) %>%
         select(geo_value) %>% pull()
-    out$train_end_date <- train_end_date
-
-    return(out)
+    
+    # make sure the response columns are unique
+    for (a in ahead) {
+        responses_at_ahead <- lagged_df %>%
```
dajmcdon, line 83: maybe call this `n_responses_at_ahead`?

R-packages/modeltools/R/matrix.R
```diff
@@ -50,50 +65,106 @@
...
+                summarize(max(time_value)) %>%
+                pull()
+            train_end_date <- min(max(lagged_df$time_value), response_end_date)
+            
+            # Training matrices
+            out$train_x <- lagged_df %>%
+                select(geo_value, time_value, tidyselect::starts_with("value")) %>%
+                filter(between(time_value,
+                               train_end_date - training_window_size + 1,
+                               train_end_date)) %>%
```
dajmcdon, line 107: There's a lot of duplication of this batch of code and the `aheads_separate == FALSE` case below. Can we maybe break out a function to wrap it (and allow for simple tests)?

R-packages/modeltools/R/matrix.R
```diff
@@ -50,50 +65,106 @@
...
+            
+            all_out[[paste0("ahead+", a)]] <- out
+        }
+        
+        if (length(ahead) == 1) {
+            return(all_out[[1]])
+        } else {
+            return(all_out)
+        }
+    } else {
```
dajmcdon, line 130: Let's note here what the `else` is doing explicitly: `! aheads_separate` or some such

dajmcdon (changes requested):

R-packages/modeltools/R/matrix.R
kjytay: Done.

R-packages/modeltools/R/matrix.R
kjytay: Done (in comment).

R-packages/modeltools/R/matrix.R
kjytay: I tried my best at refactoring this, let me know if this version is better.

R-packages/modeltools/R/matrix.R
```diff
@@ -87,84 +82,133 @@ create_train_and_predict_matrices <- function(lagged_df, ahead, training_window_
...
+#' aheads. If more than one ahead is given, the date return corresponds to the
+#' last possible date corresponding to the smallest ahead value. This is because
+#' the smallest ahead value will have the latest possible date.
+#'
+#' @param lagged_df Data frame of lagged data as in `create_train_and_predict_matrices()`.
+#' @param ahead Number of days ahead to forecast. Can be a single positive 
+#'     integer or a vector of positive integers.
+#' 
+#' @return Single date corresponding to the last possible date of the training
+#' data.
```
dajmcdon, line 146: I appreciate the comments, but my memory is that Roxygen2 gets converted to an .Rd file even if the function is not exported. If that's the case, better to convert `#'` to `#`.

R-packages/modeltools/R/matrix.R
```diff
@@ -87,84 +82,133 @@ create_train_and_predict_matrices <- function(lagged_df, ahead, training_window_
...
-            return(all_out[[1]])
-        } else {
-            return(all_out)
-        }
-    } else {
-        # We want a single training data matrix and a matrix of 
-        # training responses containing all the aheads
+        all_out <- purrr::map2(ahead, train_end_dates,
+                              ~ create_train_matrices(lagged_df, .x, training_window_size,
+                                                      .y))
```
dajmcdon, line 103: Weird multiline formatting. Maybe a return before `lagged_df`?

R-packages/modeltools/R/matrix.R
```diff
@@ -87,84 +82,133 @@ create_train_and_predict_matrices <- function(lagged_df, ahead, training_window_
...
-            
-            # Add prediction matrices and training end date
-            out$predict_x <- predict_x
-            out$predict_geo_values <- predict_geo_values
-            out$train_end_date <- train_end_date
-            
-            all_out[[paste0("ahead+", a)]] <- out
-        }
+        train_end_dates <- Reduce(c, lapply(ahead, 
+                                            function(a) get_train_end_date(lagged_df, a)))
```
dajmcdon, line 99: Also weird multiline behavior. Maybe a return before `lapply`? Can this be done with `sapply` instead? Or `purrr::map_dbl() %>% lubridate::ymd()`?

dajmcdon (approved):

dajmcdon:
Just a few minor changes.

R-packages/modeltools/R/matrix.R
kjytay: Yes, this does get converted to an .Rd file. Personally I prefer more documentation than less, even for internal functions that are not exported (the glmnet package does this as well). But if you prefer not to have the .Rd file I can change it. Let me know.

R-packages/modeltools/R/matrix.R
kjytay: Done.

R-packages/modeltools/R/matrix.R
kjytay: `apply` and `sapply` don't work here because they coerce dates into numerics. The `Reduce` code was a solution I found here (https://stackoverflow.com/questions/46357983/r-apply-function-returns-numeric-value-on-date-variables). I've replaced it with `purrr::map_dbl() %>% lubridate::as_date()`, perhaps that is more readable.

PR #521, opened 2021-04-02
Update the condition for epidata in the response

chinandrew:
The Flask API is returning `{'epidata': [], 'result': -2, 'message': 'no results'}` instead of `{'result': -2, 'message': 'no results'}`, so this check would pass half the time and lead to downstream problems. Open discussion on the flask/epidata side, but this might be a good robustification regardless.

### DIFF
--- a/Python-packages/covidcast-py/covidcast/covidcast.py
+++ b/Python-packages/covidcast-py/covidcast/covidcast.py
@@ -184,6 +184,8 @@ def signal(data_source: str,
         dfs = _fetch_epidata(
             data_source, signal, start_day, end_day, geo_type, geo_values, as_of, issues, lag
         )
+    print(dfs)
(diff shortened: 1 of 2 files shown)


krivard (approved):
👍

krivard:
Tagging in @statsmaths, who should probably start taking over from @capnrefsmmat here

chinandrew:
Also realized that another way to do this is to convert the whole block to `if...elif...else...`(and probably switch to result codes)

krivard:
ehhhhh result codes would be cleaner and less brittle to any future changes we make for the not-a-number project, but that change doesn't have to be now.

capnrefsmmat (approved):
now that Flask is deployed, do we need to cut a release to include this?

PR #527, opened 2021-04-21
Alternate implementation of get_forecaster_predictions() for dashboard

eujing:
Added a new function `get_forecaster_predictions_alt()` with the same functionality as `get_forecaster_predictions()`, but with reduced memory usage with the side effect of downloading CSVs to disk. A new named parameter `use_disk = TRUE` can be passed to `get_covidhub_predictions()` to use this alternate implementation, otherwise the original is used by default.

Details of the alternate implementation:
1) For the specified forecaster, download all the required CSVs to disk
2) Load just the `target` column from all the CSVs using the `arrow` package
3) Pre-process just the distinct `target` values with `separate` and `mutate` for the derived columns to make a small table `target_separated`
4) Join the full data set with `target_separated`


dajmcdon (approved):
This looks fine to me. It has the desired behavior?

PR #529, opened 2021-04-22
Switch away from arrow's type parsing

eujing:
Fixes errors that appear with some forecasters such as:
```
Invalid: Could not open CSV input source 'data/IQVIA_ACOE-STAN/2020-08-23/data.csv': Invalid: In CSV column #5: CSV conversion error to double: invalid value 'NA'
Invalid: Could not open CSV input source 'data/BPagano-RtDriven/2020-10-25/data.csv': Invalid: In CSV column #2: CSV conversion error to date32[day]: invalid value '2020-11-7'
```
Arrow could not natively parse quoted NA (`"NA"`) as doubles, and non-zero padded dates.

Only affects the new `get_forecaster_predictions_alt()` function.


R-packages/evalcast/R/get_covidhub_predictions.R
```diff
@@ -384,7 +384,11 @@ get_forecaster_predictions_alt <- function(covidhub_forecaster_name,
       collect() %>%
       left_join(target_separated, by = "target") %>%
       select(-.data$target) %>%
-      mutate(forecaster = covidhub_forecaster_name) %>%
+      mutate(forecaster = covidhub_forecaster_name,
+             forecast_date = as.Date(forecast_date, format = "%Y-%m-%d"),
+             target_end_date = as.Date(target_end_date, format = "%Y-%m-%d"),
```
dajmcdon, line 389: Can you use `lubridate::ymd()` for these? Seems a bit more robust. (I think `ymd()` is already imported. 

Also, when you mutate, it can change the ordering of the columns. Can you ensure that they remain in the same order?

dajmcdon (changes requested):

dajmcdon (approved):

PR #537, opened 2021-05-04
Final CRAN modifications

capnrefsmmat:
These are the changes I made both before our CRAN submission and during two rounds of review, leading to our [acceptance onto CRAN](https://cran.r-project.org/package=covidcast). (Yay!)

Mostly requesting review pro forma so you see what changed. Then I propose we rebuild all the docs, merge this into r-pkg-devel, and then merge r-pkg-devel to main so the docs update to describe the CRAN version and no longer recommend using devtools to install.

### DIFF
--- a/R-packages/covidcast/DESCRIPTION
+++ b/R-packages/covidcast/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: covidcast
 Type: Package
 Title: Client for Delphi's 'COVIDcast Epidata' API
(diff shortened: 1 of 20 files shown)


statsmaths:
That's great news! Glad you managed to get it up on CRAN. All of the changes seem good to me. Let's merge into main so that users can be directed to the new link as soon as possible.

statsmaths (approved):

PR #540, opened 2021-05-04
Deploy covidcast 0.4.2 to main

capnrefsmmat:


capnrefsmmat:
Since this is already on CRAN and I've verified the only changes are in the R package, I'm going to use my super admin powers and merge.

PR #544, opened 2021-05-10
Prepare covidcast-py v0.1.5 for release

capnrefsmmat:
This means #531 and #521 will become available in the published version.

I also bumped the copyright year and fixed an outdated link while I was here.


docs/covidcast-py/html/_static/language_data.js
```diff
@@ -5,16 +5,15 @@
  * This script contains the language-specific data used by searchtools.js,
  * namely the list of stopwords, stemmer, scorer and splitter.
  *
- * :copyright: Copyright 2007-2021 by the Sphinx team, see AUTHORS.
+ * :copyright: Copyright 2007-2020 by the Sphinx team, see AUTHORS.
```
chinandrew, line 8: i wonder why these went backwards...

PR #557, opened 2021-07-16
Fcast all hands talk

dajmcdon:


dshemetov:
This should probably move to the [talks repo](https://github.com/cmu-delphi/covidcast-talks/), since we don't store talks here anymore.

krivard:
Moved to covidcast-talks ([commit](https://github.com/cmu-delphi/covidcast-talks/commit/428bf3d013371190ad202fcf606edb24a2e96701))

PR #558, opened 2021-08-06
Support the new API endpoint and its higher row limit

capnrefsmmat:
The new endpoint supports a much higher row limit (1 million). This will
speed up API access dramatically by batching requests into fewer queries
-- generally only one, even for dense signals over large time ranges.

I had some vdiffr test failures, so I'll let the CI run before requesting review.


capnrefsmmat:
@krivard This is the pull request I mentioned in the meeting. I don't understand why the unit test is failing (`test-covidcast.R` line 265), because it passes locally. If you have someone with R background who can spend a couple hours testing locally, that would help.

nmdefries:
@qx-teo, @statsmaths, and I looked into the test failure. None of us was able to get the tests to pass locally with the code as-is. However, all tests pass after directly setting `MAX_RESULTS` to 3650 in `covidcast.R`, suggesting that the issue is the deep assignment `MAX_RESULTS <<- 3650`. That line doesn't seem to actually be updating the global variable within the package.

Rather than using `<<-` which is opaque and relies on `testthat` having particular environment behavior, "[i]t would be much better to provide some (semi-)direct access to `MAX_RESULTS`. Perhaps checking to see if the user has set the option `covidcast.MAX_RESULTS`, and if they have and the number is less than 1e6, use that instead" (recommendation from discussion with Taylor).

capnrefsmmat:
Thanks for looking into this, everyone. I can believe that it's something weird related to the deep assignment and however environments are handled in testthat and R CMD CHECK. I'll switch this to use an `option`.

capnrefsmmat:
Looks like the updated version passes the covidcast tests on CI now. (evalcast and covidcast-py test failures are unrelated.)

@statsmaths, are you able to review the current version so we can proceed with releasing it?

PR #562, opened 2021-09-21
Update quantgen dependency folder structure

dshemetov:
[Quantgen](https://github.com/ryantibs/quantgen/tree/master/quantgen) recently reorganized its folder structure, which makes the installation
```
devtools::install_github("cmu-delphi/covidcast", ref="modeltools", subdir="R-packages/modeltools")
```
error out with a 404 when installing the quantgen dependency.

This PR should fix the ref. Tested locally by a manual git clone with
```
devtools::install("src/covidcast/R-packages/modeltools")
```

### DIFF
--- a/R-packages/modeltools/DESCRIPTION
+++ b/R-packages/modeltools/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: modeltools
 Type: Package
-Title: Tools for Building COVID Forecasters and Related Models 
+Title: Tools for Building COVID Forecasters and Related Models
 Version: 0.1.0
 Authors@R:
 	c(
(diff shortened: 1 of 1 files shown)


brookslogan:
Destination branch seems right, AFAIK.

File changes look good.

Test looks good:
```
dev.mode.dir = "dev-mode-dir-covidcast-pr-562"
if (!dir.exists(dev.mode.dir)) {
  dir.create(dev.mode.dir)
} else {
  stop ('only expected to run this once')
}

## XXX quantgen should not already be installed!

devtools::dev_mode(TRUE, dev.mode.dir)

## this should currently fail
devtools::install_github("cmu-delphi/covidcast", ref="modeltools", subdir="R-packages/modeltools")
## and it does 404

## this should work
devtools::install_github("cmu-delphi/covidcast", ref="quantugen_update", subdir="R-packages/modeltools")
## and it does get past that point (didn't wait for full install, s2 installation for sf for covidcast for modeltools was taking forever)

devtools::dev_mode(FALSE, dev.mode.dir)
```

LGTM

PR #563, opened 2021-09-24
Remove duplicate join in `evaluate_covid_predictions`

nmdefries:
### Description
Currently, truth data is joined on to `predictions_cards` in both `evaluate_covid_predictions`, before calling `evaluate_predictions`, and in `evaluate_predictions`. `evaluate_predictions` is passed separate `actual_data` and doesn't use the truth data contained in `predictions_cards`, so the first join is doing unnecessary work. Remove it, but enforce the same column ordering.

### Changes
- Remove join in `evaluate_covid_predictions` that adds on truth data.
- Pass `grp_vars` to `evaluate_predictions`.


dajmcdon (commented):
approved

PR #564, opened 2021-09-24
Fix failing `zoltr` dependency

nmdefries:


krivard (approved):
format lgtm, no clue if it will work though

PR #565, opened 2021-09-27
Fix failing `zoltr` dependency on `evalcast` branch

nmdefries:
### Description
Get `zoltr` to install.

Note: failing python linting is unrelated to this. The two `evalcast` failures (`Error: unused argument (count_median_twice = FALSE)`) also predate this change.

### Changes
- Use the workflow's token as the GITHUB_PAT to get around rate limits.

### Fixes
`zoltr` has been failing to be installed. This seems to have been a mid-August change in rate limits for non-registered users downloading repos. Used the fix discussed in r-lib/remotes#638.


nmdefries:
Superseded by #573.

PR #567, opened 2021-10-03
Improve speed of process_target by only processing unique targets.

jacobbien:
This leads to roughly a 10x speed improvement for `get_covidhub_predictions` in the following example:

```
library(lubridate)
library(evalcast)
dates <- seq(ymd("2020-07-27"), ymd("2021-09-01"), by = 7)
system.time({pred <- get_covidhub_predictions("COVIDhub-ensemble",
                                              forecast_dates = dates,
                                              ahead = 4,
                                              incidence_period = "epiweek",
                                              signal = "deaths_cumulative_num")})
```

Before the change:

```
   user  system elapsed 
118.291   9.533 167.169 
```

After the change:

```
   user  system elapsed 
 11.724   3.901  29.252 
```

The idea is to do less string processing by first getting the unique target strings, then doing the processing, and finally doing a `left_join`.


jacobbien:
Note: Three unit tests are failing on the current version of evalcast on main.  This PR doesn't lead to any additional tests failing.

dajmcdon:
Same query using `zoltar`:

```{r}
> system.time({preds <- get_zoltar_predictions("COVIDhub-ensemble",
+                                              forecast_dates = dates,
+                                              ahead = 4,
+                                              incidence_period = "epiweek",
+                                              signal = "deaths_cumulative_num")})
get_token(): POST: https://zoltardata.com/api-token-auth/
get_resource(): GET: https://zoltardata.com/api/projects/
get_resource(): GET: https://zoltardata.com/api/project/44/timezeros/
   user  system elapsed                                                        
  2.995   0.253  20.649 
```

I also tried doing something like this by grabbing the predictions from the S3 bucket (no cumulatives there). That takes quite a bit longer and way more memory because you first download all predictions for everything and then filter. But I wonder at what point we should just put these in our API?

dajmcdon (approved):

dajmcdon:
I'm not sure what's failing. The `evalcast` check looks ok above, but there's a python failure? Also, we should definitely bump the R build we're using (3.5???)

jacobbien:
Good to know that `get_zoltar_predictions` is the fastest.  Would it make sense to make that the default then?  Or to mention in `?get_covidcast_predictions` that `get_zoltar_predictions` may be faster?

nmdefries (approved):

PR #572, opened 2021-11-05
Revert "Remove duplicate join in `evaluate_covid_predictions`"

nmdefries:
Reverts cmu-delphi/covidcast#563

Above change causes the forecast eval pipeline to fail with `Error: weighted_interval_score: quantiles must be unique.`. Reverting so we have a working pipeline for next week while I debug.


dajmcdon (approved):
Ugh.

PR #580, opened 2022-02-02
Add download_signals caching to evalcast

dshemetov:
# Features

This PR adds extra functionality to `download_signals` to let it cache downloaded signals locally. This can save a lot of development time when calling evalcast many times within a short time period (one the order of a day). By default there is no caching, so this is backwards compatible with old code. This functionality is exposed and threaded (with a NULL default) through:
- `get_predictions`
- `evaluate_covid_predictions`
- `get_covidcast_data`
- `get_target_response`

Minor features: 
- adds a `parallel_execution=TRUE` flag to get_predictions, so that each `forecast_date` is run on a separate core
- adds an `honest_as_of=TRUE` which ensures that `forecast_date = as_of = end_date` when making predictions.

## Details

By specifying a cache directory `offline_signal_dir=cache_path`, `download_signals` will store its call to `covidcast_signal` in the file `cache_path/%g/%S_%s_%e_%a` where 
- `%g` is the geo_type (e.g. 'state')
- `%S` is the data source
- `%s` is the signal
- `%e` is the end day date
- `%a` is the as of date

The next time `download_signals` is called with the same arguments as above, the data will instead be read from that file. 

Potential issues:
- standard stale cache issues - if you run this before a day's acquisitions are finalized, your cache may be stale tomorrow (a warning is included for the case when `end_day` is later than the latest timestamp in the data)
- we assume the `start_day` argument stays constant, so the same cache file will be used if the other arguments are the same but `start_day` changes (a warning is included when `start_day` is earlier than the earliest `start_day` in the data)
- clearing cache needs to be done manually (i.e. delete the cache folder)
- using relative cache paths is ok, but might leave a trail of caches when running code from inconsistent directories


dajmcdon:
@dshemetov @nmdefries I just approved #581, but it hasn't merged yet. But that one is needed to support the dashboard. So we need to be sure that they are compatible before merging this.

dshemetov:
@nmdefries What is the behavior of `download_signal` when the download fails? Does `covidcast_signal` raise an error? If so, then I think this wrapper should work seamlessly with this PR, just feed the `offline_signal_dir` to the wrapper as you would the regular function. Otherwise things would get more complicated.

R-packages/evalcast/R/get_predictions.R
```diff
@@ -69,22 +81,36 @@ get_predictions <- function(forecaster,
...
+              apply_corrections = apply_corrections,
+              forecaster_args = forecaster_args,
+              honest_as_of = honest_as_of,
+              offline_signal_dir = offline_signal_dir))
+  }
+  if(parallel_execution) {
+    out <- mclapply(
+      forecast_dates, 
+      get_predictions_single_date_, 
+      mc.cores=max(detectCores()-1, 1)
```
brookslogan, line 108: `mc.cores=max(detectCores()-1, 1)` should be something like
- `getOption("mc.cores", 2L)` --- the default value of this argument
- `getOption("mc.cores", max(detectCores()-1, 1))` --- uses the current value unless the user overrides beforehand using `options(mc.cores=<value>)`
- changing `get_forecasts` to take an integer number of cores rather than a Boolean whether to use parallelism

dshemetov: In favor of the second one, so the user doesn't have to figure out their core number by default, but can also set `mc.cores` manually.

R-packages/evalcast/R/download_signals.R
```diff
@@ -1,30 +1,80 @@
...
+#' (signal, forecast day) pair. If this is null, no caching is done and the data is
+#' downloaded from covidcast.
 #' @param ... arguments to be passed to `covidcast::covidcast_signal()`.
 #' @return `covidcast_signal` data frame.
 #'
-download_signal <- function(...) {
+#' @importFrom fs dir_create
+#'
+#' @export
+download_signal <- function(offline_signal_dir=NULL, ...) {
```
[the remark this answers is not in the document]
dshemetov: Makes sense, didn't realize that `...` could go in front of other keyword args.

R-packages/evalcast/R/get_predictions.R
brookslogan: The third option could also be modified to have some user input value mean "auto-select some number of cores for me greater than 1 if possible".  Then we'd avoid the need to go through the side-effectful `options` interface while allowing the user to not have to figure out the core number.  But this is just trading one confusing/complicated thing for another, unless you can think of a good name and scheme for such a parameter for option 3'.  If you can think of a good configuration for 3', that might be preferable.

If proceeding with option 2, it'd be good to mention the possibility of using `options(mc.cores=<...>)` (or `withr::with_options(......)`) in the function's documentation.

R-packages/evalcast/R/download_signals.R
```diff
@@ -1,30 +1,80 @@
...
+    }
+
+    max_time_value <- (df %>% summarize(max(time_value)))[[1]]
+    if (read_from_disk && (max_time_value < args$end_day)) warning("Data in cache ends earlier than `end_day`.")
+
+    if (is.null(args$start_day)) {
+      df <- df %>% filter(time_value <= args$end_day)
+    } else {
+      min_time_value <- (df %>% summarize(min(time_value)))[[1]]
+      if (read_from_disk && (min_time_value > args$start_day)) warning("Data in cache starts later than `start_day`.")
```
brookslogan, line 64: It might be good to use `rlang`'s utilities to make classed warnings, etc., for the two warnings above, to help users more easily filter them out.

R-packages/evalcast/R/get_predictions.R
dshemetov: What if `parallel_execute` is a mixed Boolean+Integer type, that if TRUE defaults to `detectCores()-1` and if an integer then uses that many cores?

dajmcdon:
I think we this PR suggests making a fundamental change. The most public use of evalcast is the dashboard, so nothing we do can break that. But I suspect the dashboard needs rather little of the functionality. @nmdefries would you be able to create a branch that contains the minimally necessary content to provide the dashboard? Or even put it in a separate repo (one that's lightweight and fast to install from)? That would allow this branch to handle forecast team duties without any risk of breakage.

R-packages/evalcast/R/download_signals.R
```diff
@@ -1,30 +1,90 @@
...
+      dir_create(dirname(signal_fpath), recurse = TRUE)
+      saveRDS(df, signal_fpath)
+      read_from_disk <- FALSE
+    }
+
+    max_time_value <- (df %>% summarize(max(time_value)))[[1]]
+    if (read_from_disk && (max_time_value < args$end_day)) warn("Data in cache ends earlier than `end_day`.", "evalcast::download_signals:cache_warning_end_day")
+
+    if (is.null(args$start_day)) {
+      df <- df %>% filter(time_value <= args$end_day)
```
brookslogan, line 71: Refactoring: might be better to move this, and the corresponding part of line 75, outside the branching on `is.null(args$start_day)`.  Also, if `end_day` is made optional (in response to a note up above), then this filtering step will fail if they do not provide it; the fix would make the `end_day` filtering look very similar to the `start_day` filtering.  We may also need to also check that if they provide both `start_day` and `end_day`, that `start_day <= end_day`.

R-packages/evalcast/R/download_signals.R
```diff
@@ -1,30 +1,90 @@
...
+        sprintf("Downloading signal from API and storing in: %s", signal_fpath),
+        "evalcast::download_signals:cache_info_write"
+      )
+      df <- suppressMessages({covidcast_signal_wrapper(...)})
+      dir_create(dirname(signal_fpath), recurse = TRUE)
+      saveRDS(df, signal_fpath)
+      read_from_disk <- FALSE
+    }
+
+    max_time_value <- (df %>% summarize(max(time_value)))[[1]]
```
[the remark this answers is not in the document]
brookslogan: Another very minor thing: we could slightly improve error messaging when the user requests completely invalid/missing data:
```
download_signal(data_source="jhu-csse",signal="confirmed_incidence_num",geo_type="state",geo_values="invalid", end_day=Sys.Date()+106L, start_day=NULL, offline_signal_dir=tempdir())

Error: Problem with `summarise()` input `..1`.
ℹ `..1 = max(time_value)`.
✖ object 'time_value' not found
Run `rlang::last_error()` to see where the error occurred.
In addition: Warning message:
Fetching confirmed_incidence_num from jhu-csse for 2020-01-22 to 2022-05-28 in geography 'invalid': no results 
```
I think the warning message makes it clear what has gone wrong.

R-packages/evalcast/R/get_predictions.R
```diff
@@ -69,22 +83,36 @@ get_predictions <- function(forecaster,
                             apply_corrections = function(signals) signals,
                             response_data_source = signals$data_source[1],
                             response_data_signal = signals$signal[1],
-                            forecaster_args = list()) {
-  
+                            parallel_execution = TRUE,
+                            forecaster_args = list(),
```
brookslogan, line 87: For backwards compatibility with code calling args positionally, might want to swap the order of `parallel_execution` and `forecaster_args` here.

brookslogan (changes requested):
Major:
- There are some caching issues when the user selects different `geo_values` in a sequence of `download_signal` operations.
- Somehow running `get_predictions` in non-parallel mode followed by parallel mode triggers errors.  Not sure what is going on.

Minor:
- Various messaging, styling, and testing things.

R-packages/evalcast/R/get_predictions.R
dshemetov: Implemented the above idea in the latest commit.

R-packages/evalcast/R/download_signals.R
```diff
@@ -1,30 +1,90 @@
...
+  if (is.null(args$as_of)) args$as_of <- Sys.Date()
+  if (is.null(args$geo_type)) args$geo_type <- "county"
+  if (!is.null(offline_signal_dir) & is.null(args$end_day)) stop("end_day must be set.")
+
+  if (is.null(offline_signal_dir)) {
+    if (is.null(args$start_day)) {
+      msg <- stringr::str_glue(
+        "Downloading {signal} from covidcast through {end}.",
+        end = args$end_day,
+        signal = args$signal,
```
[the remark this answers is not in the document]
dshemetov: First, `download_signals` is never called without `end_day` or `signal` in evalcast. Second, these code lines are already in `download_signals` and I think fixing old code is out of scope for this PR.

R-packages/evalcast/R/download_signals.R
```diff
@@ -1,30 +1,90 @@
...
+        .sep = " "
+      )
+    }
+
+    message(msg)
+    df <- suppressMessages({covidcast_signal_wrapper(...)})
+  } else {
+    signal_fpath <- file.path(offline_signal_dir, args$geo_type, sprintf("%s_%s_%s_%s.RDS", args$data_source, args$signal, args$end_day, args$as_of))
```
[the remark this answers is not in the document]
dshemetov: To the first comment, not sure we can do much about guarding a user using positional arguments aside from explicitly writing the arguments instead of using the ellipsis.

To the second, you're right, it's a no-op at this point. I like the idea of having the file tagged with the as_of date instead and then the end_day can actually filter values if end_day < as_of.

R-packages/evalcast/R/download_signals.R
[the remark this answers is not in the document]
brookslogan: Okay, maybe something to file away in a new issue.

R-packages/evalcast/R/download_signals.R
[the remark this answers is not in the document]
brookslogan: Regarding the first: we move from having a bunch of arguments hidden in the dots to explicitly having them as parameters, or check for unnamed arguments and throw a more descriptive error.

Regarding the second: I like the idea of tagging with as_of date only.  But I think the post-cache-fetch filter would need to be something like .data$time_value < .env$end_day, not involving as_of at all.

R-packages/evalcast/R/download_signals.R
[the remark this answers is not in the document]
dshemetov: I think here covidcast_signal returned nothing and therefore time_value couldn't be found as a column. I added a check to see if df is empty before doing filtering ops.

R-packages/evalcast/R/get_predictions.R
```diff
@@ -69,22 +83,36 @@ get_predictions <- function(forecaster,
...
+            list(forecaster = forecaster,
+              signals = signals,
+              forecast_date = forecast_date,
+              apply_corrections = apply_corrections,
+              forecaster_args = forecaster_args,
+              honest_as_of = honest_as_of,
+              offline_signal_dir = offline_signal_dir))
+  }
+  if(parallel_execution) {
+    out <- mclapply(
```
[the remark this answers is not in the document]
dshemetov: I'm going to remove the progress bar. It's not critical.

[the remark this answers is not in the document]
dshemetov: However, even with the bar present, I wasn't able to reproduce your error...

```
r$> evalcast::get_predictions( 
        forecaster = evalcast::baseline_forecaster, 
        name_of_forecaster = "baseline", 
        signals = forecaster_signals[["baseline1_forecaster"]], 
        forecast_dates = seq(as.Date("2020-12-01"), as.Date("2020-12-31"), by=7), 
        incidence_period = "epiweek", 
        parallel_execution = FALSE, 
        forecaster_args = list(ahead = 3) 
    )                                                                                                                                                                                                                                                                      
Downloading confirmed_admissions_covid_1d from covidcast for period from 2020-08-01 to 2020-12-01.
Downloading confirmed_admissions_covid_1d from covidcast for period from 2020-08-01 to 2020-12-08.
Downloading confirmed_admissions_covid_1d from covidcast for period from 2020-08-01 to 2020-12-15.
Downloading confirmed_admissions_covid_1d from covidcast for period from 2020-08-01 to 2020-12-22.
Downloading confirmed_admissions_covid_1d from covidcast for period from 2020-08-01 to 2020-12-29.
...
Warning messages:
1: Data not fetched for the following days: 2020-11-29, 2020-11-30, 2020-12-01 
2: Data not fetched for the following days: 2020-12-06, 2020-12-07, 2020-12-08 
3: Fetching confirmed_admissions_covid_1d from hhs for 2020-12-13 to 2020-12-15 in geography '*': no results 
4: Data not fetched for the following days: 2020-12-21, 2020-12-22 
5: Data not fetched for the following days: 2020-12-28, 2020-12-29 

r$> evalcast::get_predictions( 
        forecaster = evalcast::baseline_forecaster, 
        name_of_forecaster = "baseline", 
        signals = forecaster_signals[["baseline1_forecaster"]], 
        forecast_dates = seq(as.Date("2020-12-01"), as.Date("2020-12-31"), by=7), 
        incidence_period = "epiweek", 
        parallel_execution = TRUE, 
        forecaster_args = list(ahead = 3) 
    )                                                                                                                                                                                                                                                                      
Downloading confirmed_admissions_covid_1d from covidcast for period from 2020-08-01 to 2020-12-01.
Downloading confirmed_admissions_covid_1d from covidcast for period from 2020-08-01 to 2020-12-08.
Downloading confirmed_admissions_covid_1d from covidcast for period from 2020-08-01 to 2020-12-15.
Downloading confirmed_admissions_covid_1d from covidcast for period from 2020-08-01 to 2020-12-22.
Downloading confirmed_admissions_covid_1d from covidcast for period from 2020-08-01 to 2020-12-29.
...
```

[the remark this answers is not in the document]
dshemetov: Added the following to catch NULL results

```
get_predictions_single_date_ <- function(forecast_date) {
    preds <- do.call(get_predictions_single_date,
      list(
        forecaster = forecaster,
        signals = signals,
        forecast_date = forecast_date,
        apply_corrections = apply_corrections,
        forecaster_args = forecaster_args,
        honest_as_of = honest_as_of,
        offline_signal_dir = offline_signal_dir
      )
    )
    if (is.null(preds)) abort("A child process was terminated and returned a null .")
    return(preds)
  }
```

Though, according to [the docs](https://search.r-project.org/CRAN/refmans/bettermc/html/mclapply.html), a fatal child event will make mclapply fail by default.

R-packages/evalcast/R/download_signals.R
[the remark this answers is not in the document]
dshemetov: I addressed these issues in a recent update.

R-packages/evalcast/R/download_signals.R
```diff
@@ -1,30 +1,90 @@
...
+      )
+      df <- readRDS(signal_fpath)
+      read_from_disk <- TRUE
+    }
+    else {
+      inform(
+        sprintf("Downloading signal from API and storing in: %s", signal_fpath),
+        "evalcast::download_signals:cache_info_write"
+      )
+      df <- suppressMessages({covidcast_signal_wrapper(...)})
```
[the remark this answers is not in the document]
dshemetov: I'm going to do the simple thing of just downloading everything and filtering down to the selected geo_values. If we want something more intelligent, we can put it in a separate enhancement PR.

R-packages/evalcast/R/get_predictions.R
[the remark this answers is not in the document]
brookslogan: I don't think this would catch the NULLs, as it is inside the function run by mclapply.  The issue I've had is that NULLs appear within the list returned by mclapply, when the function would never return NULL regularly, due to the child process being killed for memory.

The docs you linked are for `bettermc`, not `parallel`, which looks like it's designed to fix the problem!  Are we actually set up to use this package right now?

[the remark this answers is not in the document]
dshemetov: Oh jeez, I didn't even notice I linked to a wrapper for `parallel`. We can probably just drop it in, since we don't depend on complicated uses of mclapply.

R-packages/evalcast/R/get_predictions.R
```diff
@@ -86,6 +86,17 @@ get_predictions <- function(forecaster,
...
 
+  if (is.logical(parallel_execution) & parallel_execution == TRUE) {
+    num_cores <- max(1, parallel::detectCores() - 1)
+  } else if (is.logical(parallel_execution) & parallel_execution == FALSE) {
+    num_cores <- 1
+  } else if (is.integer(parallel_execution)) {
+    num_cores <- max(1, min(parallel::detectCores(), parallel_execution))
+  }
+  else {
+    stop("parallel_execution argument is neither logical nor integer.")
```
brookslogan, line 97: Technically, maybe "Boolean" is better than logical here (logical could be NA).

brookslogan:
Some very minor comments above.  Why change to not export `download_signal`?  If this is now a cached download again, I imagine it would be quite useful to get cached downloads without having to fake out `get_predictions`.

R-packages/evalcast/R/download_signals.R
```diff
@@ -1,30 +1,104 @@
...
+    if (is.null(start_day)) msg <- str_glue("{msg} the earliest date possible to", msg = msg)
+    else msg <- str_glue("{msg} {start_day} to", msg = msg, start_day = start_day)
+
+    if (is.null(end_day)) msg <- str_glue("{msg} the latest date possible.", msg = msg)
+    else msg <- str_glue("{msg} {end_day}", msg = msg, end_day = end_day)
+
+    message(msg)
+    df <- suppressMessages(covidcast_signal_wrapper(data_source = data_source, signal = signal, start_day = start_day, end_day = end_day, geo_type = geo_type, geo_values = geo_values, as_of = as_of, ...))
+  } else {
+    signal_fpath <- file.path(offline_signal_dir, sprintf("%s_%s_%s_%s.RDS", data_source, signal, geo_type, as_of))
```
brookslogan, line 45: Does `time_type` need to be here too?  Or is it limited to be `"day"` somewhere else?

dshemetov (commented):
overall looks good to me, mostly adding comments about minor knits, possible refactors, and a few cautions (biggest one is making sure the load table is empty before deleting). seconding Andrew's comment about a lot of TODOs.

R-packages/evalcast/R/download_signals.R
dshemetov: Looking at `get_predictions` it looks like there are no protections for incidence period being day or week. It's all treated the same way. However, `get_predictions_single_date` has a line that converts its `forecast_date` argument using `lubridate::ymd(forecast_date)`, which seems like it has a hidden assumption that the provided dates are not epiweeks. That codepath seems _very_ shaky and we're probably just lucky that no one ever used `get_predictions` with epiweeks. 

To address this, we can probably include time_type, but treat caching in the same way.

dshemetov:
That's a fair point, I'll export `download_signals` again.

R-packages/evalcast/R/download_signals.R
brookslogan: Right, IIRC `get_predictions` weekly predictions are based on `time_type="day"` and `incidence_period="week"`, and probably would blow up with `time_type="week"`.  So either the recording `time_type` or hard-coding `time_type="day"` seems sufficient.

R-packages/evalcast/R/download_signals.R
```diff
@@ -1,30 +1,104 @@
...
+#' @importFrom rlang inform warn
+#' @importFrom fs dir_create
+#' @importFrom magrittr %>%
+#' @importFrom plyr empty
+download_signal <- function(data_source, signal, start_day = NULL, end_day = NULL, geo_type = "county", geo_values = "*", as_of = NULL, offline_signal_dir = NULL, ...) {
+  if (is.null(geo_type)) geo_type <- "county"
+  if (is.null(geo_values)) geo_values <- "*"
+  if (is.null(as_of)) as_of <- Sys.Date()
+  if (is.null(end_day)) end_day <- max(as_of)
+  if (!is.null(start_day)) assert_that(as.Date(start_day) < as.Date(end_day))
```
brookslogan, line 31: should be `<=`

brookslogan: (otherwise this breaks `evaluate_covid_predictions`)

brookslogan:
I implemented some of the minor comments + a way to pass additional args to bettermc::mclapply.  Wasn't able to push directly, but it's up [here](https://github.com/brookslogan/covidcast/tree/evalcast-download-caching).

brookslogan:
Other than the `evaluate_covid_predictions` error from the `start_day < end_day` check and the weird streams not closed cleanly errors (that I can't reproduce now), all experimentation with the caching & parallel changes was successful!

R-packages/evalcast/R/download_signals.R
dshemetov: It looks like `covidcast::covidcast_signal` takes only YYYYMMDD format dates for `start_day` and `end_day`, so I'm going to keep those same requirements here. However in the comparison below, I'm going to convert `start_day` and `end_day` to epiweek format so we can do the time comparison with the response (I'm assuming the response time_value is in epiweek format, will need to double check).

brookslogan: I'm guessing that evalcast will break elsewhere if you get `time_type="week"` downloads and caching working, so `get_forecasts` will need to check for and complain if `time_type="week"` to break "nicely" with a complaint.  If it seems like substantial work to get the weekly downloading or caching working, might as well just complain there instead if someone tries downloading `time_type="week"` data (it wouldn't be `get_predictions` requesting it --- earlier I forgot how incidence_period="epiweek" worked).

dshemetov: FWIW, I just tested this and it turns out that `time_type='week'` signals still return regular dates.

```r
r$> d <- covidcast::covidcast_signal("nchs-mortality", "deaths_allcause_incidence_num", geo_type = "state", start_day = "2021-01-01", time_type = "week")
r$> d %>% tibble
# A tibble: 3,744 × 12
   data_source    signal                        geo_value time_value issue        lag missing_value missing_stderr missing_sample_size value stderr sample_size
   <chr>          <chr>                         <chr>     <date>     <date>     <int>         <int>          <int>               <int> <dbl> <lgl>  <lgl>      
 1 nchs-mortality deaths_allcause_incidence_num ak        2020-12-27 2021-10-31    44             0              1                   1   105 NA     NA         
 2 nchs-mortality deaths_allcause_incidence_num al        2020-12-27 2021-11-21    47             0              1                   1  1804 NA     NA         
 3 nchs-mortality deaths_allcause_incidence_num ar        2020-12-27 2021-11-28    48             0              1                   1   967 NA     NA
```

But sure, if you think that evalcast was just never really intended to be used with `time_type='week'` and it hasn't been, then we can just make that explicit by having it fail with an explanation.

brookslogan: I think it actually silently ignores `time_type="week"` I believe (any `time_type` arg doesn't make its way to `download_signal`).  For weekly stuff the `evalcast` user was expected to use `incidence_period="epiweek"`.

dshemetov:
Alright, I've made some updates myself and merged with your changes. Should be good to go!

R-packages/evalcast/R/get_predictions.R
```diff
@@ -69,79 +82,112 @@ get_predictions <- function(forecaster,
...
+                            additional_mclapply_args = list(),
+                            honest_as_of = TRUE,
+                            offline_signal_dir = NULL) {
+
+  assert_that(is_tibble(signals), msg = "`signals` should be a tibble.")
+  assert_that(xor(honest_as_of, "as_of" %in% names(signals)), msg = "`honest_as_of` should be set if and only if `as_of` is not set. Either remove as_of specification or set honest_as_of to FALSE.")
+  incidence_period <- match.arg(incidence_period)
+
+  if (incidence_period == "epiweek") rlang::warn("incidence_period of 'epiweek' only selects weekly (Saturday) target end dates, actual 'epiweek' signals not supported.", "evalcast::get_predictions")
+  if (!is.null(signals$time_type) & (signals$time_type == "week" || signals$time_type == "epiweek")) rlang::abort("time_type in the signals arg can only be 'day'.", "evalcast::get_predictions")
```
brookslogan, line 96: tibble's generating some warnings in the exploration script: "Unknown or uninitialised column: `time_type`."

Background: `data.frame`'s `$` has some undesirable behaviors when programming:
```
data.frame(aaa=1:5)$aaa # 1:5, good
data.frame(aaa=1:5)$b # NULL, "convenient"
data.frame(aaa=1:5)$a # 1:5, "convenient"
data.frame(aaa=1:5, aab=2:6)$a # NULL, ..............
data.frame(time_type_NOT = c("a","b"))$time_type # "a" "b" rather than NULL; unlikely to come up with time_type but a concern for shorter names
```
Now, `tibble` tries to make this better behaved.  The partial matching behavior was not common(ly intended to be used in) code and brains, but  `is.null(df$col)` was.  I imagine this is why they chose to disable partial matching (giving NULL + warning instead), but instead of giving an error when requesting a column that does not exist, which would be natural if we ignore historical context, giving NULL + a warning as a compromise.  So how to do this without a warning?
```
!is.null(signals$time_type) # bad behavior for data.frame, okay but warning behavior for tibble
# -->
!is.null(signals[["time_type"]]) # expected behavior without warning for both
"time_type" %in% names(signals) # another way
rlang::has_name(signals, "time_type") # another way
```
the three ways here work and seem equivalent for data.frames, tibbles, as well as for lists provided that you aren't putting in "" instead of "time_type".

Short-circuiting scalar vs. non-short-circuiting vector logical operations:
```
c(FALSE, TRUE) || c(FALSE,TRUE) # FALSE, without even a warning!  (`if` directly will give a warning, but not `||` apparently)
any(c(FALSE, TRUE) | c(FALSE,TRUE)) # TRUE

rlang::has_name(signals, "time_type") & signals$time_type == "blah" # if no column, logical(0) + warning
rlang::has_name(signals, "time_type") & (signals$time_type == "blah" || signals$time_type == "blah2") # if no column, FALSE + 2 warnings
rlang::has_name(signals, "time_type") && signals$time_type == "blah" # if no column, right result, but if there is a column, result based on 1st entry only, no warning!
# vs say
rlang::has_name(signals, "time_type") && any(signals$time_type == "blah") # looking better
rlang::has_name(signals, "time_type") && any(signals$time_type == "blah" & ......) # looking better
```



This is intended to warn people away from usage that could have them trying to use a column that they thought would exist but doesn't, getting NULL, and then getting confusing errors or garbage results, OR, using a prefix that they thought would uniquely identify a column they expected to exist, but it doesn't, and getting NULL or some random column instead, OR, as here, checking whether a column exists when it doesn't but it matches a prefix exactly one other existing column, and getting false positive results.....  Lots can go wrong.
`!is.null(signals$time_type)` -> `!rlang::has_name(signals, "time_type")`

brookslogan: (By the way, partial matching is also used for argument passing, without even a warning by default.  "Convenient".  This is probably one of the motivations for using dots as prefixes in dplyr function parameter names.  Something relevant to epiprocess and epipredict more than evalcast and the hospitalization forecaster, but keep in mind.)

brookslogan: (Corrected some comment above about behavior being the same for lists... it isn't quite.)

R-packages/evalcast/tests/testthat/test-download_signal.R
```diff
@@ -0,0 +1,146 @@
...
+        signal = "confirmed_incidence_num",
+        start_day = "2020-08-01",
+        end_day = "2022-01-01",
+        geo_type = "state",
+        geo_values = "*",
+        as_of = "2022-01-01"
+      )
+    )
+  )
+})
```
brookslogan, line 146: Nice tests!  Running them + others and looking at CI yields some more annoying cleanup to do:
- "Unknown or uninitialised column:" many times: resolved by moving away from `is.null(tbl$maybecol)`
- Need to expect warning at
  - at test-download_signal.R:115:4 ("cache ends earlier than")
  - at test-predictions.R:51:4 ("only selects weekly (Saturday) target end dates")
  - at test-predictions.R:154:4 (same)
  - at test-predictions.R:210:4 (same)
  - at test-predictions.R:291:4 (same)
- test failure: need to add time_type = "day" to expected mock args in populate_cache test.
  
Other CI checks I am able to understand / guess at:
```
Warning: Undocumented arguments in documentation object 'download_signal'
  ‘time_type’
```
^ Self-explanatory

```
Warning: get_predictions.Rd:117: Bad \link text
problem found in ‘get_predictions.Rd’
```
^ Looks like `\code` needs to be outside `\link` when linking to bettermc::mclapply for `\link` to work.  If we have `@md` on this entry or roxygen markdown mode enabled for the package, then it could just be
```
[`bettermc::mclapply`]
```
instead.

R-packages/evalcast/R/get_predictions.R
dshemetov: Thanks for writing this out, I will need to keep a link to this comment for future reference. Do you recommend any particular version that seems to avoid these issues? I'm thinking of just consistently using `"key" in names(df)`.

brookslogan: Writing this out was also helpful for me.  A few more notes.
- I'm considering using [this advice](https://kevinushey.github.io/blog/2015/02/02/rprofile-essentials/) minus `options(warn=2)` (that's a nice option to use while debugging stuff though).
- Some of the short-circuiting silent gotchas surprised me.  From [here](https://stat.ethz.ch/R-manual/R-devel/library/base/html/Logic.html): "Using vectors of more than one element in && or || will give a warning (as from R 4.2.0), or an error if the environment variable _R_CHECK_LENGTH_1_LOGIC2_ is set to a true value (this is intended to become the default in future). ".  Don't think 4.2 is out in many OS's yet, though(?).  Hope at least `Sys.setenv("_R_CHECK_LENGTH_1_LOGIC2_"=TRUE)` might help catch some of these if it works on <4.2.
- Might want to research if there is a way to try to check these issues and make them into errors within the context of our package's operations.

Regarding the name-check: not sure, I might recommend with caveats what you suggested: `"key" %in% names(df)` sounds good for checking for column names we know at "compile time", and should also work for atomic vectors and lists as well; just don't try `"" %in% names(df)` or for unchecked user-input(-derived) things in place of "key" that could include "" --- whatever you want it to do / makes sense given `?names` docs etc., it's probably not going to consistently do it.  Since it's not completely robust/general for programming purposes, I'm not completely happy with it, but I'm also not completely happy with any of the other choices I listed above (or the additional other choice I just stumbled across, `%has_name%` from `assertthat`), for similar sorts of reasons.

brookslogan:
Okay, I re-ran some experiments using `get_predictions` and the package checks.  Looks good!  (There may be a new warning about how many packages are in `Imports:`, but that's an issue for another day.)

R-packages/evalcast/R/get_predictions.R
dshemetov: Nice finds. I'll try out `options(warn = 2)` for a while and see how it works.

PR #585, opened 2022-06-09
[Testing] Use r2u-ci for evalcast CI

dshemetov:
This branch tests the new [PPA r2u](https://eddelbuettel.github.io/r2u/) from [Eddelbuettel](http://dirk.eddelbuettel.com/) and uses it to install R packages in the CI using his handy [r2u-ci](https://eddelbuettel.github.io/r-ci/) scripts. If it works, this could dramatically speed up installation speeds in CI.

- [ ] Make sure that the package caching `${{ env.R_LIBS_USER }}` works (or update it, if it doesn't)

### DIFF
--- a/.github/workflows/r_evalcast_ci.yml
+++ b/.github/workflows/r_evalcast_ci.yml
@@ -27,13 +27,6 @@ jobs:
     defaults:
       run:
         working-directory: R-packages/evalcast/
-    strategy:
-      matrix:
-        include:
-          - {os: macOS-latest}
-          - {os: ubuntu-latest}
-
-    runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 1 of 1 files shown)


dshemetov:
@krivard @brookslogan Is there a way to test CI changes on a PR like this?

dshemetov:
Alright, well if I do `R_VERSION=4.0` and `runs-on: ubuntu-latest`, I get the error `E: Unable to locate package r-cran-bspm
`. On the other hand, if I do `R_VERSION=3.5` and `runs-on: ubuntu-18.04`, I get the error `Error: package ‘bspm’ was installed before R 4.0.0: please re-install it`. 

So that `run.sh` script would need to be modified.

PR #587, opened 2022-07-19
Fix `evalcast:::get_forecast_dates` bug

dshemetov:
An alternative approach to the same ends as in #586. 

Probably more long term safe, but maybe moving too fast in deprecating `get_covidhub_forecast_dates`. A more complete version of this work would switch us completely to using zoltr, while putting in the html parse fix, to continue to support code that continues to use `get_covidhub_forecast_dates`. I just don't have the bandwidth to take that on right now, especially since the zoltr get predictions function appears to be broken now.

```r
r$> get_zoltar_predictions("CMU-TimeSeries", forecast_dates = "2022-07-18")
get_token(): POST: https://zoltardata.com/api-token-auth/
get_resource(): GET: https://zoltardata.com/api/projects/
get_resource(): GET: https://zoltardata.com/api/project/44/timezeros/
Error: POST status was not 200. status_code=400, json_response=Invalid query. error_messages='["target with name not found. name=1 wk ahead inc hosp, valid names=['17 day ahead inc hosp', '17 wk ahead cum death', '17 wk ahead inc death', '18 day ahead cum death', '18 day ahead inc death', '2 wk ahead inc case', '101 day ahead inc hosp', '102 day ahead cum death', '102 day ahead inc death', '102 day ahead inc hosp', '103 day ahead cum death', '16 wk ahead cum death', '16 wk ahead inc death', '17 day ahead cum death', '17 day ahead inc death', '18 day ahead inc hosp', '18 wk ahead cum death', '1 wk ahead inc case', '3 wk ahead inc case', '4 wk ahead inc case', '5 wk ahead inc case', '6 wk ahead inc case', '7 wk ahead inc case', '8 wk ahead inc case', '18 wk ahead inc death', '19 day ahead cum death', '19 day ahead inc death', '19 day ahead inc hosp', '19 wk ahead cum death', '19 wk ahead inc death', '20 day ahead cum death', '20 day ahead inc death', '20 day ahead inc hosp', '20 wk ahead
```

cc @nmdefries @brookslogan for input.

### DIFF
--- a/R-packages/evalcast/R/get_covidhub_predictions.R
+++ b/R-packages/evalcast/R/get_covidhub_predictions.R
@@ -152,7 +152,7 @@ get_forecast_dates <- function(forecasters,
   forecaster_dates <- vector("list", length = length(forecasters))
   for (i in seq_len(length(forecasters))) {
     forecaster_dates[[i]] <- tryCatch({
-      lubridate::as_date(get_covidhub_forecast_dates(forecasters[i]))
+      lubridate::as_date(get_zoltar_forecast_dates(forecasters[i]))
     },
     error = function(e) cat(sprintf("%i. %s\n", i, e$message))
     )
@@ -228,7 +228,7 @@ get_forecaster_predictions <- function(covidhub_forecaster_name,
   url <- "https://raw.githubusercontent.com/reichlab/covid19-forecast-hub/master/data-processed"
   pcards <- list()
   if (is.null(forecast_dates))
-    forecast_dates <- get_covidhub_forecast_dates(covidhub_forecaster_name)
+    forecast_dates <- get_zoltar_forecast_dates(covidhub_forecaster_name)
   forecast_dates <- as.character(forecast_dates)
   for (forecast_date in forecast_dates) {
     filename <- sprintf("%s/%s/%s-%s.csv",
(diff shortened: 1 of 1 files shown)


nmdefries:
For now, I think it's fine to combine the changes here with the html fix in case anyone is still using `get_covidhub_forecast_dates`.

The `target with name not found` error is because of improper target construction. Hospitalizations are forecast on a daily basis (`1 day ahead inc hosp`) but when `epiweek` `incidence_period` is selected we construct invalid weekly hospitalization targets. Same error happens for cases and deaths if `incidence_period` is set to `day`. I'm opening an issue with more info.

brookslogan:
Do we know whether covidhub and zoltar always have the same results?  If not, it's probably better to just do the html fix, and keep this zoltar mix-and-match as an emergency backup plan.

brookslogan:
I couldn't immediately find anything guaranteeing compatibility between covidhub and zoltar, and I assume that there are at the very least delays in syncing zoltar.  So I think for now we should just do #588 and refer back to this PR (#587) if GItHub updates the HTML again and breaks things.

nmdefries:
The GitHub REST API also provides a stable way of grabbing file names from a repo, e.g. [this chunk](https://github.com/cmu-delphi/forecast-eval/blob/d76b0432d5a2d47e3c96fa096d75f5ca90383372/Report/create_reports.R#L63-L160) uses the API to get names of changed files.

dshemetov:
Thanks for the link @nmdefries! 

Did some digging in that so I understand it better. Here are my notes.

You can get the files/folders present in a repo with the [`trees` endpoint](https://docs.github.com/en/rest/git/trees#get-a-tree):

```
# Example:
https://api.github.com/repos/reichlab/covid19-forecast-hub/git/trees/master
```

So to find the dates, you'd need to make the request above, find the hash for `data-processed`, make another request, find the hash for the particular forecaster, and collect the available forecast names, parsing out the dates.

brookslogan:
I believe we ended up fixing this instead with https://github.com/cmu-delphi/covidcast/pull/588, which merged the changes from #586 from the `evalcast-forecast-eval` (used by forecast eval dashboard) branch into `evalcast` (used by production forecast display Rmd).  I think we can just close this PR.

PR #588, opened 2022-07-20
Update `evalcast` branch with covidhub xpath fix on `evalcast-forecast-eval`

brookslogan:
This fix makes `get_covidhub_forecast_dates` and `get_covidhub_forecasts` work again.  It's from #586.  An alternative fix for `get_covidhub_forecasts` is #587, relying on using `get_zoltar_forecast_dates`.

Aside: I think the download caching in the evalcast branch is opt-in (and only concerns `get_predictions`, not fetching predictions, and is fairly robust now I think), so the evalcast branch can probably be used by the forecast eval dashboard.  (Although maybe it's still a good idea to keep this separate branch for stability and consciously update it when needed.  It's a tradeoff.)


brookslogan:
@nmdefries @dshemetov

dshemetov:
lgtm.

Also agree the forecast_eval would likely not be affected by syncing with the current evalcast branch.

dshemetov:
I'm planning to just merge this without codeowners approval tomorrow @brookslogan

dajmcdon (approved):

nmdefries:
I'm switching the forecast eval pipeline to use the `evalcast` branch so making fixes is easier.

PR #571, opened 2021-10-23
Prepare covidcast 0.4.3 release for CRAN

capnrefsmmat:
This branch contains a variety of changes I made while preparing for CRAN release. I will solicit some testers and then prepare to submit to CRAN.

PR #590, opened 2022-09-19
Fix for dev tidyselect

hadley:
Which no longer uses .data (since that's really a data-masking thing)


hadley:
@capnrefsmmat Just a note that I'm starting the release process for tidyselect with the goal of submitting to CRAN in two weeks, on October 10.

capnrefsmmat:
Thanks for the notice and the pull request. We had some changes ready for release (#571) that were blocked on a problem with our API server, but that appears to be resolved now, so we should be able to submit a new version to CRAN soon.

PR #595, opened 2022-09-27
Update codeowners for `evalcast`

nmdefries:
Switch from Dan and Jacob, who are no longer active on `evalcast`, to @nmdefries, @dshemetov, and @brookslogan.


capnrefsmmat (approved):
I added @brookslogan to the repo with write access so this would be valid (only people with write access can be owners, apparently), but this looks good to me

PR #601, opened 2022-11-17
Drop duplicate truth data join in `evaluate_covid_predictions`

nmdefries:
### Description
Currently, truth data is joined on to `predictions_cards` in both `evaluate_covid_predictions`, before calling `evaluate_predictions`, and in `evaluate_predictions`. `evaluate_predictions` is passed separate `actual_data` and doesn't use the truth data contained in `predictions_cards`, so the first join is doing unnecessary work.

### Changes
- Remove join in `evaluate_covid_predictions` that adds on truth data.

### DIFF
--- a/R-packages/evalcast/R/evaluate_predictions.R
+++ b/R-packages/evalcast/R/evaluate_predictions.R
@@ -129,12 +129,6 @@ evaluate_covid_predictions <- function(predictions_cards,
   assert_that("predictions_cards" %in% class(predictions_cards),
               msg = "predictions_cards must be of class `predictions_cards`")
   geo_type <- match.arg(geo_type)
-  grp_vars <- c("data_source",
-               "signal",
-               "incidence_period",
-               "geo_value",
-               "forecast_date",
-               "ahead")
   actual_data <- get_covidcast_data(predictions_cards, backfill_buffer,
                                     geo_type, offline_signal_dir)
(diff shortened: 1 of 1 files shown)


brookslogan:
Note/technicality as I read through old implementation:
- It looks like the old/current `evaluate_predictions` would actually "favor" the actuals already attached to the `predictions_cards`.  More precisely, it looks like, if `predictions_cards` already has actuals attached, it takes each row of the `predictions_cards`, and sees how many times a matching actual measurement (that is, same index vars and `actual` value) appears in the second `truth_data`: 0 times -> add row once into the output; >0 times -> add row that many times into the output.  So if the second `truth_data` had different `actual`s they will be ignored; and if it had matching measurements that were duplicated for some reason, it'd duplicate some `predictions_cards` rows.  But having duplicate rows doesn't make sense in `truth_data` anyway.

Technically, people could still attach `actual`s to predictions cards outside of `evaluate_covid_predictions` and run into similar behavior.  I'm not sure if that's an intended feature, so I've just made this a warning for now.  It also seemed like an opportunity to get rid of the noisy `by`-selection messages.

brookslogan (commented):
Looked good to me, but I added a couple other changes.  @nmdefries could you please look over my edits?

nmdefries:
Edits look good :+1:

PR #606, opened 2022-12-12
Support Ubuntu 22.04 in CI

nmdefries:
https://github.com/cmu-delphi/covidcast/pull/604 fixes test workflow issues on `main`. We want to apply the same changes to the `evalcast` branch. To do so without making duplicate commits, merge `main` into `evalcast` branch.

Note: `main` makes no changes to the evalcast package, so behavior of the package on the evalcast branch should be unchanged. `main` makes changes to the covidcast package and documentation.


krivard (approved):
YES thank you

brookslogan (approved):
Looks good to me.  Interesting situation where some of the check warnings/failures have already been addressed in [the] `evalcast` [branch].  It seems like it might be more proper to do checks on whatever the PR merge result would be rather than on the branch-to-be-merged-in.

PR #607, opened 2022-12-13
Enforce column order of predictions to match `arrow` schema

nmdefries:
### Description
Support newer version of `arrow` by:
- Enforcing column order in forecast files. Do this by reading header in (`readLines` is a fast way of doing so), and if it disagrees with the expected order, read, reorder, and re-write forecast file.
- Ignoring header row in forecast files on read.

Test failure is resolved in https://github.com/cmu-delphi/covidcast/pull/606. Tests pass locally for me.

### Context
Some submitted predictions have a column order that doesn't agree with the `arrow` schema specified in `get_forecaster_predictions_alt`. Due to a change in behavior in newer versions of `arrow`, the column order in input files and in the arrow `schema` must now be the same (previously were reordered). We pinned `arrow` to a working version as a patch, but that version is now quite old.


brookslogan (approved):
I think this works & addresses the issue, but have a few minor efficiency, user messaging, and code comment notes.  (Also encountered https://github.com/cmu-delphi/covidcast/issues/608, but I think that's an old issue, not introduced in this PR.)

nmdefries:
Superseded by https://github.com/cmu-delphi/covidcast/pull/652

PR #610, opened 2023-01-09
covidcast 0.4.5

### LINKED ISSUE #605: covidcast R package fails additional checks on CRAN
This is blocking us from releasing updates to the package, including the changes in #598 that improve compatibility with the latest ggplot.

CRAN does checks on standard platforms that ensure all tests run and the package can be built. (Despite the ggplot update, these checks pass on the current package because vdiffr makes plot tests all pass on CRAN, to avoid platform-specific graphical output issues.) It also does "additional checks" on non-standard setups. One of these is a system with [ATLAS](https://math-atlas.sourceforge.net/) used as the BLAS implementation; the system setup is [documented here](https://www.stats.ox.ac.uk/pub/bdr/Rblas/README.txt).

The package tests and vignettes fail with ATLAS. The [output is here](https://www.stats.ox.ac.uk/pub/bdr/Rblas/ATLAS/covidcast.out). The failures are in sf as it transforms various geometries and, apparently, transforms them into invalid geometries.

Logan set up a VM with ATLAS using [flexiblas](https://cran.r-project.org/package=flexiblas); see [Slack thread](https://delphi-org.slack.com/archives/C013AH5N01E/p1667917260676239). He was unable to reproduce the failures.

If we want to update the package, we'll need to reproduce and diagnose the issue. My hypothesis is that the problem is due to `shift_pr()`, which multiplies by a rotation matrix and hence could be affected by BLAS implementation. This is supported by the specific errors in the CRAN output:

```
══ Failed tests ════════════════════════════════════════════════════════════════
  ── Error ('test-plot.R:57'): simple state choropleths ──────────────────────────
  Error in `(function (mapping = aes(), data = NULL, stat = "sf", position = "identity", 
      na.rm = FALSE, show.legend = NA, inherit.aes = TRUE, ...) 
  {
      c(layer_sf(geom = GeomSf, data = data, mapping = mapping, 
          stat = stat, position = position, show.legend = show.legend, 
          inherit.aes = inherit.aes, params = list2(na.rm = na.rm, 
              ...)), coord_sf(default = TRUE))
  })(color = "white", size = 0.1, mapping = structure(list(geometry = ~geometry), class = "uneval"), 
      fill = "gray", data = structure(list(STATEFP = "72", STATENS = "01779808", 
          AFFGEOID = "0400000US72", GEOID = "72", STUSPS = "PR", 
          NAME = "Puerto Rico", LSAD = "00", ALAND = 8868701898, 
...
```

All of the errors that indicate the geometry in the output give Puerto Rico as the invalid geometry.

But without a way to reproduce the issue, we'll be unable to test this hypothesis, fix the problem, and convince the CRAN maintainers we've fixed it so they'll accept an update.

capnrefsmmat:
This is the version published on CRAN as 0.4.5, plus rebuilt documentation.

Closes #605.


capnrefsmmat:
Since this is already on CRAN and we're only making minor changes to the package, I'm just going to merge it.

PR #614, opened 2023-02-03
Update `get_predictions` to not use `bettermc` when `parallel_execution=FALSE`

### LINKED ISSUE #599: Provide way to avoid `bettermc::mclapply` in `get_predictions`
Currently, `get_predictions` always uses `bettermc::mclapply`, even with just 1 core selected or with `parallel_execution = FALSE` (this just sets the number of cores to 1).  This can complicate debugging in the case of metaforecasters (ensembles, calibrated forecasters, forecasters with sensorized covariates, etc.) that run `get_predictions` within their bodies, because, e.g., while running `debugger` on a debug dump and inspecting objects from stack frames, another debug dump can be generated and overwrite the first one, and not all frames from the traceback are selectable in the debugger.

We could
- add a parameter to just use `lapply` regardless of all other arguments, and make sure not to pass `lapply` any of the `bettermc::mclapply` extra arguments in this case, or
- see if any of the arguments rely on a feature of `bettermc::mclapply` (this isn't just having more than one core; there is also, e.g., retrying, and maybe other stuff), and if not, dispatch to `lapply` as above, making sure not to pass extra args)

dshemetov:
- Fixes #599 
- Also addresses a failing test and a [series of new warnings from dplyr==1.1.0 when using joins](https://github.com/tidyverse/dplyr/releases/tag/v1.1.0)

### DIFF
--- a/R-packages/evalcast/R/get_predictions.R
+++ b/R-packages/evalcast/R/get_predictions.R
@@ -40,10 +40,11 @@
 #'   (e.g. predict 1 day, 2 days, ..., k days ahead). Note that `ahead` is a
 #'   required component of the forecaster output (see above).
 #' @param parallel_execution TRUE, FALSE, or a single positive integer. If TRUE,
-#'   uses bettermc::mclapply to execute each forecast date prediction in
-#'   parallel on max number of cores - 1. If FALSE, the code is run on a single
-#'   core. If integer, runs in parallel on that many cores, clipping to the max
-#'   number of detected cores if a greater number is requested.
+#'   executes each forecast date prediction in parallel on the number of
+#'   detected cores available minus 1. If FALSE, the code is run on a single
+#'   core. If integer, runs in parallel on that many cores, clipping to the
+#'   number of detected cores available if a greater number is requested. Uses
+#'   [`bettermc::mclapply`] for parallelism if using more than one core.
 #' @param additional_mclapply_args a named list of additional arguments to pass
 #'   to [`bettermc::mclapply`] (besides `X`, `FUN`, and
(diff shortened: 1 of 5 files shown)


R-packages/evalcast/R/get_predictions.R
```diff
@@ -126,7 +126,11 @@ get_predictions <- function(forecaster,
     return(preds)
   }
 
-  out <- rlang::inject(bettermc::mclapply(forecast_dates, get_predictions_single_date_, mc.cores = num_cores, !!!additional_mclapply_args)) %>% bind_rows()
+  if (parallel_execution) {
```
brookslogan, line 129: issue: parallel_execution could be an integer as well, right?  Or have we overwritten it with a Boolean in this case?

brookslogan: (think R does auto coerce to Boolean here but doesn't do what we want)

brookslogan (approved):
I like the tests!  I pushed some edits addressing some nitpicks, e.g., using `rlang::is_bool` in more places.  See commit message for listing and justification (one thing I missed: I tried to make things more type-consistent in our code (integer-type with `L` suffixes) and more type-forgiving ("is_integerish") for the user).  As discussed via Slack, I'm keeping the logic here geared at disabling bettermc whenever multiple cores are requested, to avoid issues with error dumps in recursive parallelism situations and to make `options(error=recover)` work when using 1 core.  This could potentially trigger some lower level `mclapply` with `mc.allow.recursive=FALSE` to start parallelizing, which could be surprising, but this is an issue to solve another time.

Please check over my edits and merge if they look okay.

PR #616, opened 2023-02-13
Update `modeltools` dependencies

dshemetov:
Modeltools dependencies are out of date and incomplete. Fixing them:
* switch evalcast@main -> evalcast@evalcast
* make quantgen dependency explicit
* make lubridate, MMWRweek, tibble dependencies explicit


dshemetov:
This changes the `evalcast` branch which is only needed in the vignettes and the vignettes compile. Merging after confirming with @brookslogan offline.

PR #629, opened 2023-05-31
Fix Epidata.auth in covidcast.py

dmytrotsko:
Epidata.auth should be tuple with "epidata" as first element. This change is done in order to make client to send proper request to the backend

### DIFF
--- a/Python-packages/covidcast-py/covidcast/covidcast.py
+++ b/Python-packages/covidcast-py/covidcast/covidcast.py
@@ -34,7 +34,7 @@ def use_api_key(key):
     <https://cmu-delphi.github.io/delphi-epidata/api/api_keys.html>`_
     for details on our API key policies.
     """
-    Epidata.auth = key
+    Epidata.auth = ("epidata", key)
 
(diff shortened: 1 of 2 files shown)


melange396:
Looks good!  We should also update https://github.com/cmu-delphi/covidcast/blob/072e31727b7b5ad84b2c8a0af9b6eb5822a02c92/Python-packages/covidcast-py/setup.py#L25 to `>=4.1.1` to make sure users are also in sync with the client library version (@
 [`delphi-epidata/src/client/packaging/pypi/setup.py`](https://github.com/cmu-delphi/delphi-epidata/blob/e977b014e6e5a82ab391a135b463a7672b66ad37/src/client/packaging/pypi/setup.py#L8) ) that actually uses the `auth` class variable.

melange396:
I dont see a reason to make the version dependency anything other than `"delphi-epidata>=4.1.1"`.  If someone is not concerned with API keys, and is using delphi-epidata v0.0, but a recent version of this, something strange is going on.

Issue #631, opened 2023-06-01
covidcast Python release 0.2.1 checklist 


krivard:
- [x] Increment the package version in `setup.py` and in Sphinx's `docs/conf.py`
- [x] Update `docs/changelog.rst`
- [x] Install the package and its dependencies locally with `make install`
- [x] Rebuild the documentation with `make sphinx`
- [x] Commit the documentation, open a PR, get it approved
- [x] Build the release artifacts with `make build`
- [x] Upload to PyPI

PR #632, opened 2023-06-01
Prepare covidcast-py for release 0.2.1

krivard:
Partially addresses #631 

includes:

* Fix to how API keys are installed in the underlying Epidata object
* Updated registration link
* Updated delphi-epidata dependency version

[Package documentation preview](http://rinkitink.ml.cmu.edu/covidcast/covidcast-py/html/index.html)


capnrefsmmat:
The documentation home page still contains the direct Google Forms link. Maybe shorten that warning to just point at the `covidcast.use_api_key()` documentation, so it's centralized there?

melange396 (commented):
👍

PR #641, opened 2023-06-23
bug(evalcast): remove title from labs in plots

### LINKED ISSUE #640: plot_calibration gives .data errors, based on type of plot, when following evalcast tutorial
I am trying to replicate the tutorial found at https://cmu-delphi.github.io/covidcast/evalcastR/articles/evalcast.html on my local machine. Things generally work (thanks Dmitry!) up until `plot_calibration`.

The first instance of `plot_calibration` is giving this issue, which occurs in `format_wedgeplot`, but refers to subsetting .data outside of a data mask context.
image

The second instance gives an identical issue and traceback to the first.

The third instance gives a similar subsetting .data issue, but says it occurs in `format_traditional_calib_plot`.
image

Are there packages on my end that I am missing and need to install/update?

dshemetov:
closes #640 

- removes `title` argument from the labs
- also updates the DESCRIPTION file with remotes that work with devtools, remotes, and pak

### DIFF
--- a/R-packages/evalcast/DESCRIPTION
+++ b/R-packages/evalcast/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: evalcast
 Type: Package
 Title: Tools For Evaluating COVID Forecasters
-Version: 0.3.1
+Version: 0.3.2
 Authors@R:
   c(
     person(given = "Daniel",
--- a/R-packages/evalcast/DESCRIPTION
+++ b/R-packages/evalcast/DESCRIPTION
@@ -56,9 +56,9 @@ LazyData: true
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.2.3
 Remotes:
-    bettermc=github::gfkse/bettermc@v1.1.2,
-    zoltr=github::reichlab/zoltr,
-    covidHubUtils=github::reichlab/covidHubUtils
+    gfkse/bettermc@v1.1.2,
+    reichlab/zoltr,
+    reichlab/covidHubUtils
 Imports:
     assertthat,
     covidcast,
--- a/R-packages/evalcast/NEWS.md
+++ b/R-packages/evalcast/NEWS.md
@@ -1,3 +1,9 @@
+# evalcast 0.3.2
+
+- Fix DESCRIPTION remotes so `remotes` and `devtools` work
+- Fix `plot_calibration` by removing badly specified titles in the `lab` argument
+- Pin `bettermc` to the GitHub version 1.1.2 to avoid a CRAN error
+
 # evalcast 0.3.1
 
 - Added a download_signal caching folder option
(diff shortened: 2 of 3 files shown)


brookslogan:
Thanks; do you know what the plot labels look like now in the vignette?

dshemetov:
They looked unchanged to me 🤷, but I didn't look suuuper closely

brookslogan (approved):
Looks good!  This should fix vignette rendering and some user bugs.  We might want to revisit in another Issue if ~the guide labels are unclear or eating up a bunch of space now~ [we want a title].

brookslogan:
If you look at the wedge plots [here](https://cmu-delphi.github.io/covidcast/evalcastR/articles/evalcast.html), for example, the text "Proportion below" and "Proportion above" labeling the colors should have changed.  It might be a better hack to just say "Proportion below" and "Proportion above" instead of the corresponding `sprintf` failures; I'm not sure what it's going to look like by default.

brookslogan:
Sorry, neverymind, my bad. There shouldn't be a change; `sprintf` with `NULL` returns `character(0)`.

dshemetov:
This is what the first plot looks like

image

brookslogan:
Yep, it makes sense now; I thought `NULL` acted like `""` in `sprintf` but it actually acts like `character(0)`, so we're going from buggily saying "no title" ----> not saying anything about the title & getting no title.

dshemetov:
Once CI finishes, I'll merge. I don't think we should continue adding features to this repo, so I wouldn't open a title issue.

zterner-mitre:
Thanks for all your work on this!! I really appreciate it. I'd be happy to confirm that things work on my end once the changes are finished, just let me know how to proceed at that time if you can.

dshemetov:
Thanks @zterner-mitre! Can you try running through the tutorial again after installing with

```r
pak::pkg_install("cmu-delphi/covidcast/R-packages/evalcast@ds/evalcast-plot")
```

zterner-mitre:
The first `plot_calibration` is giving this warning and this plot, which I think is probably not what you are aiming for?
image
image

The second `plot_calibration` seems good:
image

Third also seems good
image

zterner-mitre:
`plot_coverage` and `plot_width` are producing too many plots given the setup of the plots (all in one row, for instance), and also may be missing some data. I can share those with you here, or open another issue, whichever you prefer

dshemetov:
What does your `predictions_cards` look like? My first plot does not match yours.

dshemetov:
I think the issue might be that your plots are using the full forecast_dates instead of the reduced forecast_dates_dec

zterner-mitre:
Yep! That's it. I had manually defined `forecast_dates_dec` yesterday when I was having issues with `get_covidhub_forecast_dates` to be all dates within December. But using the sequence of 4 dates instead of 31 fixed everything. Thanks!

dshemetov:
Excellent!

PR #642, opened 2023-06-26
Attempt at adding time_type arg

natemcintosh:
I noticed that the documentation for `covidcast_signals()` says 

> The argument structure is just as in `covidcast_signal()`, except the first four arguments `data_source`, `signal`, `start_day`, `end_day` are permitted to be vectors.

However, `covidcast_signal()` has the keyword argument `time_type` and `covidcast_signals()` does not. When using `data_source = "nchs-mortality"`, you need to use `time_type = "week"` or it will error out. Hence, to call `covidcast_signals()` with `data_source = "nchs-mortality"`, we need to tell `covidcast_signal()` to use `time_type = "week"`.

Note: I'm having trouble getting the tests up and running locally, so I'm hoping the CI here will pick up any issues. Is there a suggested way to get the correct package versions?

### DIFF
--- a/R-packages/covidcast/R/covidcast.R
+++ b/R-packages/covidcast/R/covidcast.R
@@ -560,6 +560,7 @@ covidcast_signals <- function(data_source, signal,
                                 "hhs",
                                 "nation"
                               ),
+                              time_type = c("day", "week"),
                               geo_values = "*",
                               as_of = NULL, issues = NULL, lag = NULL) {
   N <- max(length(data_source), length(signal))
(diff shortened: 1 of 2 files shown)


capnrefsmmat:
Are your test problems related to the plotting tests? Those tend to be platform-dependent, since R's graphics engine behaves slightly differently on Windows, macOS, and Linux.

I've approved the tests to run on CI, but I expect this should work fine. Thanks for catching this omission. We're scheduled to release a new version of the package soon, so hopefully we can include this fix with that update.

capnrefsmmat:
That did it. Thanks for the fix!

capnrefsmmat (approved):

PR #645, opened 2023-06-30
Cache covidcast_meta response when possible

krivard:
We need to cache covidcast_meta responses so that vignettes will stay within the anonymous rate limit for Epidata queries.

Changes include:
* .request: default to text content of response, but permit returning the raw response
* covidcast_meta: conditionally request or rerequest data from covidcast_meta based on whether it's been called already or not
* new global: META_RESPONSE

This hasn't been tested yet; easiest is to wait for the release of cmu-delphi/delphi-epidata#1222


R-packages/covidcast/R/covidcast.R
```diff
@@ -642,7 +645,14 @@ covidcast_signals <- function(data_source, signal,
 #' @importFrom utils read.csv
 #' @export
 covidcast_meta <- function() {
-  meta <- .request("covidcast_meta", list(format = "csv"))
+  # use cached metadata whenever possible
+  META_RESPONSE <- ifelse(
+    is.na(META_RESPONSE),
+    .request("covidcast_meta", list(format = "csv"), raw = TRUE),
+    httr::rerequest(META_RESPONSE)
+  )
```
capnrefsmmat, line 653: This probably should be:

```suggestion
  META_RESPONSE <<- if (is.na(META_RESPONSE)) {
      .request("covidcast_meta", list(format = "csv"), raw = TRUE)
    } else  {
      httr::rerequest(META_RESPONSE)
  }
```

The important change is `<<-`, so we write to the package environment instead of the function local environment. Using `ifelse()` works here because it lazily evaluates its arguments, so we don't evaluate both eagerly, but that's non-obvious enough that I had to look it up. An `if` expression also seems nice.

krivard:
It seems like `with_mock_api` doesn't handle `httr::rerequest` calls, and you can't use both `with_mock_api` and `stub` at the same time 😱

krivard:
Example: 
```
> devtools::test_active_file("tests/testthat/test-stub.R")
[...]
══ Testing test-stub.R ═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
[ FAIL 2 | WARN 0 | SKIP 0 | PASS 0 ]list()
[ FAIL 3 | WARN 0 | SKIP 0 | PASS 0 ]

── Error (test-stub.R:12): with_mock_api supports httr::rerequest ──────────────
Error in `request_perform(resp$request, resp$handle)`: inherits(handle, "curl_handle") is not TRUE
Backtrace:
 1. covidcast::covidcast_meta()
      at test-stub.R:12:4
 2. covidcast:::.request_meta()
      at covidcast/R/covidcast.R:648:2
 3. httr::rerequest(META_RESPONSE)
      at covidcast/R/covidcast.R:1027:4
 4. httr:::reperform(r)
 5. httr:::request_perform(resp$request, resp$handle)
 6. base::stopifnot(is.request(req), inherits(handle, "curl_handle"))

── Error (test-stub.R:21): mockery::stub takes effect even when httptest::with_mock_api is active ──
Error in `request_perform(resp$request, resp$handle)`: inherits(handle, "curl_handle") is not TRUE
Backtrace:
 1. base::tryCatch(...)
      at test-stub.R:21:4
 3. covidcast::covidcast_meta()
 4. covidcast:::.request_meta()
      at covidcast/R/covidcast.R:648:2
 5. httr::rerequest(META_RESPONSE)
      at covidcast/R/covidcast.R:1027:4
 6. httr:::reperform(r)
 7. httr:::request_perform(resp$request, resp$handle)
 8. base::stopifnot(is.request(req), inherits(handle, "curl_handle"))

── Failure (test-stub.R:21): mockery::stub takes effect even when httptest::with_mock_api is active ──
mock object has not been called 1 time
Backtrace:
 1. base::tryCatch(...)
      at test-stub.R:21:4
 2. mockery::expect_called(mock_identical, 1)
      at test-stub.R:25:8

[ FAIL 3 | WARN 0 | SKIP 0 | PASS 0 ]
```

krivard:
Okay there are two places META_RESPONSE is assigned in this code: 
* initial creation & set to `NA`
* every time `.request_meta` is called

& I don't think there's a way to pick assignment operators so that we can both pass tests and run vignettes :(

| initial creation | .request_meta | tests | vignettes |
| - | - | - | - |
| `<-` | `<<-` | fail with `cannot change value of locked binding for 'META_RESPONSE'` | |
| `<<-` | `<<-` | pass | fail with `object 'META_RESPONSE' not found` |

so i'm flipping over to [using a dedicated environment](https://stackoverflow.com/a/12605694/551433), and that seems to be okay?

R-packages/covidcast/R/covidcast.R
```diff
@@ -48,6 +48,9 @@ COVIDCAST_BASE_URL <- 'https://api.covidcast.cmu.edu/epidata/'
 # Max rows returned by API
 MAX_RESULTS <- 1000000
 
+# Cache the httr covidcast_meta response when available
+META_RESPONSE <<- NA
```
capnrefsmmat:
```suggestion
META_RESPONSE <- NA
```
This one can stay `<-`; it's only inside the function that we need to save to the parent scope

capnrefsmmat:
Just pushed an error reporting fix. I'll try to test this out a bit more soon

R-packages/covidcast/R/covidcast.R
```diff
@@ -1057,6 +1060,22 @@ covidcast <- function(data_source, signal, time_type, geo_type, time_values,
                            body = params)
   }
 
+  report_api_errors(response)
+
+  write(c(response$url, as.character(testthat::is_testing())), file = "~/requests.txt", append = TRUE, ncolumns = 2)
```
krivard, line 1065: (is this still needed?)

```suggestion
```

melange396:
@dsweber2 and i successfully tested this PR and https://github.com/cmu-delphi/delphi-epidata/pull/1222 together in our dev environments -- the R client does not make a new request on subsequent calls to `covidcast_meta()` when the Cache-Control header is provided by the api server.

capnrefsmmat (approved):

PR #647, opened 2023-07-07
bug: fix get_covidhub_forecast_dates

### LINKED ISSUE #646: Issue 639 (get_forecast_dates) is showing previous issues
The issue presented here https://github.com/cmu-delphi/covidcast/issues/639 has come back. Maybe this happened as a result of the package getting updated?

image

(It also happens using get_covidhub_predictions, as specified in my second comment in that thread)

To resolve this issue, I updated the evalcast package using pak::pkg_install("cmu-delphi/covidcast/R-packages/evalcast") as instructed last time, but that didn't resolve the issue :(

PS - I commented at the end of the previous issue, but wasn't sure that would notify you, so I thought it might help to start the issue again here.

dshemetov:
Fixes #646.

My R installation is broken, so I can't run the package locally. Hopefully CI can test it here...

### DIFF
--- a/R-packages/evalcast/R/get_covidhub_predictions.R
+++ b/R-packages/evalcast/R/get_covidhub_predictions.R
@@ -432,16 +432,14 @@ get_covidhub_forecast_dates <- function(forecaster_name) {
     httr::GET() %>%
     httr::content() %>%
     purrr::pluck("tree") %>%
-    magrittr::extract(purrr::map_chr(., "path") == "data-processed") %>%
-    purrr::pluck(1)
+    magrittr::extract2(which(purrr::map_chr(., "path") == "data-processed"))
 
   # Find the forecaster folder.
   forecaster_folder <- submissions_folder$url %>%
     httr::GET() %>%
     httr::content() %>%
     purrr::pluck("tree") %>%
-    magrittr::extract(purrr::map_chr(., "path") == forecaster_name) %>%
-    purrr::pluck(1)
+    magrittr::extract2(which(purrr::map_chr(., "path") == forecaster_name))
 
   # Get the forecaster submission files.
   submission_file_pattern <- sprintf("(20\\d{2}-\\d{2}-\\d{2})-%s.csv", forecaster_name)
@@ -454,6 +452,7 @@ get_covidhub_forecast_dates <- function(forecaster_name) {
 
   # Extract the dates.
   submission_dates <- submission_files %>%
+    # get first group of each match, requiring exactly one match per filename:
     stringr::str_match_all(submission_file_pattern) %>%
     purrr::map_chr(2) %>%
     as.Date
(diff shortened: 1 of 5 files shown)


R-packages/evalcast/R/get_covidhub_predictions.R
```diff
@@ -414,26 +414,51 @@ get_forecaster_predictions_alt <- function(covidhub_forecaster_name,
...
-  lubridate::as_date(out[[1]][, 2])
+  url <- "https://api.github.com/repos/reichlab/covid19-forecast-hub/git/trees/master"
+
+  # Get the submissions folder.
+  submissions_folder <- url %>%
+    httr::GET() %>%
+    httr::content() %>%
+    purrr::pluck("tree") %>%
+    magrittr::extract(purrr::map_chr(., "path") == "data-processed") %>%
+    purrr::pluck(1)
```
brookslogan, line 436: suggestion: reverse my misguided suggestion regarding pluck-which; `purrr::pluck(which(purrr::map_chr(., "path") == "data-processed"))` seems like it would better catch unexpected formatting of the response.

R-packages/evalcast/R/get_covidhub_predictions.R
```diff
@@ -414,26 +414,51 @@ get_forecaster_predictions_alt <- function(covidhub_forecaster_name,
...
+
+  # Find the forecaster folder.
+  forecaster_folder <- submissions_folder$url %>%
+    httr::GET() %>%
+    httr::content() %>%
+    purrr::pluck("tree") %>%
+    magrittr::extract(purrr::map_chr(., "path") == forecaster_name) %>%
+    purrr::pluck(1)
```
brookslogan, line 444: ditto

R-packages/evalcast/R/get_covidhub_predictions.R
```diff
@@ -414,26 +414,51 @@ get_forecaster_predictions_alt <- function(covidhub_forecaster_name,
...
+  submission_files <- forecaster_folder$url %>%
+    httr::GET() %>%
+    httr::content() %>%
+    purrr::pluck("tree") %>%
+    purrr::map_chr("path") %>%
+    purrr::keep(function(x) stringr::str_detect(x, submission_file_pattern))
+
+  # Extract the dates.
+  submission_dates <- submission_files %>%
+    stringr::str_match_all(submission_file_pattern) %>%
```
brookslogan, line 456: suggestion (non-blocking): might be clearer to use something based on `^.....$` regex + `str_match`

brookslogan (approved):
Looks good to me.  I've done a little testing treating it as a black box, and skimmed the code:
- checks pass (except unrelated vdiffr stuff) on my machine
- I can fetch recent forecasts based on the most recent returned date(s).  Single or multiple dates as character or Date vectors work.
- `get_covidhub_predictions("CMU-TimeSeries")` doesn't appear to fail when using the default dates from this function, but is taking too long for my unstable internet connection to handle and confirm it completely works.
- Some other testing on `get_covidhub_predictions` that probably doesn't really test these changes...

I'm made some slight code tweaks + some non-blocking really minor suggestions as I correct some misunderstandings I had regarding some purrr and stringr functionality and try to figure out what seem to be clean paradigms we could apply more confidently in messier situations.

PR #649, opened 2023-07-10
Try to use GitHub PAT when fetching forecast dates; and check response status for rate limit

nmdefries:
Followup to #647. Because dates are now being fetched using the GitHub API, scripts that request a lot of dates (e.g. forecast eval pipeline) can run into rate limits. When the API is rate-limited, `get_covidhub_forecast_dates` raises a non-informative warning and returns empty dates, causing downstream failures that are hard to debug.

Look for a `GITHUB_PAT` environment variable first, then `GITHUB_TOKEN` if `*_PAT` doesn't exist. If either is found, use it to authenticate the GET requests (using [this header format](https://docs.github.com/en/rest/overview/authenticating-to-the-rest-api?apiVersion=2022-11-28#about-authentication)). Response status codes are checked and if a 403 status is returned, the function exits with an error. It doesn't make sense to warn and return anyway because the results are empty; or to retry with backoff because the rate limit resets after an hour.

Note: I used the API's message as the error message because I'm not sure a 403-status is unique to reaching the rate limit and I want the actual error to be clear, and checking the text of the message wouldn't be robust.

### DIFF
--- a/R-packages/evalcast/R/get_covidhub_predictions.R
+++ b/R-packages/evalcast/R/get_covidhub_predictions.R
@@ -426,17 +426,33 @@ get_forecaster_predictions_alt <- function(covidhub_forecaster_name,
 #' @export
 get_covidhub_forecast_dates <- function(forecaster_name) {
   url <- "https://api.github.com/repos/reichlab/covid19-forecast-hub/git/trees/master"
+  # Look for a GitHub API token to avoid rate limits.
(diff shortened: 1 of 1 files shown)


dshemetov (approved):
looks great, thanks for taking care of this!

i checked locally that both auth and unauth requests work and that the error comes through well 👍 

:shipit: :shipit: :shipit:

Issue #609, opened 2023-01-04
Handle both standard and non-standard column orders of CovidHub forecast files in `get_covidhub_predictions_alt`


nmdefries:
`arrow` version 4.0.1 _reorders_ columns to match the provided schema. Newer versions require that all csvs have columns in the same order as each other and as the schema.

This is a problem for forecasts from CovidHub. About 25% of forecast files use a non-standard column ordering. Files for different submission dates can use different orders even within the same forecaster.

The issue with different column ordering needs to be resolved in an efficient way before we can start using `arrow` 10.0.1.

https://github.com/cmu-delphi/covidcast/pull/607 offers an inefficient solution. It reads the header for every downloaded forecast file. If the header order disagrees from the standard order, the entire file is read in, columns reordered, and rewritten to disk. What with the reading and re-writing and re-reading again later (on the actual read by `arrow`), this isn't fast especially since a large proportion of files use non-standard column orders.

[Supports newer `arrow` versions](https://github.com/cmu-delphi/covidcast-docker/issues/14).

nmdefries:
Alternate proposal deprecating `arrow` entirely:

Fetch forecast files directly into memory with `data.table::fread`, which can read from urls. This is faster than downloading to disk with `arrow` by 4x but uses 50% more memory, although this particular spot in the code isn't memory-limited.

Haven't tested the speed of splitting the `target` column in the `arrow` vs `data.table` approaches.

brookslogan:
Two cents: I don't think we're taking advantage of `arrow` much, because we immediately `collect` the dataset into a data frame / tibble in RAM, and `arrow` adds significant install time.  I would suspect there'd be an R library that can already do that just as fast.  `data.table::fread` and `readr::read_csv`.  Whether they directly support reading from URLs doesn't seem like a determining factor; like I believe we already do with `_alt` (with `arrow`), we could download to a file, then read that file.

PR #652, opened 2023-07-18
Use `fread` instead of `arrow` to get data in `get_forecaster_predictions_alt`

nmdefries:
### Description
Some submitted predictions have a column order that doesn't agree with the `arrow` schema specified in `get_forecaster_predictions_alt`. Due to a change in behavior in newer versions of `arrow`, the column order in input files and in the `arrow` schema must now be the same (previously were reordered). We pinned `arrow` to a working version as a patch, but that version is now quite old.

Incidentally, `arrow` is [not providing much additional benefit](https://github.com/cmu-delphi/covidcast/issues/609#issuecomment-1381065204), so we aren't tied to it. Since we want to prioritize speed, `data.table::fread` is a good alternative. `data.table` also offers opportunities for future optimization. Since `fread` can read from URL (downloads to a temp file), we don't need to do our own download step, which simplifies the process somewhat.

### Fixes
Closes https://github.com/cmu-delphi/covidcast/issues/609


dshemetov (approved):
Tested locally and works fine with the stated edge cases (forecasters that change their column order from date to date), e.g.:

```
evalcast:::get_forecaster_predictions_alt("COVIDhub_CDC-ensemble", forecast_dates = a
    s.Date(c("2021-11-08", "2021-11-15"))) %>% tibble
```

returns consistent prediction cards.

dshemetov:
Consider bumping the patch version on evalcast btw and adding a note in the CHANGELOG? Non-blocking.

nmdefries:
@dshemetov Do you have merge ability? This is ready

PR #654, opened 2023-07-25
Handle cases where `value` and `quantile` columns can't be read as numeric

nmdefries:
Format differences in forecaster input files means that `value` and `quantile` fields can't all be read as numeric. To avoid errors, read as character and cast to double later. Also specify several `NA`-equivalent values to get `get_forecaster_predictions` working in more cases and to avoid warnings when casting to double in `get_forecaster_predictions_alt`.

The `NA` addition fixes `get_forecaster_predictions` for 8 of the 10 forecasters where read fails for some dates. To fix the remaining failures (for `PSI-DRAFT` and `PSI-DICE` for some dates), apply the same character->double conversion as above.


dshemetov (approved):
Looks good. Given how varied the data seems to be, actually pretty nice that we can handle it all with `fread` and a little bit of manual conversion.

nmdefries:
@dshemetov could you merge this?

PR #657, opened 2023-08-22
Rebuild documentation for covidcast 0.5.2

capnrefsmmat:
I realized I never merged this, so the docs site still shows the old version. This is already released on CRAN.


dsweber2 (approved):
Depending on how frequently we're releasing, might make sense to add something like a `sync-dev` bot which automatically makes a pull request for changes from any release branch to `main`. https://github.com/cmu-delphi/delphi-epidata/blob/8918506dd4c5f362243819313f8cf29eed7636ed/.github/workflows/release-helper.yml#L129C12-L129C12

dsweber2:
Wait this is actually not building. Looks like `covidcast.Rmd`, `external-data.Rmd`, and `plotting-signals.Rmd` all failed b/c of missing plots.

PR #658, opened 2023-08-24
Make sure GitHub date-request errors are raised

nmdefries:
Use `httr::stop_for_status` to error on any non-successful, non-403 (for which we have special handling) status. Remove the `tryCatch` wrapper so that the errors actually cause the call to exit.

I'm not sure what errors the `tryCatch` was originally intended to catch, but there are a couple reasons to remove it. First, without valid dates, downstream calls won't get valid forecast data. And fetching forecast dates is relatively fast. It isn't a big problem to rerun. It also tests GitHub API authentication, which is necessary for time-consuming forecast downloads later.


R-packages/evalcast/R/get_covidhub_predictions.R
```diff
@@ -151,11 +151,7 @@ get_forecast_dates <- function(forecasters,
   forecast_dates <- as_date(forecast_dates)
   forecaster_dates <- vector("list", length = length(forecasters))
   for (i in seq_len(length(forecasters))) {
-    forecaster_dates[[i]] <- tryCatch({
-      lubridate::as_date(get_covidhub_forecast_dates(forecasters[i]))
-    },
-    error = function(e) cat(sprintf("%i. %s\n", i, e$message))
-    )
+    forecaster_dates[[i]] <- lubridate::as_date(get_covidhub_forecast_dates(forecasters[i]))
```
dshemetov, line 153: mostly out of scope, but if you know off the top of your head, im curious: why are we using lubridate here instead of just as_date?

dshemetov (commented):
Same here, let's bump either minor or patch version number, and otherwise lgtm

R-packages/evalcast/R/get_covidhub_predictions.R
nmdefries: I'm not sure. (All the other `as_date`s should really be loading directly from `lubridate` since we're not importing the function.)

dshemetov:
Thoughts on adding a retry to get_forecast_dates in case it fails due to some HTTP hiccup?

nmdefries:
@dshemetov I added a retry step. This should be ready so please re-review/merge when you have a chance.

Issue #655, opened 2023-08-15
WIS numbers look wrong for cumulative deaths; CovidHub forecasts are identical


zterner-mitre:
Hi,

I am not sure this merits an issue, but I wanted to ask if the results I am getting are correct, since they _seem off_ without further explanation. In particular, there are a few things I have been noticing:

1. The WIS numbers for cumulative deaths as given by CDC ensembles seem very high for summer 2021. The "baseliner" gives more reasonable results, and in general, other forecasts give much more reasonable results. I omitted other forecastsfrom this plot since I was using IHME-CurveFit, which gives poor results and takes a very long time to download due to all their forecasts, but other forecasters do seem to perform better in cumulative deaths. (I also wanted to use one of the UChicago forecasts, but it does not seem to have this data available then.)
[Issues File - Timepoint2 - CDC - Cumulative Deaths Comparisons.pdf](https://github.com/cmu-delphi/covidcast/files/12345114/Issues.File.-.Timepoint2.-.CDC.-.Cumulative.Deaths.Comparisons.pdf)

2. At least two of the CovidHub ensembles give exactly the same forecasts and error metrics. This is apparent both in the attached PDF, since the two ensembles are colinear, and in this screenshot, where all the error metrics are the same.
![image](https://github.com/cmu-delphi/covidcast/assets/117375468/1bdf0bf3-fbec-49df-8fbf-5697afbe0859)

I am also attaching R code with a reproducible example, but I saved it as a .txt file since Github wouldn't let me upload an R file.
[reproducible-ex-evalcast-github-as-txt.txt](https://github.com/cmu-delphi/covidcast/files/12345147/reproducible-ex-evalcast-github-as-txt.txt)

My questions here are:
1. Is the WIS calculation correct for cumulative deaths? Or are there some issues with how these ensembles are created, and the fact that they have high WIS is correct? In particular, the cumulative deaths WIS seem here are much higher than the incident deaths one can see on the Delphi dashboard. Is that a function of things being cumulative, or something else?
2. Should these CDC ensembles have the exact same forecasts and errors?

nmdefries:
Thanks for reaching out.

[...]

PR #659, opened 2023-08-24
Don't sum cumulative truth values

nmdefries:
If "cumulative" is in requested signal name, take most recent value per group rather than summing.

Closes https://github.com/cmu-delphi/covidcast/issues/655


dshemetov:
Could you bump the minor version and put a note in the changelog? Otherwise looks good

nmdefries:
@dshemetov This is ready to merge

PR #662, opened 2023-09-20
Call `gc` during `_alt` forecast download function used for forecast eval pipeline

nmdefries:
The `forecast-eval` pipeline is running out of memory. Try adding in some `gc()`s to free memory.

Note: There are other places that could be useful to call `gc()`, e.g. during scoring, during truth data fetch, but they would also impact non-forecast eval use. Given that calling `gc()` is a crude fix for OOM issues and that it would be nice for forecast eval to move away from `evalcast` , I don't want to go deep on this change. This is just an easy intervention on a "private" function.


brookslogan (approved):
(If this were being used inside a fork-parallel loop run from a parent process with a bunch of stuff already in memory, we might be afraid that gc() would actually increase memory usage due to copy-on-write on OS memory pages.  Checked with @nmdefries that this wasn't the case.)

dshemetov (approved):
Thank you both for looking deeply into the consequences of this 👍

PR #663, opened 2023-09-22
Remove usage of PHP alias in the Python client

rzats:
Secondary PR to https://github.com/cmu-delphi/delphi-epidata/pull/1288 that should be merged & deployed alongside it - fixed the delphi-epidata version to the newest client.


capnrefsmmat:
Since the Epidata cient API is changing, we'll need to update the version number required in `setup.py` along with these changes, so users are guaranteed to have the updated version of the client.

rzats:
@capnrefsmmat good point! I bumped the version once and will change it again if the release ends up being e.g. 4.1.3.

Python-packages/covidcast-py/setup.py
```diff
@@ -22,7 +22,7 @@
     install_requires=[
         "pandas<2",
         "requests",
-        "delphi-epidata>=4.1.1",
+        "delphi-epidata>=4.1.2",
```
melange396, line 25:
```suggestion
        "delphi-epidata>=4.1.10",
```
`4.1.10` should be the next version of `delphi-epidata`

capnrefsmmat: since the next version has a breaking API change, shouldn't we be incrementing more than the minor version number?

dshemetov: agree that we should bump a bigger number

even more, that's the [patch number being bumped](https://semver.org/)

melange396: In even a fairly loose interpretation of the semantic versioning rules: yeah, we probably should be.  

At the same time, bumping the major number would belie the fact that this is a relatively tiny change in the scope of the rest of the repository...  Theres no new functionality and the only real effective change is to the signature of one public method, `Epidata.async_epidata()`.  I would also argue that method is probably ***very*** rarely used, though thats just based on a hunch -- i find it hard to believe that many (if any?) people are actually using it because it is poorly documented and it doesnt appear in established call chains, but i have no demonstrable evidence for this.  [It ***IS*** used [once in `covidcast-indicators`](https://github.com/cmu-delphi/covidcast-indicators/blob/39c02fa228fcb4de2b3f3d19469e862861351a18/nowcast/delphi_nowcast/epidata.py#L51), but we have full control over that code & execution.]

Still, your point stands and this change is worth more than a single increment of the patch-level version.

This would be a lot easier if the client and its versioning were decoupled from the rest of the API code, but we wont truly get that until we release and get adoption of `epidatpy`, which lives in its own repo.

This discussion properly belongs over at https://github.com/cmu-delphi/delphi-epidata/pull/1288 so i will copypasta much of this text there, and we can continue in the appropriate place.

melange396 (approved):
This looks good, aside from the pending versioning discussion and any resulting version number changes.  I would wait to merge til then.

melange396:
Did you get a chance to run tests for this with the updated client from https://github.com/cmu-delphi/delphi-epidata/pull/1288 ?  I think you might need to update the `.return_value`s for the `@patch`ed `async_epidata()`: https://github.com/cmu-delphi/covidcast/blob/474dc20ad2f123b5b89d21f8443ee107537fe84e/Python-packages/covidcast-py/tests/test_covidcast.py#L261-L263

melange396:
can we update [the changelog](https://github.com/cmu-delphi/covidcast/blob/f9a80e9d675c4f1734ddf69d046160d7c88facdc/Python-packages/covidcast-py/docs/changelog.rst?plain=1#L4-L10), and then follow [this set of instructions](https://github.com/cmu-delphi/covidcast/blob/f9a80e9d675c4f1734ddf69d046160d7c88facdc/Python-packages/covidcast-py/DEVELOP.md?plain=1#L53-L72) to update the version number for the package and rebuild docs and other artifacts?  then we will be prepped to upload to PyPI when everything else is done.

melange396:
@dshemetov makes a good point in https://github.com/cmu-delphi/delphi-epidata/pull/1288#issuecomment-1749368736 ...  we should probably put documentation and show warning messages for deprecating `_async_fetch_epidata()`, just in case anyone is using it.

Python-packages/covidcast-py/docs/changelog.rst
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+v0.2.2, October 10, 2023
+--------------------
+
+- Remove usage of PHP alias in the Python client.
+- Use Python package-specific "user agent" string.
+- Mark :py:func:`_async_fetch_epidata` as deprecated.
```
melange396, line 9:
```suggestion
- New feature: Web requests now identify themselves with a `covidcast`-specific "user agent" string.
- Update: Now uses newer version of the `delphi-epidata` libarary.
- Bug fix: Removed usage of PHP alias endpoint for compatibility with new `delphi-epidata` version.
- Deprecated function: Private method :py:func:`_async_fetch_epidata` has been deprecated.
```

melange396:
@rzats this is failing at the linter now, i guess you didnt go through that part of the tests?

Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -463,14 +473,16 @@ def _async_fetch_epidata(data_source: str,
                          issues: Union[date, tuple, list],
                          lag: int,
                          time_type: str = "day") -> Union[pd.DataFrame, None]:
-    """Fetch data from Epidata API asynchronously.
+    """[DEPRECATED] Fetch data from Epidata API asynchronously.
```
melange396, line 476:
```suggestion
    # pylint: disable=D401
    """[DEPRECATED] Fetch data from Epidata API asynchronously.
```
to appease linter

Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -463,14 +473,17 @@ def _async_fetch_epidata(data_source: str,
                          issues: Union[date, tuple, list],
                          lag: int,
                          time_type: str = "day") -> Union[pd.DataFrame, None]:
-    """Fetch data from Epidata API asynchronously.
+    # pylint: disable=D401
+    """[DEPRECATED] Fetch data from Epidata API asynchronously.
```
melange396, line 477:
```suggestion
    # pylint: disable=check_imperative_mood
    """[DEPRECATED] Fetch data from Epidata API asynchronously.
```
i guess im old and pylint disable doesnt accept codes anymore?

Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -463,14 +473,17 @@ def _async_fetch_epidata(data_source: str,
...
-    """Fetch data from Epidata API asynchronously.
+    # pylint: disable=check-imperative-mood
+    """[DEPRECATED] Fetch data from Epidata API asynchronously.
 
     signal() wraps this to support fetching data over a range of dates
     and stacks the resulting data frames.
 
     If no data is found, return None, so signal() can easily filter out these
     entries.
     """
```
melange396, line 484:
```suggestion
    """  # noqa: D401
```

melange396: looks like the error is actually from pydocstyle and not pylint

Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -463,14 +473,17 @@ def _async_fetch_epidata(data_source: str,
                          issues: Union[date, tuple, list],
                          lag: int,
                          time_type: str = "day") -> Union[pd.DataFrame, None]:
-    """Fetch data from Epidata API asynchronously.
+    # pylint: disable=check-imperative-mood
```
melange396, line 476:
```suggestion
```

Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -463,14 +473,16 @@ def _async_fetch_epidata(data_source: str,
...
-    """Fetch data from Epidata API asynchronously.
+    """[DEPRECATED] Fetch data from Epidata API asynchronously.
 
     signal() wraps this to support fetching data over a range of dates
     and stacks the resulting data frames.
 
     If no data is found, return None, so signal() can easily filter out these
     entries.
-    """
+    """  # noqa: D401
```
melange396, line 483:
```suggestion
    """
```
this didnt work

Python-packages/covidcast-py/covidcast/covidcast.py
melange396, line 476:
```suggestion
    """Fetch data from Epidata API asynchronously. [DEPRECATED]
```

Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -463,14 +473,16 @@ def _async_fetch_epidata(data_source: str,
                          issues: Union[date, tuple, list],
                          lag: int,
                          time_type: str = "day") -> Union[pd.DataFrame, None]:
-    """Fetch data from Epidata API asynchronously.
+    """Fetch data from Epidata API asynchronously. [DEPRECATED]
```
melange396, line 476:
```suggestion
    """Fetch data from Epidata API asynchronously [DEPRECATED].
```
this is a stupid rule being applied to a private function, and its beginning to seriously irk me...

melange396:
linting/pydocstyle is happy now, but now python tests are failing...  @rzats can you take a look?

rzats:
@melange396 latest round of tests is passing!

capnrefsmmat (approved):

PR #665, opened 2023-10-05
use python package-specific "user agent" string

### LINKED ISSUE #660: use a distinct user-agent for python client
...to distinguish it from [the Epidata client from `delphi-epidata`](https://github.com/cmu-delphi/delphi-epidata/blob/9954505be2e2754a4bf7c887fab41ad1a74931cc/src/client/delphi_epidata.py#L27) that it builds upon.

local R client already has its own user-agent: https://github.com/cmu-delphi/covidcast/blob/8d045b799630dbb3472276cb0ef927fa0b1abbeb/R-packages/covidcast/R/covidcast.R#L1054

melange396:
closes #660

### DIFF
--- a/Python-packages/covidcast-py/covidcast/covidcast.py
+++ b/Python-packages/covidcast-py/covidcast/covidcast.py
@@ -7,12 +7,22 @@ from typing import Union, Iterable, Tuple, List
 import pandas as pd
 import numpy as np
 from delphi_epidata import Epidata
+from delphi_epidata.delphi_epidata import _HEADERS
+from pkg_resources import get_distribution, DistributionNotFound
 from epiweeks import Week
 
 from .errors import NoDataWarning
 
+
 # Point API requests to the default endpoint
-Epidata.BASE_URL = "https://api.covidcast.cmu.edu/epidata"
+Epidata.BASE_URL = "https://api.covidcast.cmu.edu/epidata/api.php"
+# Prepend to Epidata client's user agent to specify this package and version
+try:
+    _ver = get_distribution("covidcast").version
+except DistributionNotFound:
+    _ver = "0.0.0"
+_HEADERS['user-agent'] = f"covidcast/{_ver} " + _HEADERS['user-agent']
+
 
 VALID_GEO_TYPES = {"county", "hrr", "msa", "dma", "state",  "hhs", "nation"}
 
--- a/Python-packages/covidcast-py/covidcast/covidcast.py
+++ b/Python-packages/covidcast-py/covidcast/covidcast.py
@@ -15,7 +15,7 @@ from .errors import NoDataWarning
 
 
 # Point API requests to the default endpoint
-Epidata.BASE_URL = "https://api.covidcast.cmu.edu/epidata/api.php"
+Epidata.BASE_URL = "https://api.covidcast.cmu.edu/epidata"
 # Prepend to Epidata client's user agent to specify this package and version
 try:
     _ver = get_distribution("covidcast").version


melange396:
@capnrefsmmat thoughts?

capnrefsmmat:
Seems reasonable to me. I had considered putting the version information in the user-agent previously, but didn't know how to get the version information without hardcoding it in a constant somewhere.

Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -7,12 +7,22 @@
 import pandas as pd
 import numpy as np
 from delphi_epidata import Epidata
+from delphi_epidata.delphi_epidata import _HEADERS
+from pkg_resources import get_distribution, DistributionNotFound
```
dshemetov, line 11: I couldn't find documentation for `pkg_resources`. My guess is it comes from `setuptools` which is being deprecated.

Maybe use `importlib.metadata` to future-proof? From [docs](https://docs.python.org/3.11/library/importlib.metadata.html?highlight=pkg_resources):

> importlib.metadata is a library that provides access to the metadata of an installed [Distribution Package](https://packaging.python.org/en/latest/glossary/#term-Distribution-Package), such as its entry points or its top-level names ([Import Package](https://packaging.python.org/en/latest/glossary/#term-Import-Package)s, modules, if any). Built in part on Python’s import system, this library intends to replace similar functionality in the [entry point API](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#entry-points) and [metadata API](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#metadata-api) of pkg_resources. Along with [importlib.resources](https://docs.python.org/3.11/library/importlib.resources.html#module-importlib.resources), this package can eliminate the need to use the older and less efficient pkg_resources package.

```py
> import pkg_resources
> import importlib.metadata
> print(pkg_resources.get_distribution('wheel'))
> print(importlib.metadata.version('wheel'))
wheel 0.41.2
0.41.2
```

dshemetov (changes requested):

Python-packages/covidcast-py/covidcast/covidcast.py
melange396: I shamelessly stole the version determination code from the delphi-epidata python client (that this client depends on): https://github.com/cmu-delphi/delphi-epidata/blob/d42517e714d6d1deebb6cffcdf3fb5adadc54481/src/client/delphi_epidata.py#L17-L25

I dont see where setuptools (as a whole) is being deprecated, but [pkg_resources](https://setuptools.pypa.io/en/latest/pkg_resources.html) did get [marked as such](https://github.com/pypa/setuptools/pull/3843) about half a year ago...  But before we rush into yanking it out: `importlib.metadata` requires python3.8 or a fancy requirements directive to pull in its backport, so its not such a simple drop-in replacement.  There are also two other files in this repo that already use pkg_resources (https://github.com/cmu-delphi/covidcast/blob/31c8bf806aee39dcab07d6257a5f6efaf2538052/Python-packages/covidcast-py/covidcast/plotting.py#L12 and https://github.com/cmu-delphi/covidcast/blob/31c8bf806aee39dcab07d6257a5f6efaf2538052/Python-packages/covidcast-py/covidcast/geography.py#L7), plus one in covidcast-indicators (https://github.com/cmu-delphi/covidcast-indicators/blob/39c02fa228fcb4de2b3f3d19469e862861351a18/_delphi_utils_python/delphi_utils/geomap.py#L15) as well as the aforementioned delphi-epidata client.

Lets kick this down the road a bit -- ill make GH issues to remove pkg_resources in each of the 3 repos.

Python-packages/covidcast-py/covidcast/covidcast.py
```diff
@@ -7,12 +7,22 @@
...
 from epiweeks import Week
 
 from .errors import NoDataWarning
 
+
 # Point API requests to the default endpoint
-Epidata.BASE_URL = "https://api.covidcast.cmu.edu/epidata"
+Epidata.BASE_URL = "https://api.covidcast.cmu.edu/epidata/api.php"
```
melange396, line 18:
```suggestion
Epidata.BASE_URL = "https://api.covidcast.cmu.edu/epidata"
```
whoops, i mustve edited the wrong version of the file which got `api.php` back into the `BASE_URL`

Python-packages/covidcast-py/covidcast/covidcast.py
dshemetov: Got it, thanks for the clarifications. Not super critical as Python packaging evolves super slowly, so makes sense to just stay consistent with our other code.

dshemetov: Found a [`pkg_resources` to `importlib_metadata` migration guide](https://importlib-metadata.readthedocs.io/en/latest/migration.html).

PR #667, opened 2024-01-04
Replace deprecated pkg_resources

### LINKED ISSUE #666: replace deprecated `pkg_resources`
see https://github.com/cmu-delphi/covidcast/pull/665#discussion_r1348146649

rzats:
Closes #666.

### Summary:

Removes uses of setuptools' deprecated `pkg_resources` library - these are replaced by the native `importlib.metadata` library, but also the `importlib_resources` backport. This is due to the features needed in the codebase, like [accessing resources within subdirectories](https://gitlab.com/python-devs/importlib_resources/-/issues/58), only becoming available in Python 3.9+.

### Prerequisites:

- [X] Unless it is a documentation hotfix it should be merged against the `dev` branch
- [X] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [X] Build is successful
- [X] Code is cleaned up and formatted

### DIFF
--- a/Python-packages/covidcast-py/covidcast/geography.py
+++ b/Python-packages/covidcast-py/covidcast/geography.py
@@ -4,14 +4,19 @@ import warnings
 from typing import Union, Iterable
 
 import pandas as pd
-import pkg_resources
+import importlib_resources
 
-COUNTY_CENSUS = pd.read_csv(
-    pkg_resources.resource_filename(__name__, "geo_mappings/county_census.csv"), dtype=str)
-MSA_CENSUS = pd.read_csv(
-    pkg_resources.resource_filename(__name__, "geo_mappings/msa_census.csv"), dtype=str)
-STATE_CENSUS = pd.read_csv(
-    pkg_resources.resource_filename(__name__, "geo_mappings/state_census.csv"), dtype=str)
+county_census = importlib_resources.files(__name__) / 'geo_mappings' / 'county_census.csv'
+with importlib_resources.as_file(county_census) as path:
+    COUNTY_CENSUS = pd.read_csv(path, dtype=str)
+
+msa_census = importlib_resources.files(__name__) / 'geo_mappings' / 'msa_census.csv'
+with importlib_resources.as_file(msa_census) as path:
+    MSA_CENSUS = pd.read_csv(path, dtype=str)
+
+state_census = importlib_resources.files(__name__) / 'geo_mappings' / 'state_census.csv'
+with importlib_resources.as_file(state_census) as path:
+    STATE_CENSUS = pd.read_csv(path, dtype=str)
 
 # Filter undesired rows from CSVs.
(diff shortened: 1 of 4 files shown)


Python-packages/covidcast-py/setup.py
```diff
@@ -30,7 +30,8 @@
         "imageio-ffmpeg",
         "imageio",
         "tqdm",
-        "epiweeks"
+        "epiweeks",
+        "importlib_resources==6.1.1",
```
capnrefsmmat: Should we be pinning the exact version? I'd assume we just want a compatible version, as in `importlib_resources~=6.1`, or something like that. Or is there a reason we need this exact version?

melange396: It definitely makes sense for this to less rigid than requiring an exact version, particularly since the other packages are specified with open-ended constraints (or none at all).  Based on the last comment linked gitlab issue, it looks like this could be:
```suggestion
        "importlib_resources>=1.3",
```

Backports are nice for functionality but potentially headache-inducing w.r.t. dependency complexity...  We could make the requirement "`importlib_resources>=1.3;python_version<"3.9"`" but that requires the code to know when it should use `importlib_resources` vs `importlib.resources`

Python-packages/covidcast-py/covidcast/geography.py
```diff
@@ -4,14 +4,19 @@
...
-
-COUNTY_CENSUS = pd.read_csv(
-    pkg_resources.resource_filename(__name__, "geo_mappings/county_census.csv"), dtype=str)
-MSA_CENSUS = pd.read_csv(
-    pkg_resources.resource_filename(__name__, "geo_mappings/msa_census.csv"), dtype=str)
-STATE_CENSUS = pd.read_csv(
-    pkg_resources.resource_filename(__name__, "geo_mappings/state_census.csv"), dtype=str)
+import importlib_resources
+
+county_census = importlib_resources.files(__name__) / 'geo_mappings' / 'county_census.csv'
```
melange396, line 9: since this is the top-level namespace for this sub-package, we should either prefix this variable name with an underscore or avoid creating the variable in the first place.  this is especially true because its name is the lowercase version of the name of a constant, so maybe make the name more descriptive if you want to keep it around.
```suggestion
_county_census_file_path = importlib_resources.files(__name__) / 'geo_mappings' / 'county_census.csv'
```

Python-packages/covidcast-py/covidcast/geography.py
```diff
@@ -4,14 +4,19 @@
...
+with importlib_resources.as_file(county_census) as path:
```
melange396, line 10: does this context manager afford us anything?  it looks like `pd.read_csv()` should be able to accept the `county_census` variable directly.

melange396 (changes requested):

Issue #669, opened 2025-06-23
Compatability with upcoming ggplot2 version


teunbrand:
Hi there,

We've been preparing a new major release for ggplot2 and found an issue during a reverse dependency check.
The issue we detect is that somewhere a `fill` variable is provided outside the `aes()` fucntion that is not of the same length as the data.
We believe the source of the issue is one of the fill arguments listed here:

https://github.com/cmu-delphi/covidcast/blob/c89e4d295550ba1540d64d2cc991badf63ad04e5/R-packages/covidcast/R/plot.R#L684-L693

In addition, there are also warnings about setting 0-length `fill` variables, but that may not be as pressing a concern.
You can test your code with the development version of ggplot2 by installing it as follows:

```r
# install.packages("pak")
pak::pak("tidyverse/ggplot2")
```

We aim to release the new ggplot2 version in about 2 weeks, and hope you can submit a fix to CRAN around that time. Hopefully this will inform you in a timely manner.

Best wishes,
Teun

capnrefsmmat:
Thanks for the notice. I think I've identified the issue and have a simple fix.

In this case the data was also 0-length, so the warning wasn't really necessary -- would it make more sense for geom_sf() to simply do nothing when there's no data, rather than complaining the corresponding fill is also zero-length?

teunbrand:
The suggestion makes sense but practically it'd get hairy because the 0-length parameter check occurs when `geom_sf()` is called whereas the data is only resolved once the plot is being build with `ggplot_build()`. It might be possible to delay the 0-length parameter check though.

capnrefsmmat:
I can go ahead and make the change then, unless you think delaying the check is something that will happen before your new release.

teunbrand:
Thanks that'd be great! We're a bit late in the release cycle for making such a change, but I can write it down for the future. Can also consider demoting the warning to a message, but that may also be the next cycle.

capnrefsmmat:
The fixes are now on CRAN, so we should be good to go.

teunbrand:
Thanks this is great!

PR #670, opened 2025-06-30
Fixes in anticipation of ggplot update

capnrefsmmat:
Changes in response to #669. The commit messages explain each issue that had to be fixed to keep the upcoming ggplot2 happy.

I'll test this again, and then we can submit to CRAN to avoid covidcast being deleted after the new ggplot2 version is released.


capnrefsmmat:
I'll need to check if the test failures are just rendering differences between Mac and Linux; I suspect they are. If so, I'll revert to the version that works on Linux.

dshemetov (approved):
We should update codeowners, since Rostyaslav no longer works for Delphi. Overall this looks fine, though I haven't tested locally.

capnrefsmmat:
I think the failing tests are all just cross-platform differences in plot rendering. Given that we're not going to be adding any features to the package or doing any more than minimal maintenance, I don't think it's worth trying to fix that, so I'll go ahead and submit the release. (The plot tests are not run on CRAN.)

